PR #247, opened 2013-04-15
Make using custom Dependency Resolvers way easier

anaisbetts:
Right now we have a method on top of `IMutableDependencyResolver` that will fill in the RxUI framework classes. This is great for people using `ModernDependencyResolver`, but most dependency resolvers _don't_ implement Register the way we do, so trying to get the default dependencies registered on IoC containers like MEF or NInject is a royal pain and completely non-obvious.

This PR makes it way easier, by adding a method to `RxApp`. Now this can be as simple as (for NInject):

``` cs
var kernel = new StandardKernel();

RxApp.InitializeCustomServiceLocator((obj, type) => kernel.Bind(type).ToConstant(obj));

RxApp.DependencyResolver = new FuncDependencyResolver(
    (type, contract) => kernel.GetAll(type, contract));
```

### DIFF
--- a/ReactiveUI/RxApp.cs
+++ b/ReactiveUI/RxApp.cs
@@ -196,7 +196,7 @@ namespace ReactiveUI
         /// 
         /// (obj, type) => kernel.Bind(type).ToConstant(obj)
         /// </param>
-        public static void InitializeCustomServiceLocator(Action<object, Type> registerMethod)
+        public static void InitializeCustomResolver(Action<object, Type> registerMethod)
         {
             var fakeResolver = new FuncDependencyResolver(null, 
                 (fac, type, str) => registerMethod(fac(), type));


systemsymbiosis:
I am bumping into this issue just now and can't quite work out what to do. I have to agree that this "is a royal pain and completely non-obvious!"

There is one pretty good example of how to use Autofac with  RxUI in the RxUI repository
https://github.com/reactiveui/RxUI6WithAutofac I am not 100% sure that the Autofac container in this sample is correctly initialized. When I was debugging only the Logger and LogManager appeared to be registered with the container. I could be wrong.

From what I can gather by reading the RxUI and Splat source code the following methods are how RxUI does it's magic to register the appropriate types with the Dependency resolver.

```c#
        public static void InitializeSplat(this IMutableDependencyResolver This)
        {
            This.Register(() => new DefaultLogManager(), typeof(ILogManager));
            This.Register(() => new DebugLogger(), typeof(ILogger));
        }

```

```c#
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;

namespace ReactiveUI
{
    public class Registrations : IWantsToRegisterStuff
    {
        {            
            registerFunction(() => new INPCObservableForProperty(), typeof(ICreatesObservableForProperty));
            registerFunction(() => new IROObservableForProperty(), typeof(ICreatesObservableForProperty));
            registerFunction(() => new POCOObservableForProperty(), typeof(ICreatesObservableForProperty));
            registerFunction(() => new NullDefaultPropertyBindingProvider(), typeof(IDefaultPropertyBindingProvider));
            registerFunction(() => new EqualityTypeConverter(), typeof(IBindingTypeConverter));
            registerFunction(() => new StringConverter(), typeof(IBindingTypeConverter));
            registerFunction(() => new DefaultViewLocator(), typeof(IViewLocator));
            registerFunction(() => new CanActivateViewFetcher(), typeof(IActivationForViewFetcher));
            registerFunction(() => new CreatesCommandBindingViaEvent(), typeof(ICreatesCommandBinding));
            registerFunction(() => new CreatesCommandBindingViaCommandParameter(), typeof(ICreatesCommandBinding));
        }
    }
}

```

I messed around for a bit with Ninject and I tried 
```C#
            var resolver = new NinjectDependencyResolver(container);
            resolver.InitializeSplat();
            Locator.Current = resolver;
```
But I do not think this worked

If the custom Dependency resolver has these types registered it should be good, right? I am finding it difficult to know how to automatically obtain what interfaces map to what concrete types and what lifetime scope to assign in the Register method of a custom IMutableDependencyResolver

The Autofac sample uses the following Register method

```C#
 public void Register(Func<object> factory, Type serviceType, string contract = null)
{
var builder = new ContainerBuilder();
if (string.IsNullOrEmpty(contract))
{
builder.Register(x => factory()).As(serviceType).AsImplementedInterfaces();
}
else
{
builder.Register(x => factory()).Named(contract, serviceType).AsImplementedInterfaces();
}
builder.Update(_container);
}

```

What is the equivalent code to configure Unity? 

I understand that you recommend just to stick with the default ModernDependencyResolver in most cases. Making this easier will lower the barriers for adoption of ReactiveUI. Developers could plugin the DI library they are already familiar with (want to learn) and could leverage the existing DI ecosystem. It appears that many DI containers are already making the transition to PCL so they should start working better with RxUI across platforms Ha!

Maybe we could expand the existing Autofac IMutableDependencyResolver example to include other popular DI containers? Then at least there would be a working reference example for folks to use when starting out. I'd be happy to contribute to this idea if it seems reasonable and I can get a working version of UnityDependencyResolver, NinjectDependencyResolver or both. But Unity for sure!

Maybe the following is a related issue? https://github.com/reactiveui/ReactiveUI/issues/590

PR #503, opened 2014-01-23
View / ViewModel Activation

anaisbetts:
Consider the following ViewModel constructor:

``` cs
public MyBrokenViewModel()
{
    UserError.RegisterHandler(x => {
        // NB: Stuff
    });
}
```

This is broken because every time we create MyBrokenViewModel, we create another error handler. What Do? What we really want for certain global things like UserError, is for UserError to be subscribed only when the View associated with the ViewModel is _visible_. However, that information isn't available to ViewModels, and even if it was, it's not super obvious. Let's fix it
## How does this work:

Activation allows you, for both Views and ViewModels, to set up _the things that should be_ **active** _when the View is visible_. Here's how you do it for ViewModels:

``` cs
public class MyWorkingViewModel : ReactiveObject, ISupportsActivation
{
    public ViewModelActivator Activator { get; protected set; }

    public ActivatingViewModel()
    {
        Activator = this.WhenActivated(d => {
            // d() registers a Disposable to be cleaned up when
            // the View is deactivated / removed
            d(UserError.RegisterHandler(x => {
                // NB: Stuff
            }));
        });
    }
}
```

Here's how it works for Views:

``` cs
public class MyWorkingView : UserControl, IViewFor<MyWorkingViewModel>
{
    public ActivatingView()
    {
        this.WhenActivated(d => {
            Console.WriteLine("Helloooooo Nurse!")
            d(Disposable.Create(() => Console.WriteLine("Goodbye, Cruel World")));
        });
    }
}
```

Note that calling `WhenActivated` in a View automatically means that the associated ViewModel gets notified for activated / deactivated changes (and in fact, you _must_ call `WhenActivated` in the View to get the ViewModel to be notified).
## TODO:
- [x] Finish Cocoa Views (waiting on #502)
- [x] Testability (i.e. a 'With' type method for activation)
- [x] How do subclasses of VMs work?

### DIFF
--- a/ReactiveUI/Activation.cs
+++ b/ReactiveUI/Activation.cs
@@ -57,6 +57,12 @@ namespace ReactiveUI
             });
         }
 
+        public static IDisposable WithActivation(this ISupportsActivation This)
+        {
+            This.Activator.Activate();
+            return Disposable.Create(This.Activator.Deactivate);
+        }
+
         public static IDisposable WhenActivated(this IViewFor This, Func<IEnumerable<IDisposable>> block)
         {
             var activationFetcher = activationFetcherCache.Get(This.GetType());


ramonahosu:
Hello,

What do you recommend to use for activating the view models only when controls are made visible? (other than subscribe on VisibleChanged)

By checking the implementations of IactivationForViewFetcher for both winforms and wpf, I can see that activation can happen even if a control does not become visible.
Eg. For Winforms, when having a user control that is not visible by default, the activation happens on HandleCreated.. 
Eg. For WPF activation will happen for user controls that are not visible due to activation on Load.

Regards

PR #531, opened 2014-02-28
Add a combine extension method to ReactiveList<T>

tberman:
Tried a simple test, and it worked without issues.


ghuntley:
Doing some PR cleanup, @paulcbetts :+1:  or :-1: or :fire:  /cc: @niik and @tberman

tberman:
:fire: it was a terrible idea every time we used it.

PR #547, opened 2014-03-26
Fix memory leaks with events.

jlaanstra:
This PR adds a `WeakEventManager` class to fix all kinds of memory leak issues that the different Xaml platforms have. Since these changes could break existing applications that depend on the strong references it would be great to add this to RxUI 6 before it is released.
## Problem
- `Button` instances attach to the `CanExecuteChanged` event on an `ICommand` instance. This causes a leak when the ViewModel exists longer than the View. Details: https://github.com/MvvmCross/MvvmCross/issues/552 and http://social.msdn.microsoft.com/Forums/silverlight/en-US/34d85c3f-52ea-4adc-bb32-8297f5549042/command-binding-memory-leak?forum=silverlightbugs
- `LongListSelector` instances attach to the CollectionChanged event on an `INotifyCollectionChanged` instance. Same problem as above. Details: http://social.msdn.microsoft.com/Forums/en-US/11684ea3-5c71-4354-a4e4-c337198f52b1/wp8-longlistselector-memory-leak?forum=wpdevelop
  It seems that other controls like `Pivot` have similar issues.
## Solution

To fix all memory leaks related to event subscription from the UI, `WeakEventManager` instances have been added. 
- The `CanExecuteChanged` event of `ReactiveCommand` is now handled with a `CanExecuteChangedEventManager`.
- The `CollectionChang(ing|ed)` events of `ReactiveList` are now handled with a `CollectionChang(ing|ed)EventManager`.
- The `PropertyChang(ing|ed)` events of `IReactiveObject` are now handled with a `PropertyChang(ing|ed)EventManager`.
## Implementation

The implementation is based on the `WeakEventManager` class in WPF and the code in http://social.msdn.microsoft.com/Forums/silverlight/en-US/34d85c3f-52ea-4adc-bb32-8297f5549042/command-binding-memory-leak?forum=silverlightbugs
It contains a `ConditionalWeakTable` for the Target <-> Delegate mapping to keep the reference to the event handler alive as long as the control itself is alive.
The second `ConditionalWeakTable` exists to make sure the event manager does not leak the source of the event. This would be the `ICommand` instance in the case of the `CanExecuteChanged` event.
## TODO:
- [x] Add unit tests that verify there are no leaks. I already verified with test applications that these changes solve the above issues.
- [ ] Verify that this does not break RxUI View bindings that might rely on the strong reference.

### DIFF
--- a/ReactiveUI.Platforms/Android/LayoutViewHost.cs
+++ b/ReactiveUI.Platforms/Android/LayoutViewHost.cs
@@ -99,19 +99,16 @@ namespace ReactiveUI.Android
             set { _ViewModel = (TViewModel)value; }
         }
 
-        public event PropertyChangingEventHandler PropertyChanging
-        {
+        public event PropertyChangingEventHandler PropertyChanging {
             add { PropertyChangingEventManager.AddHandler(this, value); }
             remove { PropertyChangingEventManager.RemoveHandler(this, value); }
         }
 
-        void IReactiveObject.RaisePropertyChanging(PropertyChangingEventArgs args)
-        {
+        void IReactiveObject.RaisePropertyChanging(PropertyChangingEventArgs args) {
             PropertyChangingEventManager.DeliverEvent(this, args);
         }
 
-        public event PropertyChangedEventHandler PropertyChanged
-        {
+        public event PropertyChangedEventHandler PropertyChanged {
             add { PropertyChangedEventManager.AddHandler(this, value); }
             remove { PropertyChangedEventManager.RemoveHandler(this, value); }
         }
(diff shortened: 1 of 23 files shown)


bratsche:
@jlaanstra Why does ReactiveList need to use the event manager here?  It seems like it should only be needed in the controls and view controllers. Other containers like ObservableCollection don't do this, so I'm trying to understand why ReactiveList needs to.

This seems to be breaking Xamarin.Forms ListViews that bind to ReactiveList since Xamarin.Forms already uses an event manager.

https://github.com/reactiveui/ReactiveUI/issues/806

PR #557, opened 2014-04-09
RoutingState.NavigateAndReset does not work

maratoss:
WPF RxUI 5.99.5 .NET4.5.
in 5.4.0 it works


kiwipiet:
I'm using 6.5 (Portable-Net45+Win8+WP8+WPA81) and NavigateAndReset does not seem to work.
Was this fixed before?
Is there a work around?

kiwipiet:
FYI, I copied the latest RoutedViewHost to my project and used that instead and now NavigateAndReset  works.

PR #594, opened 2014-05-12
Item Changing Zipped With Item Changed Test

jlaanstra:
Git doesn't like the same branch name apparently, so I pushed this to a new branch.

This the new PR for #508.


ghuntley:
@jlaanstra :+1: or :-1: or :fire: this PR?

ghuntley:
Closing out this PR as it's hella old (sorry about that). If you still want to PR this in then please do it against `develop` branch which is RxUI v7.

PR #599, opened 2014-05-19
Chang this warning to DEBUG level output

Haacked:
This just adds noise to our log files. Thoughts on making it a DEBUG warning?

### DIFF
--- a/ReactiveUI/POCOObservableForProperty.cs
+++ b/ReactiveUI/POCOObservableForProperty.cs
@@ -22,7 +22,7 @@ namespace ReactiveUI
         {
             var type = sender.GetType();
             if (!hasWarned.ContainsKey(type)) {
-                this.Log().Warn(
+                this.Log().Debug(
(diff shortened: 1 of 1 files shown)


ghuntley:
Let's schedule this in for RxUI v7? Are there any other areas you want to also slap with `[Conditional("DEBUG")]`?

kentcb:
I think there's still some misunderstanding (possibly mine) of the conditional compilation feature. How would we use it in this scenario?

We can't mark `GetNotificationForProperty` as conditional because it has a return value (if you think about it, this makes sense because how would it be removed if the call site used the return value in a subsequent calculation?).

And if we were to separate the logging call into a separate method and mark _that_ with `Conditional("DEBUG")`, it would be ReactiveUI calling that method, not the client code. Therefore, its inclusion would be contingent upon the compilation settings of ReactiveUI, not the client library.

Conditional compilation only works for parameterless methods that the client code calls directly, unless I'm missing something.

PR #647, opened 2014-06-19
By default fill in the values for IObservedChange.

jlaanstra:
This PR changes the default behavior of `IObservedChange` and `ObservedChange` to always have their value filled in. 
## Why

With the refactorings done to the binding system, `IObservedChange` is only used for `WhenAny` and `ObservableForProperty`. These two were already guaranteed to fill in the value.

`GetValue` has been removed as you can now just use the Value property. If the value is not provided initially, it is retrieved the first time it is accessed and is cached.
## Related improvements
- `IReactiveObject` now includes the value in the change notifications, to save unnecessary reflection.
- The values coming from `DependencyObject` change notifications are now also directly propagated to `IObservedChange`.
## TODO:
- [ ] When beforeChange was set to true the `ObservedChange` previously contained the current values of the property as can be seen in the `OFPNamedPropertyTestBeforeChange` test. I think it does make more sense to include the new value in the `ObservedChange` when `beforechange` is true.
- [ ] Decide what to do with indexer change notifications. Including the value does not make sense. Including the index is more useful, but I'm not sure what to do with ranges.
- [ ] On iOS the `UIControl` does not seem to have a `Value` property. I commented it out for now, but what value do we need to return? Is this a bug?
- [ ] We need a clear policy for the `Value` property, as its value now depends on when you access it. I prefer an `It is always filled in initially` policy.

### DIFF
--- a/Playground-Android/Playground-Android.csproj
+++ b/Playground-Android/Playground-Android.csproj
@@ -42,6 +42,8 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
+    <Reference Include="System.Runtime" />
+    <Reference Include="System.Threading.Tasks" />
     <Reference Include="System.Xml" />
     <Reference Include="System.Core" />
     <Reference Include="Mono.Android" />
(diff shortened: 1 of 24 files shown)


kentcb:
@jlaanstra is this still a thing? Obviously this is very stale now and am wondering whether I can close it.

ghuntley:
closing due to age. 

@jlaanstra is this still a thing?

PR #706, opened 2014-08-20
Nsview activation

mteper:
Worked around Xamarin.Mac bug where overriding certain NSView methods (e.g. `WillMoveToSuperview` results in `ArgumentNullException` being thrown when calling base method with null value).

ReactiveView will trigger deactivation notifications when a view gets swapped with another view, but it will not currently trigger a notification when the containing window goes away.


ghuntley:
@paulcbetts /cc: @mteper 

:+1: or :-1:  or :fire: this PR?

kentcb:
This already appears to be part of RxUI7 code base, so I'm closing this.

PR #772, opened 2014-12-09
Use an AutoResetEvent instead of a Monitor in TestScheduler.WithScheduler

flagbug:
This prevents `TestScheduler.WithAsync` from throwing a SynchronizationLockExceptionObject exception, because we can't `await` inside a Monitor


paulcbetts:
Thanks @flagbug!

PR #783, opened 2014-12-29
Added ReactiveNavigationController

BrianSakhai:
I copied ReactiveTabBarController to make this, but changed the activation to be on ViewWillAppear instead of ViewDidAppear to line up with pull request #779.

Not sure if nobody has needed/wanted this yet or if it just a bad idea to make UINavigationController reactive.


paulcbetts:
Thanks @BrianSakhai!

PR #794, opened 2015-01-21
Created ReactiveActionBarActivity

marcosbozza:
I created the reactive implementation for ActionBarActivity from Android AppCompat library.
I added this to the AndroidSupport library.


paulcbetts:
Thanks!

PR #771, opened 2014-12-08
Documentation

ghuntley:
# Introduction
It's time for versioned documentation and better samples!

If you want to help out with this please comment on this issue and I'll add you as a contributor to my personal repository.

See following for inspiration:

* http://jacobian.org/writing/what-to-write/
* http://jacobian.org/writing/technical-style/
* http://jacobian.org/writing/editors/

See following for collection of projects with beautiful documentation: https://github.com/PharkMillups/beautiful-docs for inspiration

## Operations:
- [x] Use Readthedocs.org to automatically build ReactiveUI documentation as it supports versioned documentation via branches.
- [x] Resolve https://github.com/tomchristie/mkdocs/issues/238
- [x] CNAME created for [docs.reactiveui.net pointed towards readthedocs](http://read-the-docs.readthedocs.org/en/latest/alternate_domains.html).
- [x] Determine what to use and how to automatically build class documentation.
- [x] Documentation Theme
- [x] Structure of documentation defined.
- [ ] Migrate existing docs/* across.
- [ ] Clean existing docs/* to drop all stale Rx4 and Rx5 references.

=========================================================================

## Ideas for samples?
- [ ] Sub-classing of UserError
- [ ] Handling subclassed UserError Inline or via Global Registration
- [ ] Using IRecoveryCommand and RecoveryOptionResult? "I have a LoginCommand which internally does an http request to my server. Depending on the Response (200, 401, <5xx and timeouts>), I'd like to either (200 -> dismiss login dialog / 401 -> show "bad credentials" alert message ("OK") / <5xx and timenouts> -> show "Error" alert message ("Cancel", "Retry"). I thought throwing an UserError for 401 and   <5xx and timenouts> with a retry command would make sense, but I'm not sure how to  use the built-in Recovery stuff in ReactiveUI."
- [ ] General Navigation - WPF and WinForms
- [ ] Xamarin Forms - Navigation, Binding on all platforms.
- [ ] Suspension Host on all platforms
- [ ] Service Location
- [ ] Caching using Akavache
- [ ] Integrating ModernHTTPClient with Fusillade and Refit
- [ ] Using Splat/IBitmap x-platform within the PCL and value converters on each platform to display images.
- [ ] Reactive Dialogs. PCL Dialog Service -> iOS/Android impl which is observable based on which button pressed. Yes, No, Cancel.
- [ ] Example of a custom RoutedViewHost  (say iOS tab menu or navigation stack)

PR #800, opened 2015-02-05
Added implementation of IViewFor<T> for XAML

kmcginnes:
I think things will be much clearer with an example of each platform's implementation of `IViewFor<T>`. I've provided the one for WPF. I assume it works equally well for the other XAML based platforms.

I would provide the other two as well, but I don't have much experience in them. If some one can leave a comment on this pull request, I'd be glad to add them to my changes.

### DIFF
--- a/docs/basics/bindings.md
+++ b/docs/basics/bindings.md
@@ -24,6 +24,35 @@ implement it differently:
   RaiseAndSetIfChanged, *or* implement `INotifyPropertyChanged` on your View and
   ensure that ViewModel signals changes.
 
+```csharp
+public partial class NotificationsListViewController : ReactiveTableViewController, IViewFor<NotificationsListViewModel>
+{
+    public NotificationsListViewController()
+    {
+    }
+
+    public override void ViewDidLoad()
+    {
+        base.ViewDidLoad();
+        ViewModel = new NotificationsListViewModel();
+        // ... view stuff
+    }
+
+    NotificationsListViewModel _ViewModel;
+    public NotificationsListViewModel ViewModel
+    {
+        get { return _ViewModel; }
+        set { this.RaiseAndSetIfChanged(ref _ViewModel, value); }
+    }
+
+    object IViewFor.ViewModel
+    {
+        get { return _ViewModel; }
+        set { ViewModel = (NotificationsListViewModel)value; }
+    }
+}
+```
+
 * **Android:** - change your base class to one of the Reactive Activity /
   Fragment classes (i.e. ReactiveActivity<T>), *or* implement
   `INotifyPropertyChanged` on your View and ensure that ViewModel signals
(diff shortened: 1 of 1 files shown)


pdoh00:
Great idea. I even wrote a snippet for that one. I think explaining the registration with

```Locator.CurrentMutable.Register(() => new View(), typeof(IViewFor<ViewModel>));```

would be really helpful.  To be honest it's still a little magic to me.  Here is the snippet for a WPF app if that helps

```xml
<?xml version="1.0" encoding="utf-8" ?>
<CodeSnippets  xmlns="http://schemas.microsoft.com/VisualStudio/2005/CodeSnippet">
  <CodeSnippet Format="1.0.0">
    <Header>
      <Title>IViewForProp</Title>
      <Shortcut>iv4prop</Shortcut>
      <Description>Generates the dependency prop for IViewFor</Description>
      <SnippetTypes>
        <SnippetType>Expansion</SnippetType>
      </SnippetTypes>
    </Header>
    <Snippet>
      <Declarations>
        <Literal>
          <ID>classname</ID>
          <ToolTip>View type</ToolTip>
          <Function>ClassName()</Function>
          <Default>ClassNamePlaceholder</Default>
        </Literal>
      </Declarations>
      <Code Language="csharp">
        <![CDATA[public static readonly DependencyProperty ViewModelProperty =
            DependencyProperty.Register("ViewModel", typeof($classname$Model), typeof($classname$), new PropertyMetadata(null));

        public $classname$Model ViewModel
        {
            get { return ($classname$Model)GetValue(ViewModelProperty); }
            set { SetValue(ViewModelProperty, value); }
        }

        object IViewFor.ViewModel
        {
            get { return ViewModel; }
            set { ViewModel = ($classname$Model)value; }
        }$end$]]>
      </Code>
    </Snippet>
  </CodeSnippet>
</CodeSnippets>
```

kmcginnes:
Very nice! That gives me an idea. There's a lot of boilerplate code when using ReactiveUI. I, myself, have quite a few snippets that I use all the time. 

Maybe the ReactiveUI repo could host some of these snippets in Gists. If that doesn't make sense, then maybe a new `docs/snippets` folder that could present them as you did: XML code and a description.

pdoh00:
Here are a few I use [RxUI Snippets](https://gist.github.com/pdoh00/7435255)

kmcginnes:
Ha. That looks a lot like my set of snippets. I was missing the `IViewFor` one though. The app I've been working on uses ReactiveUI extensively, but we never dived into the binding syntax. The little bit I've seen of it is great.

ghuntley:
Thanks for your contributions @kmcginnes / @pdoh00 your work has been merged into issue #771

I have also granted you read/write access to the documentation pull-request so that you can edit directly via GitHub markdown editor or via the instructions in the documentation issue so that you don't need to go via the PR system.

PR #804, opened 2015-02-09
Updated ReactiveUI_Mac64 to output to Xamarin.Mac folder.

mteper:
Not sure what else to update to have proper things done for NuGet, but at least now the unified mac build won't overwrite the MonoMac version.

### DIFF
--- a/ReactiveUI/ReactiveUI_Mac64.csproj
+++ b/ReactiveUI/ReactiveUI_Mac64.csproj
@@ -17,8 +17,8 @@
     <DebugSymbols>True</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>False</Optimize>
-    <OutputPath>bin\Debug\Xamarin.Mac</OutputPath>
-    <IntermediateOutputPath>obj\Debug\Xamarin.Mac</IntermediateOutputPath>
+    <OutputPath>bin\Debug\Xamarin.Mac10</OutputPath>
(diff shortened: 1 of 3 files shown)


ReactiveUI/ReactiveUI_Mac64.csproj
```diff
@@ -17,7 +17,7 @@
     <DebugSymbols>True</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>False</Optimize>
-    <OutputPath>bin\Debug\MonoMac</OutputPath>
+    <OutputPath>bin\Debug\Xamarin.Mac</OutputPath>
     <IntermediateOutputPath>obj\Debug\MonoMac</IntermediateOutputPath>
```
paulcbetts: Fix these too

paulcbetts:
You also need to create templates in the `NuGet` folder (just empty files), as well as patch `MakeRelease.ps1` to have the extra platform

mteper:
I think we have to start in Splat, no?

paulcbetts:
> I think we have to start in Splat, no?

Splat already has 64-bit Xamarin.Mac support afaik, we just need to make sure the minversion in our release is >= the version where it was added

mteper:
We are depending on Splat 1.6.0, I think. At least that's what the HintPath in the .xcodeproj file points to. As far as I can tell, that version included the Unified Xamarin.Mac binaries.

PR #805, opened 2015-02-10
added a failing test for ObservableAsPropertyHelper

pdoh00:
My first try at a PR to RxUI. Let me know if I am Doing It Wrong**TM**

This PR is to show a scenario I am running into, which ToProperty doesn't set the property when I think it should.  If you step through the test you will see the .Do actually executes, but then the property value is not set. I doubt the .sln change is needed, but I wasn't sure, so I left it in for now.


pdoh00:
The answer to this failing Pull Request is to only have one expression per ToProperty.  What was happening was that the second ToProperty paves over the first.  When the need arises to have multiple events fill the same property value then the answer is to use an Observable.Merge for each event and run the ToProperty off the Merge observable.

PR #811, opened 2015-02-22
Fix handling of null PropertyName values in INPCObservableForProperty

asarium:
The documentation for [INotifyPropertyChanged.PropertyChanged](https://msdn.microsoft.com/en-us/library/system.componentmodel.inotifypropertychanged.propertychanged.aspx) says that `null` and `string.Empty` are valid values for the PropertyName and indicate that the whole object has changed. The current version of `INPCObservableForProperty` ignores such values.
This pull request adds support for these values and also adds tests for the `INPCObservableForProperty` class.


paulcbetts:
Thanks @asarium!

Issue #740, opened 2014-10-17
ReactiveTableViewSource failures with dynamically changing data


kentcb:
I've been struggling with an intermittent issue with `ReactiveTableViewSource` with respect to dynamically changing section data. I've simplified this to the simplest possible thing that breaks and have [posted a repro here](http://cl.ly/3x080V281A1q/RxUIReactiveTableViewSourceRepro.zip). It's using RxUI 6.1.0.

The app just creates a list of data items, each of which has a `Name` and an `IsActive` flag. The main view is a table view controller with two sections of data: active and inactive. Data items are randomly switched between active and inactive states, and those changes need to be reflected in the UI.

My first approach was to enable change tracking on the main collection in the hope that the derived collections would not need to signal resets whenever any item's `IsActive` is toggled. Doing this causes the following exception:

```
MonoTouch.Foundation.MonoTouchException: Objective-C exception thrown.  Name: NSInternalInconsistencyException Reason: attempt to delete row 11 from section 0 which only contains 11 rows before the update
  at at (wrapper managed-to-native) MonoTouch.ObjCRuntime.Messaging:void_objc_msgSend (intptr,intptr)
  at MonoTouch.UIKit.UITableView.EndUpdates () [0x00010] in /Developer/MonoTouch/Source/monotouch/src/build/compat/UIKit/UITableView.g.cs:437
  at ReactiveUI.UITableViewAdapter.PerformBatchUpdates (System.Action updates) [0x00020] in /Users/kent/Repository/ReactiveUI/ReactiveUI/Cocoa/ReactiveTableViewSource.cs:69
  at ReactiveUI.CommonReactiveSource`4[RxUIReactiveTableViewSourceRepro.ItemViewModel,MonoTouch.UIKit.UITableView,MonoTouch.UIKit.UITableViewCell,ReactiveUI.TableSectionInformation`1[RxUIReactiveTableViewSourceRepro.ItemViewModel]].sectionCollectionChanged (Int32 section, IList`1 teas) [0x0026b] in /Users/kent/Repository/ReactiveUI/ReactiveUI/Cocoa/CommonReactiveSource.cs:375
[...]

PR #820, opened 2015-03-11
Issue 808 (and 740)

### LINKED ISSUE #808: Are dynamic table sections supported?
Hi,

I've been banging my head against a wall trying to get dynamic table sections working (using 6.4.0.1). I've run into at least three distinct issues, all probably related. The API seemingly supports my use case, but I'm wondering whether the implementation is known to be lacking...? I just want to know whether I should give up and try a different approach altogether or whether I'm possibly doing something wrong.

Essentially what I want is a table view where the sections and items within those sections are dynamically determined. My view model is written and tested, but binding to the view is throwing up errors around indexes being out of sync etcetera.

Is there a working sectioned table sample that includes this level of dynamism?

kentcb:
Fixes #808 and #740.

Apologies for the long descripion, but this was non-trivial to understand and fix so I want to elaborate for future (and current) reference. But let's start with some moving pictures:

Table View:

rxui_808_pr_-_tableview

Collection View:

rxui_808_pr_-_collectionview

The above is showing a repro that was failing very rapidly with the old code. Every 5-250 milliseconds, up to 18 items are randomly changed in a single UI message. One section (titled "Rare") is particularly, um, rare, so it pops in and out of existence quite often.

# The Problems

The existing implementation of `CommonReactiveSource` (which serves as the base class for both `ReactiveTableViewSource` and `ReactiveCollectionViewSource`) has a number of deficiencies that prevent it from successfully and accurately functioning when batches of changes are made to the underlying list of data. It's probably best to address each problem individually.

## Reload Completion Detection

Having accurate knowledge of when a reload is actually finished is crucial to being able to determine whether a change in the underlying data is already accounted for in the UI or not. If a change happens *after* iOS has finished reloading, it must be applied separately. If a change occurs *before* iOS has finished reloading, it must not be applied a second time when the reload completes.

The current code contains a hack to attempt to detect when iOS has finished processing a call to `ReloadData()`. Essentially, it uses a call to `NumberOfSections()` as a signal for completion of a previous call to `ReloadData()`. Alas, this does not always work and also occurs too early. iOS is yet to *completely* finish reloading when it calls this method - it calls other methods afterwards.

## Batching Changes

The original code takes the approach of using `Buffer()` to batch together changes and then apply them. However, the use of `RxApp.MainThreadScheduler` is allowing iOS to interject some of its messages, which means it could see an inconsistent view of the data. Moreover, the result is perhaps a little less than ideal because client code has no way of making a batch of changes appear as one. Since the application of changes is time-based rather than batch-based, things could be applied at any point in the batch (if it worked, which it doesn't).

## Index Normalization

The existing code made no attempt to "normalize" the changes within a batch to something that iOS can understand and work with. The simplest possible example is this:

```C#
someSection.Add("new item");
someSection.RemoveAt(0);
```

This will result in an exception with the old code. Suppose there are 2 items already in the list when this code executes. The index received in the `NotifyCollectionChangedEventArgs` when calling `Add()` will therefore be `2`. So the old code will tell iOS "insert at 2 and delete at 0". iOS applies deletions first and then expects insertion indices to be relative to the list *with deletions already applied*. Thus, the correct thing to tell iOS is "insert at 1 and delete at 0" which iOS will enact as "delete at 0 then insert at 1".

Far more complicated situations can arise because changes to the underlying data can be made in any order and indexes in those changes can refer to anywhere in the list. Thus, latter changes can affect the indices for prior changes and vice-versa. On top of that, iOS requires that we detect annulling changes and to not mention them. For example:

```C#
someSection.Insert(0, "transient item");
someSection.RemoveAt(0);
```

The above should result in no calls at all to insert/remove items because from iOS' perspective, nothing in the UI requires updating. On the other hand:

```C#
someSection.Insert(0, "new item");
someSection.RemoveAt(1);
```

That should result in "remove at 0 and delete at 0".

Basically, the whole index normalization thing is entirely non-trivial and difficult to get your head around (at least it was for me). Try to visualize, for example, how a batch of updates `"D0:A0:A0:A5:A2:D3:A5:D2:A3:A1"` (A=add, D=delete) would normalize to `"D0:A2:A0:A7:D1:A6:A4:A1"`. I had to spend a great deal of time shifting labelled paperclips about to get my head around the normalization process.

# The Solutions

## Reload Completion Detection

The fact that Apple hasn't surfaced an API for this is beyond terrible. There are lots of people struggling with this on the Interwebs.

The solution I settled on was to expose an `IObservable<bool>` from `IUICollViewAdapter` that ticks `false` if data is not reloading, or `true` if it is. The implementation basically involves calling `ReloadData()` and then immediately queueing another "finished" message. The idea is that 'RefreshData()' is queueing its own work and so any message scheduled afterwards will execute after the iOS work is completed.

Unfortunately, iOS screws things up by sometimes scheduling *additional* work (*after* I have scheduled my work) that will result in calls to `GetCell()`. This creates the possibility of the data changing before the calls to `GetCell()`, which could result in invalid index exceptions. So in addition to queueing the separate message, I also call `LayoutIfNeeded()` on the view once the "finish" message is executed. Only once that layout has been instigated (thus initiating the calls to `GetCell()`) do I tick `true` on the `IsReloading` observable. So now the client code (in `CommonReactiveSource`) can accurately determine whether reloading is happening or not.

## Batching Changes

The new code takes the approach of enacting changes immediately by use of the immediate scheduler. For clarity, it is purposely explicit in its use of the immediate scheduler, even though I believe it's the default for all cases.

By immediately detecting a change, the code can schedule a future message to apply those changes on the UI thread. Meanwhile, it collects changes to apply into a list. In addition, it can use the `IsReloading` observable in the adapter to determine whether changes that come in should instigate another reload or not.

The client code should also just use the immediate (default) scheduler when constructing its various reactive collections for the sections and the section's contents.

## Index Normalization

The index normalization algorithm is manifest in the `IndexNormalizer` class. There are accompanying tests, which is great because of the difficulty of reasoning about what the expected results should be.

All `NotifyCollectionChangedEventArgs`' are reduced down to one or more `Update` instances, which have a `Type` of either `Add` or `Delete`, plus an `Index`. The essence of the normalization algorithm is to take any number of these updates (whose indexes initially refer to the original index specified by client code) and return a new set of `Update` instances whose indices are iOS-specific and are calculated by examining the updates before or after them. Annulling updates are also removed.

# Possible Future Enhancements

In this PR, I've only made the changes necessary to fix the actual bugs. If I ever find the time, there are some other things I'd like to do...

## Bug in Collection View Support

There is a strange bug in the collection view implementation. Some time (> 1 minute) after my "Hammer Time!" scenario starts execution (see GIF above) I get an exception like this:

```
MonoTouch.Foundation.MonoTouchException: Objective-C exception thrown.  Name: NSInternalInconsistencyException Reason: request for index path for global index 10458164 when there are only 20 items in the collection view
  at MonoTouch.ObjCRuntime.Runtime.ThrowNSException (IntPtr ns_exception) [0x00000] in /Developer/MonoTouch/Source/maccore/src/ObjCRuntime/Runtime.cs:156 
  at MonoTouch.ObjCRuntime.Runtime.throw_ns_exception (IntPtr exc) [0x00000] in /Developer/MonoTouch/Source/maccore/runtime/Delegates.generated.cs:100 
  at (wrapper native-to-managed) MonoTouch.ObjCRuntime.Runtime:throw_ns_exception (intptr)
  at (wrapper managed-to-native) MonoTouch.ObjCRuntime.Messaging:void_objc_msgSend_IntPtr_IntPtr (intptr,intptr,intptr,intptr)
  at MonoTouch.UIKit.UICollectionView.PerformBatchUpdates (MonoTouch.Foundation.NSAction updates, MonoTouch.UIKit.UICompletionHandler completed) [0x0005f] in /Developer/MonoTouch/Source/monotouch/src/build/compat/UIKit/UICollectionView.g.cs:384 
  at ReactiveUI.UICollectionViewAdapter.PerformBatchUpdates (System.Action updates, System.Action completion) [0x0000e] in Y:\Repository\ReactiveUI\ReactiveUI\Cocoa\ReactiveCollectionViewSource.cs:84 
  at ReactiveUI.CommonReactiveSource`4[RxUIReactiveTableViewSourceRepro.ItemViewModel,MonoTouch.UIKit.UICollectionView,MonoTouch.UIKit.UICollectionViewCell,ReactiveUI.CollectionViewSectionInformation`1[RxUIReactiveTableViewSourceRepro.ItemViewModel]].ApplyPendingChanges () [0x0002b] in Y:\Repository\ReactiveUI\ReactiveUI\Cocoa\CommonReactiveSource.cs:325 
  at ReactiveUI.CommonReactiveSource`4[RxUIReactiveTableViewSourceRepro.ItemViewModel,MonoTouch.UIKit.UICollectionView,MonoTouch.UIKit.UICollectionViewCell,ReactiveUI.CollectionViewSectionInformation`1[RxUIReactiveTableViewSourceRepro.ItemViewModel]].<AttachToSectionInfo>b__20_9 () [0x00000] in Y:\Repository\ReactiveUI\ReactiveUI\Cocoa\CommonReactiveSource.cs:292 
  at System.Reactive.Concurrency.Scheduler.Invoke (IScheduler scheduler, System.Action action) [0x00000] in <filename unknown>:0 
  at ReactiveUI.NSRunloopScheduler+<>c__DisplayClass2_0`1[System.Action].<Schedule>b__0 () [0x00013] in Y:\Repository\ReactiveUI\ReactiveUI\Cocoa\NSRunloopScheduler.cs:35 
  at MonoTouch.CoreFoundation.DispatchQueue.static_dispatcher_to_managed (IntPtr context) [0x00036] in /Developer/MonoTouch/Source/maccore/src/CoreFoundation/Dispatch.cs:310 
NSInternalInconsistencyException: request for index path for global index 10458164 when there are only 20 items in the collection view
	0   CoreFoundation                      0x0077e946 __exceptionPreprocess + 182
	1   libobjc.A.dylib                     0x07415a97 objc_exception_throw + 44
	2   CoreFoundation                      0x0077e7da +[NSException raise:format:arguments:] + 138
	3   Foundation                          0x00e05810 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 118
	4   UIKit                               0x0380341b -[UICollectionViewData indexPathForItemAtGlobalIndex:] + 279
	5   UIKit                               0x037db2b3 -[UICollectionViewLayout prepareForCollectionViewUpdates:] + 634
	6   UIKit                               0x037ce9e4 -[UICollectionView _prepareLayoutForUpdates] + 322
	7   UIKit                               0x037d2fcb -[UICollectionView _updateWithItems:] + 1490
	8   UIKit                               0x037ce71c -[UICollectionView _endItemAnimations] + 14609
	9   UIKit                               0x037d45c8 -[UICollectionView _endUpdates] + 44
	10  UIKit                               0x037d477a -[UICollectionView performBatchUpdates:completion:] + 429
	11  ???                                 0x17731a14 0x0 + 393419284
	12  ???                                 0x17bb4c9c 0x0 + 398150812
	13  ???                                 0x17bb48e0 0x0 + 398149856
	14  ???                                 0x179d5403 0x0 + 396186627
	15  ???                                 0x179d5210 0x0 + 396186128
	16  ???                                 0x173080fe 0x0 + 389054718
	17  ???                                 0x1772e063 0x0 + 393404515
	18  ???                                 0x17724658 0x0 + 393365080
	19  ???                                 0x172fd464 0x0 + 389010532
	20  libdispatch.dylib                   0x0772d03f _dispatch_client_callout + 14
	21  libdispatch.dylib                   0x07716764 _dispatch_main_queue_callback_4CF + 470
	22  CoreFoundation                      0x006d895e __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 14
	23  CoreFoundation                      0x00697760 __CFRunLoopRun + 2256
	24  CoreFoundation                      0x00696bcb CFRunLoopRunSpecific + 443
	25  CoreFoundation                      0x006969fb CFRunLoopRunInMode + 123
	26  GraphicsServices                    0x098e724f GSEventRunModal + 192
	27  GraphicsServices                    0x098e708c GSEventRun + 104
	28  UIKit                               0x030a08b6 UIApplicationMain + 1526
	29  ???                                 0x1612b650 0x0 + 370325072
	30  ???                                 0x1612b450 0x0 + 370324560
	31  ???                                 0x16129788 0x0 + 370317192
	32  ???                                 0x16129534 0x0 + 370316596
	33  ???                                 0x1612966a 0x0 + 370316906
	34  RxUIReactiveTableViewSourceRepro    0x00185b45 mono_jit_runtime_invoke + 725
	35  RxUIReactiveTableViewSourceRepro    0x00223e9f mono_runtime_invoke + 127
	36  RxUIReactiveTableViewSourceRepro    0x00228ca1 mono_runtime_exec_main + 401
	37  RxUIReactiveTableViewSourceRepro    0x00228a64 mono_runtime_run_main + 628
	38  RxUIReactiveTableViewSourceRepro    0x000e41dd mono_jit_exec + 93
	39  RxUIReactiveTableViewSourceRepro    0x002c6324 xamarin_main + 2500
	40  RxUIReactiveTableViewSourceRepro    0x002dc945 main + 117
	41  libdyld.dylib                       0x07753ac9 start + 1
```

It looks like it might be a Xamarin/iOS issue, so I'll try to create a useful repro and submit to them. Note that there is nothing wrong with the indices being passed in - something funky is going on inside `PerformBatchUpdates()`, and the table view is not susceptible presumably because it does batch updates via a different mechanism.

## Section Change Support

Both the old code and new reload data whenever a section is added or removed. It would be cool if changes *to* sections underwent the same process as changes *within* a section. The existing code already has a TODO around this, but it should be far more viable now because normalizing indices for sections can use exactly the same logic as normalizing indices for changes within those sections.

Unfortunately, changes to sections can result in required changes to the section numbers recorded for changes within sections:

```C#
// change occurs against section 0
data[0].Insert(0, "Some data");
// but section 0 now becomes section 1
data.Insert(0, new List<string>());
```

Note that I have modified the `IUICollViewAdapter` interface and its implementations to include methods for manipulating individual sections, but nothing currently uses those new methods. I can remove if you think that necessary.

## Move Support

It would be awesome if the removal and insertion of the same item is detected as a move, and thus the correct move animation is triggered rather than separate remove and add animations.

# Random Thoughts

I put quite a bit of effort into a test harness (pictured in the GIFs above) that was super useful in tracking down various issues. For future reference, I have made it available [here](http://cl.ly/270m0d1q2N2U). It just needs package restore and a reference to the ReactiveUI_iOS project.


paulcbetts:
@kentcb Awesome. I'll have a look tomorrow.

ReactiveUI/Cocoa/CommonReactiveSource.cs
```diff
@@ -278,30 +234,74 @@ void resetupAll(IReadOnlyList<TSectionInfo> newSectionInfo)
             //
             // TODO: Instead of listening to Changed events and then reseting,
             // we could listen to more specific events and avoid some reloads.
-            disp.Add(sectionChanging.Subscribe(_ => {
+            disp.Add(sectionChanging.ObserveOn(ImmediateScheduler.Instance).Subscribe(_ => {
```
paulcbetts: Is there a reason we add `ImmediateSchedule` everywhere?

kentcb: Yes, I explained in the PR that `ImmediateScheduler` is required for correctness so I added it explicitly instead of relying on implicit use. Can remove if you prefer.

paulcbetts: Yeah, these are no-ops - I like the idea of being explicit for sure, but just to save the minor CPU time / stack inflation, let's document it via comments instead of via code

kentcb: Sorry for delay. Have made changes but I want to test before committing. Alas, my license recently expired. Even though I have renewed I need to refresh in XCode and it's unable to contact Apple servers atm.

image

*sigh*

kentcb: Several refreshes and restarts later, I can build again.

Have pushed an update to remove explicit immediate scheduler usage. Also fixed a minor type issue on non-UNIFIED builds, even though it had built fine until now. Seriously, Xamarin, I can't even.

paulcbetts: Doing better than I am, I can't even build _Events.iOS64 right now, it fails telling me I need to include System.Runtime 4.0

kentcb: orly? You building in XS or VS? I feel like something's balked recently with XS, and now I can't even check for updates - just hangs every time.

I would expect the error you said if you are trying to build in VS. It's the problem I mentioned on Slack whereby VS expects references that XS does not. I just Ctrl+. them all in and make sure I don't commit those project changes. Very annoying.

paulcbetts:
Alright, gonna get this merged soon and put out a new release, this is really fucking good @kentcb, amazing docs and a ton of work put into this. Thanks!

PR #824, opened 2015-03-16
Set the unit test scheduler via the property

Haacked:
This change ensures that the test
"SchedulerShouldBeCurrentThreadInTestRunner" passes. The test currently
fails.

### DIFF
--- a/ReactiveUI/RxApp.cs
+++ b/ReactiveUI/RxApp.cs
@@ -69,7 +69,7 @@ namespace ReactiveUI
                 LogHost.Default.Warn("ReactiveUI acts differently under a test runner, see the docs\n");
                 LogHost.Default.Warn("for more info about what to expect");
 
-                _MainThreadScheduler = CurrentThreadScheduler.Instance;
+                MainThreadScheduler = CurrentThreadScheduler.Instance;
                 return;
             } else {
                 LogHost.Default.Info("Initializing to normal mode");
--- a/ReactiveUI/RxApp.cs
+++ b/ReactiveUI/RxApp.cs
@@ -69,7 +69,15 @@ namespace ReactiveUI
                 LogHost.Default.Warn("ReactiveUI acts differently under a test runner, see the docs\n");
                 LogHost.Default.Warn("for more info about what to expect");
 
+                // This sets _UnitTestMainThreadScheduler but only on the current thread. It doesn't set 
+                // _MainThreadScheduler if it's already been set. And the registration code generally sets this to a
+                // WaitForDispatcherScheduler _before_ we get to this line.
                 MainThreadScheduler = CurrentThreadScheduler.Instance;
+                // This is why we _also_ have to set this field so that the default scheduler while running in unit
+                // tests is the immediate scheduler. This is particularly important on xunit 2 where every test 
+                // might be run in its own thread. We found that in those cases, _UnitTestMainThreadScheduler was 
+                // null and we would get a bunch of tests using the WaitForDispatcherScheduler for major loss.
+                _MainThreadScheduler = MainThreadScheduler;
                 return;
             } else {
                 LogHost.Default.Info("Initializing to normal mode");


Haacked:
Ok, the fix required one more step.

So when run our tests, [`Registrations.cs` line 75](https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI/Platform/Registrations.cs#L75) 

```csharp
RxApp.MainThreadScheduler = new WaitForDispatcherScheduler(() => DispatcherScheduler.Current);
```

always runs _before_ we set the `MainThreadScheduler` in [`RxApp.cs` line 72](https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI/RxApp.cs#L72).

This means that the field `_MainThreadScheduler` ends up set to the `WaitForDispatcherScheduler` even when running in unit tests.

The reason is because even though we try and set the `MainThreadScheduler` property to the immediate scheduler later (when running in unit tests), the setter for that property doesn't set `_MainThreadScheduler` if it's not `null`.

This normally wouldn't be a problem because that property _does_ set `_UnitTestMainThreadScheduler` to the immediate scheduler. The getter for the property always checks `__UnitTestMainThreadScheduler` first.

There's just one hitch. `__UnitTestMainThreadScheduler` is a thread static variable.

When I'm running all my tests in xunit 2.0, they all run in parallel on different threads. So what ends up happening is that many of my tests get the `WaitForDispatcherScheduler` which is plain wrong. The fix is to set both the property and then the field.Ok, the fix required one more step.

So when run our tests, [`Registrations.cs` line 75](https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI/Platform/Registrations.cs#L75) 

```csharp
RxApp.MainThreadScheduler = new WaitForDispatcherScheduler(() => DispatcherScheduler.Current);
```

always runs _before_ we set the `MainThreadScheduler` in [`RxApp.cs` line 72](https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI/RxApp.cs#L72).

This means that the field `_MainThreadScheduler` ends up set to the `WaitForDispatcherScheduler` even when running in unit tests.

The reason is because even though we try and set the `MainThreadScheduler` property to the immediate scheduler later (when running in unit tests), the setter for that property doesn't set `_MainThreadScheduler` if it's not `null`.

This normally wouldn't be a problem because that property _does_ set `_UnitTestMainThreadScheduler` to the immediate scheduler. The getter for the property always checks `__UnitTestMainThreadScheduler` first.

There's just one hitch. `__UnitTestMainThreadScheduler` is a thread static variable.

When I'm running all my tests in xunit 2.0, they all run in parallel on different threads. So what ends up happening is that many of my tests get the `WaitForDispatcherScheduler` which is plain wrong. The fix is to set both the property and then the field.

jlaanstra:
> So when run our tests, [`Registrations.cs` line 75](https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI/Platform/Registrations.cs#L75) 

> ```csharp
RxApp.MainThreadScheduler = new WaitForDispatcherScheduler(() => DispatcherScheduler.Current);
```

> always runs _before_ we set the `MainThreadScheduler` in [`RxApp.cs` line 72](https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI/RxApp.cs#L72).

I noticed the same, but I haven't figured out when/why this started happening. So far it looks like it has been broken since the first 6.x release.

File: ReactiveUI/RxApp.cs [hunk 1]
```diff
@@ -69,7 +69,15 @@ static RxApp()
...
-                _MainThreadScheduler = CurrentThreadScheduler.Instance;
+                // This sets _UnitTestMainThreadScheduler but only on the current thread. It doesn't set 
+                // _MainThreadScheduler if it's already been set. And the registration code generally sets this to a
+                // WaitForDispatcherScheduler _before_ we get to this line.
+                MainThreadScheduler = CurrentThreadScheduler.Instance;
+                // This is why we _also_ have to set this field so that the default scheduler while running in unit
+                // tests is the immediate scheduler. This is particularly important on xunit 2 where every test 
+                // might be run in its own thread. We found that in those cases, _UnitTestMainThreadScheduler was 
+                // null and we would get a bunch of tests using the WaitForDispatcherScheduler for major loss.
+                _MainThreadScheduler = MainThreadScheduler;
```
joshvera: Can we just remove `_MainThreadScheduler ?? value` in the setter below instead?

Haacked: I believe that would cause a problem when running in unit tests. If a unit test invokes an async operation, `_UnitTestMainThreadScheduler` is null on that other thread. This is because `_UnitTestMainThreadScheduler` is declared to be `ThreadStatic`. It's "static", but only on the current thread.

So if you don't set `_MainThreadScheduler` when it's null, then any async operation will have both `_UnitTestMainThreadScheduler` null and `_MainThreadScheduler` null.

Unless I'm missing something about your proposal.

ghuntley:
:+1: or :-1:  or :fire: this PR?

Issue #822, opened 2015-03-14
ViewModel and View are immediately deactivated on Windows Phone 8.1


flagbug:
I'm using `WhenActivated` in a Windows Phone 8.1 application and as soon as a page is loaded, the disposable in the `WhenActivated` method is immediately disposed.

I've stepped through the `ActivationForViewFetcher` class at https://github.com/reactiveui/ReactiveUI/blob/5954c53255832ea0fb9d8c8a6ccef18cd334f742/ReactiveUI/Xaml/ActivationForViewFetcher.cs#L39 and it seems like `IsHitTestVisible` changes from `true` to `false` and then back to `true` again. so when it reaches `false` the Disposable in `WhenActivated` is disposed

I don't have much insight into the activation code, so I'm not sure how this can be fixed

flagbug:
Oh, seems like someone on the RxUI mailing list had a similar problem: https://groups.google.com/forum/#!topic/reactivexaml/mgkhk3kCO5c

jcmm33:
I found its related to the built in transitions causing this, if you remove the transitions it will resolve the issue.  A better fix is needed however.

flagbug:
Hm, so basically we have to check if transitions are enabled, and if they are, ignore the first `true` and `false` value

jlaanstra:
Maybe we should just ignore IsHitTestVisible?

flagbug:
Well, that would work for me :smile:

No idea if that would break anything for anyone else

jcmm33:
Of the top of my head i can't see why removing it would have any detrimental affect.

PR #825, opened 2015-03-18
Simplify the Windows Phone ActivationForViewFetcher logic

flagbug:
While looking into https://github.com/reactiveui/ReactiveUI/issues/822, I found that the code was unnecessarily complex


paulcbetts:
Thanks @flagbug!

Issue #746, opened 2014-11-01
Xamarin Forms XAML binding to Entry hangs the application


flagbug:
So I'm not entirely sure if this is a ReactiveUI or Xamarin.Forms bug, but I'm posting it here anyway.
When binding a string to the `Text` property of an `Entry` control, and typing some text into it, the app hangs almost immediately

Here is the stacktrace when I pause the debugger:

```
>	0x26 in System.Runtime.InteropServices.GCHandle.get_Target	C#
 	0x18 in System.WeakReference.get_Target	C#
 	0x32 in ReactiveUI.WeakEventManager<System.ComponentModel.INotifyPropertyChanged,System.ComponentModel.PropertyChangedEventHandler,System.ComponentModel.PropertyChangedEventArgs>.WeakHandler.Matches at /Users/paul/code/reactiveui/ReactiveUI/ReactiveUI/WeakEventManager.cs:279,17	C#
 	0x21 in ReactiveUI.WeakEventManager<System.ComponentModel.INotifyPropertyChanged,System.ComponentModel.PropertyChangedEventHandler,System.ComponentModel.PropertyChangedEventArgs>.WeakHandlerList.RemoveWeakHandler at /Users/paul/code/reactiveui/ReactiveUI/ReactiveUI/WeakEventManager.cs:305,21	C#
 	0x53 in ReactiveUI.WeakEventManager<System.ComponentModel.INotifyPropertyChanged,System.ComponentModel.PropertyChangedEventHandler,System.ComponentModel.PropertyChangedEventArgs>.RemoveWeakHandler at /Users/paul/code/reactiveui/ReactiveUI/ReactiveUI/WeakEventManager.cs:197,17	C#
 	0x3 in ReactiveUI.WeakEventManager<System.ComponentModel.INotifyPropertyChanged,System.ComponentModel.PropertyChangedEventHandler,System.ComponentModel.PropertyChangedEventArgs>.PrivateRemoveHandler at /Users/paul/code/reactiveui/ReactiveUI/ReactiveUI/WeakEventManager.cs:181,13	C#
 	0x61 in ReactiveUI.WeakEventManager<System.ComponentModel.INotifyPropertyChanged,System.ComponentModel.PropertyChangedEventHandler,System.ComponentModel.PropertyChangedEventArgs>.RemoveHandler at /Users/paul/code/reactiveui/ReactiveUI/ReactiveUI/WeakEventManager.cs:106,13	C#
[...]

PR #826, opened 2015-03-23
Possible fix for issue #746

nsainaney:
This took a while to find but the number of handlers grows into the thousands when using Entry.Text bound to a string. This seems to have fixed the issue and greatly speeds up ReactiveUI.

See comments on the issue thread.


Dessix:
Issue #746, for those who are interested.

paulcbetts:
@nsainaney Thanks!

Dessix:
It'd be awesome if we could get this patch available on Nuget so I can include it in my projects without including binaries in the repositories. Any chance of a minor version release?

PR #829, opened 2015-03-30
Add Universal Windows Platform (UWP) support.

jlaanstra:
This PR adds UWP support to RxUI.

## TODO:

- [x] Add activation logic on the Loading event instead of the Loaded event.
- [x] Create an ObservableForProperty class that uses the new dependency property change notification APIs.
- [ ] Fix NuGet scripts.
- [x] Figure out the WINRT vs NETFX_CORE stuff.
- [ ] Fix runtime directives (.rd.xml) files.


paulcbetts:
Can we not rename the constant? That makes the diff way more noisy than it needs to be, also WINRT is protest-via-code :P

jlaanstra:
> Can we not rename the constant?

Added as a TODO.

ghuntley:
What's the status of this PR?

paulcbetts:
@ghuntley It's up-to-date, but we're going to hold off until VS2015 hits RTM or else people checking out the source have to run RC versions of VS which sucks

jlaanstra:
I need to update it for VS2015 RC, but as @paulcbetts said, we will not merge this before VS 2015 goes RTM.

moswald:
Shouldn't this also include a new ReactiveUI.Events project for the new platform as well?

jlaanstra:
This is on my list. The changes from RC to RTM are significant so this needs quite some work still.

moswald:
As far as I can tell, once one project starts using `project.json`, then all projects need to start using `project.json`. This is even after updating to UWP Tools 1.1 where each project can have its own `{project_name}.project.json` file. This is going to take some time to clean up. :confounded:

moswald:
Unless someone has a compelling reason, I'm going to abandon this PR today and retarget it to `rxui7-master`.

ghuntley:
@moswald :+1: sounds good!

PR #831, opened 2015-03-31
Fixes #830.

### LINKED ISSUE #830: WhenAnyObservable does not always protect from NREs
Perhaps related to #769, though definitely a different bug:

```C#
    class Program
    {
        static void Main(string[] args)
        {
            var container = new Container();
            container.WhenAnyObservable(x => x.Child.Foo).Subscribe(i => Console.WriteLine(i));

            var child = new Child();
            container.Child = child;
            child.Foo.OnNext(1);

            // this results in an NullReferenceException
            container.Child.Foo = null;
            child.Foo.OnNext(2);
        }
    }

    class Container : ReactiveObject
    {
        private Child child;

        public Child Child
        {
            get { return child; }
            set { this.RaiseAndSetIfChanged(ref child, value); }
        }
    }

    class Child
    {
        public Child()
        {
            Foo = new Subject<int>();
        }

        public Subject<int> Foo { get; set; }
    }
```

kentcb:
Whenever a property chain returns a `null` observable, an instance of `Observable.Empty<TRet>` is swapped in.

Happy to rearrange as you see fit. In particular, I wasn't sure about where I should put the `ObservableExtensions` helper class.

### DIFF
--- a/ReactiveUI.Tests/ReactiveNotifyPropertyChangedMixinTest.cs
+++ b/ReactiveUI.Tests/ReactiveNotifyPropertyChangedMixinTest.cs
@@ -23,8 +23,8 @@ namespace ReactiveUI.Tests
 {
     public class TestWhenAnyObsViewModel : ReactiveObject
     {
-        public ReactiveCommand<object> Command1 { get; protected set; }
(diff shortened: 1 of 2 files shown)


paulcbetts:
Thanks!

PR #833, opened 2015-04-07
Updated the view location documentation 

pdoh00:
Include instructions on how to register multiple views to a single view model. In addition there is an example on how to change the views at run time.


ghuntley:
Thank you, I have included this change into the documentation pull request.

rdavisau:
:clap:

PR #841, opened 2015-04-24
[Android] Add ReactiveActivity ctors for JNI ownership transfer

jonfuller:
In some circumstances, the Xamarin Android framework already has a reference to an object, but needs to create it's peer MCW.  In these instances, the constructor that takes `(IntPtr, JniHandleOwnership)` is called.

This is detailed here: [Java Activation](http://developer.xamarin.com/guides/android/under_the_hood/architecture/#Java_Activation)

Also, an SO answer from @jonpryor discussing this a bit: http://stackoverflow.com/a/10603714/335675

This change adds the aforementioned constructors to both the `ReactiveActivity` and `ReactiveActivity<TViewModel>` classes.

### DIFF
--- a/ReactiveUI/Android/ReactiveActivity.cs
+++ b/ReactiveUI/Android/ReactiveActivity.cs
@@ -32,8 +32,6 @@ namespace ReactiveUI
     public class ReactiveActivity<TViewModel> : ReactiveActivity, IViewFor<TViewModel>, ICanActivate
         where TViewModel : class
     {
-        protected ReactiveActivity() { }
-
         TViewModel _ViewModel;
         public TViewModel ViewModel {
             get { return _ViewModel; }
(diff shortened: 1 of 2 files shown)


paulcbetts:
Into this, do we need this for Fragments too?

jonpryor:
> Into this, do we need this for Fragments too?

Probably. Fragments can be created by Android via Reflection unless you set [`Fragment.RetainInstance`](http://androidapi.xamarin.com/?link=P%3aAndroid.App.Fragment.RetainInstance) to `true`. If Android is creating your Fragment instance first, you need the `Fragment(IntPtr, JniHandleOwnership)` constructor.

Also, be [very careful when using generic fragments](https://bugzilla.xamarin.com/show_bug.cgi?id=28789#c1). Creating instances of generic managed types from Java is not supported.

jonfuller:
Good idea.

I can go ahead and add those constructors for the fragment classes as well, if that helps things out.

Just let me know.

Thanks!

paulcbetts:
@jonfuller Please do!

jonfuller:
Fragment constructors added!  Thanks for the input/feedback @paulcbetts and @jonpryor .

paulcbetts:
Thanks @jonfuller!

Issue #802, opened 2015-02-08
ReactiveCommand.Create method that takes an Action<object>


distantcam:
With C# 6 coming there's some nice syntax saving to be made with ViewModels.

    public ICommand MyObservableCommand => ReactiveCommand.CreateAsyncObservable(MyObservableMethod);
    public ICommand MyTaskCommand => ReactiveCommand.CreateAsyncTask(MyTaskMethod);

It would be nice if it was also possible to do this with commands that just take an `Action<object>` as well.

    public ICommand MyCommand => ReactiveCommand.Create(MyMethod);

kentcb:
This seems dangerous to me, since every access to the property would be creating a new command instance. If I have this ctor code in my VM:

```C#
this.MyCommand
    .Subscribe(DoSomeOtherThingWhenTheCommandIsExecuted);
```

And I have this in my view:

```C#
this.BindCommand(ViewModel, x => x.MyCommand, x => x.someButton);
```

It won't work as expected. The binding infrastructure will obtain a different instance than that used in my VM's ctor.

distantcam:
@kentcb Ack! That's true. I took the optimisation a step too far and should have left it at this.

    public ICommand MyCommand { get; } = ReactiveCommand.Create(MyMethod);

kentcb:
Well, waddaya know. #971 does exactly this by _requiring_ execution logic be provided as creation time. It will be merged shortly.

Issue #769, opened 2014-12-02
WhenAnyObservable doesn't unsubscribe on child null.


grokys:
I'm not sure if this is intended behaviour or not, but if an observable on a child object is subscribed to using WhenAnyObservable and then the child object is removed, the observable continues to fire.

The following code demonstrates this:

```
using System;
using System.Reactive.Subjects;
using ReactiveUI;

namespace ReactiveUI_WhenAnyObservable_Bug
{
    class Program
    {
        static void Main(string[] args)
        {
            var container = new Container();
            container.WhenAnyObservable(x => x.Child.Foo).Subscribe(i => Console.WriteLine(i));

            var child = new Child();
            container.Child = child;
            child.Foo.OnNext(1);
            container.Child = null;
            child.Foo.OnNext(2);
        }
    }

    class Container : ReactiveObject
    {
        private Child child;

        public Child Child
        {
            get { return child; }
            set { this.RaiseAndSetIfChanged(ref child, value); }
        }
    }

    class Child
    {
        public Child()
        {
            Foo = new Subject<int>();
        }

        public Subject<int> Foo { get; private set; }
    }
}
```

If you run this, you'll see both 1 and 2 output, despite by the time `child.Foo.OnNext(2)` was called, `child` is no longer a child of `container`

Is this a bug? If not, is there a work-around?

grokys:
Any change of this getting fixed? Or if you could point me in the right direction I could take a look.

paulcbetts:
@grokys You can have a look at the implementation of WhenAnyObservable, it's pretty straightforward. You can probably write an alternate one in your app, it doesn't require any private methods

kentcb:
FYI, the problem is actually in `ObservableForProperty`. It's not notifying in this scenario:

```C#
        [Fact]
        public void NAME_THIS_SUCKER()
        {
            var fixture = new TestWhenAnyObsViewModel();
            var output = fixture.ObservableForProperty(x => x.Child.MyListOfInts).CreateCollection();

            Assert.Equal(0, output.Count);

            fixture.Child = new TestWhenAnyObsViewModel
            {
                MyListOfInts = new ReactiveList<int>()
            };

            Assert.Equal(1, output.Count);

            fixture.Child = null;
            Assert.Equal(2, output.Count);
        }
```
The final assertion is failing because there is 1 item in `output`, not 2.

[...]

Issue #710, opened 2014-08-26
IReactiveCommand and Subscribe


ghost:
Is the way to use 
public IReactiveCommand ShowMapCommand { get; set; }

and Subscribe() extension method?

In ReactiveUI 5 it works but not in version 6.

PR #850, opened 2015-05-11
Ship ReactiveUI 7.0

paulcbetts:
This is the main :shipit: PR for what will become ReactiveUI 7.0
## Features / TODO:
- [x] Drop legacy platforms https://github.com/reactiveui/ReactiveUI/issues/945
- [x] Get rid of unused features on bindings like Implicit binding
- [x] ReactiveCommand properties should all be Behaviors
- [x] Figure out how to make ToProperty fast but also not troll people
- [x] Universal Windows Platform (UWP) support. #1045
- [x] Clean up the nuget package description: https://github.com/reactiveui/ReactiveUI/issues/1047


paulcbetts:
https://github.com/reactiveui/ReactiveUI/issues/802 is a Neato Thing(tm)

grokys:
It would be nice to get #769 fixed. I've not managed...

ghuntley:
Can we please drop the `IReactiveCommand` interface.

See Issue #710:

> Don't use IReactiveCommand for anything, always declare the concrete type. 
> Should've removed that in RxUI 6.0 :-/

kmcginnes:
It'd be nice if the user could pull in some of the core magic of RxUI without any of the _fluf_. Perhaps this is wishful thinking on my part, but a slimming down of RxUI core would ease integration with other MVVM frameworks, and even make it easier to provide a .Net 4.0 release for those of us stuck supporting WinXP clients.

I was thinking the core features would be:

* this.WhenAny, this.WhenAnyValue, this.WhenAnyObservable
* ObservableAsPropertyHelper, ToProperty
* ReactiveObject, this.RaiseAndSetIfChanged
* ReactiveCommand, ReactiveAsyncCommand
* ReactiveList

Things not included in this core:

* Splat
* Any IoC
* Any logging frameworks or extension methods
* MessageBus
* ViewLocator, IViewFor

__Note__ I have used both RxUI 6.0 and 4.0 recently and I may have some features confused between them. Please excuse me if I listed some things that are no longer included in the main RxUI library.

mteper:
:+1: I've more than once wanted to used some of the RxUI bits in completely non-UI scenarios, but didn't want to add the extra dependencies.

csunwold:
:+1: Great idea @kmcginnes

paulcbetts:
@kmcginnes Removing Splat isn't possible because RxUI itself relies on Splat to implement the features you mention in the Core list (especially WhenAny). 

I think this problem indicates a perception / education issue than an actual code issue - if you don't want to use any of the features in the "not included" list, you can just completely ignore them, none of them are forced on your application.

jcmm33:
Any thoughts on implementation of a new validation model/approach, a modern day successor to ReactiveValidatedObject.

kmcginnes:
@paulcbetts So my suggestion was *actually* a really big complaint wrapped up along side some other smaller complaints I had. I've had this similar issue in other libraries/frameworks.

In ReactiveUI there is an extension method witch is phenomenal on its own, but conflicts with how I want to use it. Creating my own extension method, at least with the same name, is now out of the question in classes that include the ReactiveUI namespace.

That extension method is `this.Log()`.

This concept was [introduced](http://www.codeproject.com/Articles/530808/Introducingplusthis-Log) by Rob Reynolds in early 2013. I'm not sure who predated whom, but I learned it from Rob. It has become such a hit with me that I use a custom version of it in all my apps.

However, when using ReactiveUI, if I'm in the view model or view I can't use my custom `this.Log()` extension method because it conflicts with ReactiveUI's. Sure I can rename **my** extension method to something else, or I could call it statically. Those are options, but.... REALLY?

So my original intent with the suggestion to create a ReactiveUI core was to reduce the possibility of ReactiveUI opinions leaking out into it's consumer's code bases. This might've been fine in the first few releases, but as the user base grows this becomes a bigger and bigger issue.

While NuGet is far from perfect, it does allow for the easy consumption of optional additions to a library. Therefore giving the choice to the user, rather than including everything that a user may possibly use in a single package.

I hope this rant wasn't too, uh, rant-y. I just had to get that off my chest.

paulcbetts:
@kmcginnes Nope, ReactiveUI had this almost exact API a full two years before that article came out :)  https://github.com/reactiveui/ReactiveUI/blob/b00354decf7c97e5af1148a6693136935fb82d22/ReactiveUI/Logging.cs#L78

martingemme:
@paulcbetts Could the task link also include WP Universal 8.1 Support or Samples?

ghuntley:
Let's include https://github.com/reactiveui/ReactiveUI/pull/599 as well?

moswald:
How do you guys feel about finally dropping the ReactiveUI-Mobile package? It was deprecated for 6 (or earlier?), so it's just muddying the waters now.

shiftkey:
@moswald :fire: :fire: :fire:

moswald:
:fire:!

ghuntley:
ReactiveUI-Mobile must go - https://github.com/reactiveui/ReactiveUI/pull/850#issue-75053445

ghuntley:
How would you guys feel about closing down the addition of new features to RxUI 7 after a particular date. ie. define scope, prevent increase and define release theme.

* First community release under the careful supervision of our benevolent dictator.
* New ReactiveCommand
* UWP support
* Building off CI
* Releasing pre packages to nuget
* Bugfixes

For example I know Kent wants to have a stab with someone else as well at potentially looking at improving routing but that's definitely a v8 feature.

kentcb:
Dates don't work for me (and probably never will on open source work), but agreeing on features and fixes does.

ghuntley:
Features and fixes by a particular date (ie. discuss breaking is useful)

Completely agree, we are all volunteers here and contribute where we can. It's already a second job in itself and deadlines suck the spirit out of it. The only date I have in mind is "it would be nice to have v7.0" by Evolve - 24th of April.

PR #856, opened 2015-05-18
Fixes #846.

### LINKED ISSUE #846: Remove OneWayBind implicit resolution of view property type
There are several related issues here:

1. *Reflection.cs* utilizes `GetRuntimeField` and `GetRuntimeProperty` in its `getViewExpression` method. The behavior of these methods has changed in FX4.6 such that they no longer return private members.
2. *PropertyBinding.cs* uses `GetDeclaredMethod` in its `OneWayBind<TViewModel, TView, TVMProp, TVProp>` method. However, there are multiple matches for the passed in method, so it will throw due to the ambiguity.
3. In the same method as (2), there is a check to see whether `tvProp != typeof(TVProp)`. If not, a dynamic method invocation and cast is performed. However, the cast can never succeed if `tvProp != typeof(TVProp)`.

**UPDATE**: after discussing this on Slack, it seems that it is probably best to just remove implicit view property resolution altogether. It results in opaque, more fragile, less performant code. Oh, and it probably hasn't worked for a long time anyway.

kentcb:
This PR fixes #846 by removing `IDefaultPropertyBinderProvider` and everything that depends on it, including the overloads of `Bind`, `OneWayBind`, and `BindCommand`.

Note that I have fixed up project files for all platforms, but am unable to verify all builds.


paulcbetts:
Can you retarget this branch toward `rxui7-master`? Unfortunately this means you have to close it and reopen :-(   Also, can you make the PR titles more descriptive? As long as you have "Fixes #XYZ in the PR *body*, it'll do what you want"

PR #857, opened 2015-05-18
Removes support for implicitly determined view properties in bindings

### LINKED ISSUE #846: Remove OneWayBind implicit resolution of view property type

kentcb:
This PR fixes #846 by removing `IDefaultPropertyBinderProvider` and everything that depends on it, including the overloads of `Bind`, `OneWayBind`, and `BindCommand`.

Note that I have fixed up project files for all platforms, but am unable to verify all builds.


CommonAssemblyInfo.cs
```diff
@@ -1,4 +1,4 @@
 using System.Reflection;
 
-[assembly: AssemblyVersion("6.5.0")]
-[assembly: AssemblyFileVersion("6.5.0")]
+[assembly: AssemblyVersion("6.90.0")]
```
jlaanstra: Can you revert this file? It is not really part of the change made here.

jlaanstra: Nevermind noticed that it comes from the merge. Can you retarget this to rxui7-master?

kentcb: Hmm, I thought I did that, but maybe I did something wrong. How exactly should I do so?

jlaanstra: Before clicking create it should give you the option to switch the target branch.

kentcb: Ah, I misunderstood what Paul was asking. OK, should be good now but I still see lots of commits that aren't mine.

PR #873, opened 2015-06-07
Ensure batched changes result in BeginUpdates call ASAP

### LINKED ISSUE #867: Latent application of batched changes can cause data synchronization issues
On iOS, when using `ReactiveTableViewSource`, changes are collated and applied as a batch. There may be a short delay between the detection of the changes, and their application.

If the application code is calling `BeginUpdates` / `EndUpdates` against the table view, this can cause iOS to see an inconsistent view of the data, resulting in the infamous "Invalid update: invalid number of rows in..." message.

The reason application code might be calling `BeginUpdates` / `EndUpdates` itself is to trigger animations when, for example, cell layout changes are made. Here's a real-world example:

```C#
private void AnimateDetailsIn()
{
    if (this.detailView.Superview != null)
    {
        return;
    }

    this.ContentView.AddSubview(this.detailView);

    this.ContentView.ConstrainLayout(
        () =>
            this.detailView.Top() == this.nameLabel.Bottom() + Layout.HalfSiblingViewSpacing &&
            this.detailView.Bottom() == this.ContentView.Bottom() - Layout.StandardSiblingViewSpacing &&
            this.detailView.Left() == this.nameLabel.Left() &&
            this.detailView.Right() == this.nameLabel.Right());

    var tableView = this.FindParent<UITableView>();

    if (tableView == null)
    {
        return;
    }
            
    // this triggers an animation between the old state of the cell, and the new
    tableView.BeginUpdates();
    tableView.EndUpdates();
}
```

The idea here is to animate in a details panel when the user selects a cell. Similar logic animates it out.

The `BeginUpdates` / `EndUpdates` calls at the end can result in an exception if there are outstanding changes not yet applied by `CommonReactiveSource`.

I believe the solution for this is simple for table views, but looks to be impossible for collection views. 

For table views, we can simply have the `CommonReactiveSource` call `BeginUpdates` as soon as changes are detected (not in a later message). Then, once applied, `EndUpdates` is called. Since these methods can be nested, any number of calls can be made by application code in the interim with no ill effect.

Collection views, on the other hand, do not support `BeginUpdates` / `EndUpdates` (inexplicably, in my opinion). I'm not entirely sure where this leaves us yet - will need to investigate more.

I will PR this when I get time.

kentcb:
This PR fixes #867 by calling `BeginUpdates` as soon as changes are detected rather than waiting until those changes are processed. As expained in the issue, this is critical to avoid race conditions where application code is also calling `BeginUpdates` / `EndUpdates`.


paulcbetts:
Thanks!

PR #875, opened 2015-06-08
Relaxing ToProperty's generic constraints from ReactiveObject to IReactiveObject

kentcb:
This PR fixes the generic constraints for `ToProperty` such that only an `IReactiveObject` is required - not a `ReactiveObject`.


paulcbetts:
Thanks!

PR #879, opened 2015-06-13
Provide ViewContract property on ViewModelViewHost for XAML platforms

### LINKED ISSUE #870: Add scalar ViewContract property to ViewModelViewHost implementations
The inability to specify a view contract as a `string` causes headaches for certain scenarios. For example, suppose you have two different views for a backing VM: one primary view (no contract) and one for list items (contract is `"listitem"`). Now suppose I want to use my "listitem" view from XAML:

```XML
<ListBox ...>
    <ListBox.ItemTemplate>
        <DataTemplate>
            <rxui:ViewModelViewHost ??? />
        </DataTemplate>
    </ListBox.ItemTemplate>
</ListBox>
```

There is no easy way to specify the contract name in this case because:

* `ViewModelViewHost` requires an observable for the view contract, and we can't easily create one in XAML
* naming our `ViewModelViewHost` and then configuring `ViewContractObservable` in the code behind is not an option because our `ViewModelViewHost` is inside a `DataTemplate`. Thus, we cannot gain easy access to it from the code behind (because which `ViewModelViewHost` within which `ListBoxItem` would we be accessing?)

The only real recourse is to hook up an event:

```XML
<rxui:ViewModelViewHost ViewModel="{Binding}" Initialized="ViewModelViewHost_Initialized"/>
```

And stick this in the code-behind:

```C#
private void ViewModelViewHost_Initialized(object sender, System.EventArgs e) =>
            ((ViewModelViewHost)sender).ViewContractObservable = Observable.Return("listitem");
```

This is rather messy.

Therefore, I propose that `ViewModelViewHost` includes a `ViewContract` property of type `string`. The setter for this property would simply forward onto `ViewContractObservable`, passing through `Observable.Return(value)`.

This does create the possibility of both `ViewContract` and `ViewContractObservable` being set. If this happens, I would suggest a simple last-in wins resolution. Otherwise, extra state needs to be kept around just for the purposes of raising an error (to determine whether `ViewContractObservable` was set directly or via `ViewContract`).

With such a change, we could then do this:

```XML
<rxui:ViewModelViewHost ViewModel="{Binding}" ViewContract="listitem"/>
```

I'd also like to get rid of the `ViewModel` binding, but that might be a story for another issue.

kentcb:
This PR fixes #870 by adding a `ViewContract` property (of type `string`) to the `ViewModelViewHost` implementation on XAML platforms. The idea is to simplify use cases such as:

```XML
<ListBox x:Name="listBox">
    <ListBox.ItemTemplate>
        <DataTemplate>
            <rxui:ViewModelViewHost ViewContract="listitem" ViewModel="{Binding .}"/>
        </DataTemplate>
    </ListBox.ItemTemplate>
</ListBox>
```

Previously this would require some pretty ugly hackery to get the appropriate view contract in use by the `ViewModelViewHost`.

**NOTE:** I only implemented this for XAML platforms because it seems of higher importance there. Also, I'm not really set up to build other platforms yet so I didn't really want to risk committing broken code.


paulcbetts:
Thanks!

Issue #758, opened 2014-11-16
LinkLabel defines Events too?


afrentis:
Does ReactiveUI.Events work for WinForm controls? 
i.e., 
this.SomeLinkLabel.Events().LinkClicked.InvokeCommand(this,x=>x.VM.LinkClicked);

If so, what references & namespaces are required to get it to work?

I can't seem to get Events() to be properly recognized. 

Thanks.

PR #888, opened 2015-06-26
Some event building cleanups

gluck:
- Replaced FromEventPattern by FromEvent with explicit conversion
 - strongly-typed from end-to-end, no more reflection/runtime errors
- No longer maps events with more than 2 args (would've failed at runtime
 anyway)
- {}Events types now only extends the closest parent that gets built in
 the assembly
 - allows to generate event assembly for Winforms w/o referencing System.dll
 - cf #758
- {}Events types now exposes inherited events as well (up to a parent event
 type)

Overall the effective changes on generated events (for Net45) are fairly small:
- HwndHostEvents is removed (its observable wasn't working anyway)
- types inheriting ObservableCollection now get a CollectionChanged observable:
 - CalendarBlackoutDatesCollection, GridViewColumnCollection, SelectedDatesCollection

### DIFF
--- a/ReactiveUI.Events/EventBuilder.cs
+++ b/ReactiveUI.Events/EventBuilder.cs
@@ -59,13 +59,6 @@ namespace EventBuilder
 
         public static NamespaceInfo[] CreateEventTemplateInformation(AssemblyDefinition[] targetAssemblies)
         {
-            var publicTypesWithEvents = targetAssemblies
-                .SelectMany(x => SafeGetTypes(x))
-                .Where(x => x.IsPublic && !x.HasGenericParameters)
-                .Select(x => new { Type = x, Events = GetPublicEvents(x) })
-                .Where(x => x.Events.Length > 0)
-                .ToArray();
-
             var garbageNamespaceList = new[] {
                 "Windows.UI.Xaml.Data",
                 "Windows.UI.Xaml.Interop",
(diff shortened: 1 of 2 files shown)


jlaanstra:
> •Replaced FromEventPattern by FromEvent with explicit conversion ◦strongly-typed from end-to-end, no more reflection/runtime errors


I don't think we should change FromEventPattern to FromEvent, since these events all use the standard .NET pattern. https://social.msdn.microsoft.com/Forums/en-US/5f4df41c-6df1-42be-b7bb-bbf072143c24/observablefromevent-throws-in-v1010425?forum=rx

Also we already do not use reflection because of the overload we use.

gluck:
*Most* of these events indeed use the std, only [HwndHost.MessageHook](https://msdn.microsoft.com/en-us/library/system.windows.interop.hwndsourcehook(v=vs.110).aspx) (for WPF scope) doesn't, and the [generated event](https://www.symbolsource.org/Public/Metadata/NuGet/Project/reactiveui-events/6.5.0/Release/.NETFramework,Version=v4.5/ReactiveUI.Events/ReactiveUI.Events/Events.cs):

        public IObservable<System.Int32> MessageHook {
            get { return Observable.FromEventPattern<System.Windows.Interop.HwndSourceHook, System.Int32>(x => This.MessageHook += x, x => This.MessageHook -= x).Select(x => x.EventArgs); }
        }

Isn't correct and will fail at runtime AFAIK.

Use of FromEventPattern is useless anyway because we drop the generated [EventPattern](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Rx.NET/Source/System.Reactive.Linq/Reactive/EventPattern.cs) object to only select its EventArgs (see the end of the above line for the event).
The FromEvent version doesn't generate these intermediate objects.

Also to the best of my knowledge the current overload used *do* make use of reflection (because no conversion is provided), see [here](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/FromEventPattern.cs#L44)

Thx !

gluck:
Did a benchmark (results [here](https://gist.github.com/gluck/da82c1b537e44b55daa8)) to prove the perf increase (more than *3X* on the subscription path).

gluck:
Abandoned since rxui-events has been re-worked. Will try to look how it'd fit in the new one.

Issue #814, opened 2015-02-25
ObservableAsPropertyHelper calls Connect twice with cold observables


TheGrandUser:
If you call ToProperty on a cold observable (in my case, BehaviorSubjects) that is the backing for a property bound to WPF, it will connect to the source observable twice.

I think the root problem is that when Connect is called it is raising the PropertyChanged event before it returns and the _inner subscription is assigned.

JKronberger:
We are seeing the same thing happen. 

I set up this code to debug.

    var cl = Observable
                .CombineLatest
                ( peelingToolObservable
                    , selectedElementObservable
                    , (a, b) =>
                    {
                        return new Assembly(a, b);
                    });

            var cl2 = Observable.Create<Assembly>(observer =>
            {
                // Set breakpoint HERE
                return cl.Subscribe(observer);
            });

            _CompletePeelingToolAssembly = cl2
                //.Publish().RefCount()
                .ToValidatingProperty(this, p => p.CompletePeelingToolAssembly, new Assembly());

and set a breakpoint at the place as indicated. We get two breakpoints occuring. Here are the backtraces for both of them.

__First__
![image](https://cloud.githubusercontent.com/assets/3232716/8401664/2053ca44-1e32-11e5-9e1f-4f2135a55af4.png)

__Second__
![image](https://cloud.githubusercontent.com/assets/3232716/8401668/2efe3322-1e32-11e5-9cdb-d71a70a6c7c5.png)

We have a workaround by putting ``Publih().Refcount`` at the end before calling but it is not ideal.

JKronberger:
Ok I have a test case showing the failure. 

	using System.Reactive;
	using System.Reactive.Disposables;
	using System.Threading;
	using System.Threading.Tasks;
	using System.Windows.Threading;
	using FluentAssertions;
	using System;
	using System.Reactive.Linq;
	using Splat;
	using Xunit;
	
	namespace ReactiveUI.Ext.Spec
	{
	
	    public class ToValidatingPropertySpec
	    {
	        public class Fixture : ReactiveObject
	        {
	            public ObservableAsPropertyHelper<bool> _A;
	
	            public bool A
	            {
	                get { return _A.Value; }
	            }
	        }
	
	        /// <summary>
	        /// Start a new dispatcher and push the func to
	        /// that dispatcher and async wait for the result
	        /// </summary>
	        /// <typeparam name="T"></typeparam>
	        /// <param name="func"></param>
[...]

PR #894, opened 2015-07-06
Fixes issue #814 Race condition to ObservablePropertyHelper

bradphelan:
Using multicast I think was the wrong solution as there was no real need to multicast. A simpler solution which I have used in the past is to create a BehaviourSubject<IObservable<T>>() initialized with an Observable.Never<T>(). 

I've created a test case to show the previous failure and it now passes with the changed code.


bradphelan:
Fixes #814

bradphelan:
@paulcbetts Any chance to have this integrated. It's got test cases and all.

jlaanstra:
I am not entirely convinced yet that we should take this change. The actual fix seems to be the _activated flag in the getter of Value. The rest is just refactoring of code.

To me it seems weird that to setup property A, you somehow need the value of property A.

bradphelan:
@jlaanstra It's not really wierd. It's just lazy initialization. It avoids subscription to the underlying observable unless the value is actually ever used. It's possibly an over optimization but that is what the original code was attempting to do but failed to implement correctly. Using an initial sequence of Observable.Never and switching it to the real sequence on init is more robust than the previous code.

bradphelan:
@paulcbetts @jlaanstra Guys! I have a pull request here with a test case showing a verified bug and that the code fixes the error but I'm getting no feedback. What can I do to get this pull request accepted?

gluck:
Not having any authority here, but here's my 2 cents advice about this change.
While the issue is real (reentrancy on `ObservableAsPropertyHelper::get_Value` causes duplicate subscription), I agree with @jlaanstra that the proposed fix goes beyond the scope.

I'll suggest some edits in comments.

ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
```diff
@@ -132,4 +138,79 @@ public void ToPropertyShouldFireBothChangingAndChanged()
...
+            }
+        }
+
+        [Fact]
+        public void ToPropertyShouldSubscribeOnlyOnce()
+        {
+
+
+                var count = 0;
+                var obs = Observable.Create<bool>(observer =>
```
gluck: can be simply:
```
     Observable.Return(true).Do(_ => count++);
```

ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
```diff
@@ -132,4 +138,79 @@ public void ToPropertyShouldFireBothChangingAndChanged()
...
+                    // This triggers the property change firing
+                    // upon subscription in the ObservableAsPropertyHelper
+                    // constructor.
+                    observer.OnNext(true);
+                    return Disposable.Empty;
+                });
+
+                var f = new Fixture();
+
+                f._A = obs.ToProperty(f, t => t.A, false);
```
gluck: Maybe prefer using the recommended way of declaring properties ? i.e. move it to the VM class and use:
```
Observable.Return(true).Do(_ => _count++).ToProperty(this, x => x.A, out _A);
```

ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
```diff
@@ -132,4 +138,79 @@ public void ToPropertyShouldFireBothChangingAndChanged()
...
+
+                // This line is important because it triggers connect to
+                // be called recursively thus cause the subscription
+                // to be called twice. Not sure if this is a reactive UI
+                // or RX bug.
+                f.PropertyChanged += ( e, s ) => Console.WriteLine(f.A);
+
+                var v = f.A;
```
gluck: Add an AssertEquals to avoid the unused variable ?

ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -112,10 +114,17 @@ public sealed class ObservableAsPropertyHelper<T> : IHandleObservableErrors, IDi
         /// </summary>
         public IObservable<Exception> ThrownExceptions { get; private set; }
 
+        private void Activate()
+        {
+            if ( _activated )
```
gluck: I would use Interlock together with a SingleAssignmentDisposable to guarantee both single subscription and disposal, keeping the previous logic unchanged:
```
                if (Interlocked.CompareExchange(ref _activated, 1, 0) == 0) {
                    _disp.Disposable = _source.Connect();
                }
```

bradphelan:
Thanks for the feedback. I'll make some updates and resubmit

bradphelan:
Thanks @gluck for the feedback. I've reverted the changes and applied the fix only to the Value.get as you suggested. I've also cleaned up the test case as per your suggestions. I've checked the test case still fails without the fixes applied and that it passes with the fixes applied.

ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
```diff
@@ -132,4 +138,78 @@ public void ToPropertyShouldFireBothChangingAndChanged()
             Assert.Equal("Baz", resultChanged[1].Value);
         }
     }
+
+    class UnitTestMode : IModeDetector
```
gluck: pretty minor but this could be a IDisposable helper so one'd write the test like that (so that all test scope is within the test method):

    using (new ProductionMode()) {
        ...
    }

also minor but I believe you can use new PlatformModeDetector() when resetting the mode (it's the default).

bradphelan:
@gluck Changes on your latest suggestions have been made. We would really appreciate this pull request being merged with the next release. We see side effects of this bug throughout our application.

RxUI.DotSettings
```diff
@@ -1,18 +1,7 @@
...
-	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/UserRules/=9b2b4c09_002Dbe0f_002D4b8e_002D8f1a_002Dcdccf074d45c/@EntryIndexedValue">&lt;Policy&gt;&lt;Descriptor Staticness="Static, Instance" AccessRightKinds="Private" Description="Private Fields/Methods/Properties"&gt;&lt;ElementKinds&gt;&lt;Kind Name="FIELD" /&gt;&lt;Kind Name="READONLY_FIELD" /&gt;&lt;Kind Name="CONSTANT_FIELD" /&gt;&lt;Kind Name="METHOD" /&gt;&lt;Kind Name="PROPERTY" /&gt;&lt;Kind Name="EVENT" /&gt;&lt;/ElementKinds&gt;&lt;/Descriptor&gt;&lt;Policy Inspect="True" Prefix="" Suffix="" Style="aaBb" /&gt;&lt;/Policy&gt;</s:String>
-	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
\ No newline at end of file
+<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
gluck: change made on purpose ?

bradphelan: I reverted it in the next commit.

gluck: seems to me that you reverted more than your original modif, I still see this file changed on the overall diff

bradphelan:
@gluck  It's reverted correctly now.

gluck:
LGTM, now we need to get a :shipit: !

eggapauli:
@kentcb Can we merge this to reactive-ui. This bug is killing me in production!

kentcb:
@eggapauli : it's targeting the wrong branch (needs to be rxui7-master). Any chance @bradphelan could re-target?

PR #901, opened 2015-07-15
Remove exception logging on notify observable

ds-b:
By doing this we will avoid having an exception handled/logged twice


mteper:
@ds-b could you clean up whitespace, please?

ds-b:
@mteper sure, I can do that

mteper:
@ds-b Eeek! I think you've gone a bit overboard there, don't you think? :)

I am sorry if I wasn't clear, but I was only requesting that you fixed up white space in the file your original changes appeared in, so that it would be easier to see what your change actually meant to be.

ds-b:
@mteper seems so :)

ds-b:
it should be ok now, thanks!

kentcb:
I know it's been ages, @ds-b, and for that I apologise.

However, am gonna need a unit test or something before I can understand/merge this. We already have [this test](https://github.com/reactiveui/ReactiveUI/blob/006e41bef02c4785d2929cab2f9265ede682fb1f/src/ReactiveUI.Tests/ReactiveObjectTest.cs#L210) which passes with or without the removed line. Therefore, it seems to me you should only be getting 1 notification regardless.

ds-b:
Actually it may helpful to have both logging and observable stream for errors, so I will close the pull request. Thanks!

PR #904, opened 2015-07-17
Ability to override view model type used in view resolution using a  view model attribute

patsissons:
I am creating this PR because I ran into an issue yesterday where I wanted to make a generically typed view model that would all use a common view. The problem with this approach is that you cannot register your IViewFor<> associations because the generic type doesn't allow that (this might work if you use an IoC that supports generic registrations, I'm not entirely sure). The simplest solution I could think of was to allow overriding the type information used in view resolution by using a custom attribute. This attribute then supplies the type override to use in view resolution (and does not even necessarily have to be related to the view model it decorates, but that would probably be a strange setup).

Here are some examples where we could use this new attribute
```csharp
[ViewResolutionTypeOverride(Type = typeof(IMyViewModel))]
public class MyGenericViewModel<T> : ReactiveObject, IMyViewModel
{
    // ...
}

public class MyBaseViewModel : ReactiveObject
{
    // ...
}

[ViewResolutionTypeOverride(Type = typeof(MyBaseViewModel))]
public class MyDerivedGenericViewModel<T> : MyBaseViewModel
{
    // ...
}

// and in the view registrations
public void RegisterViews()
{
    Locator.CurrentMutable.Register(() => MyView(), typeof(IViewFor<IMyViewModel>));
    Locator.CurrentMutable.Register(() => MyOtherView(), typeof(IViewFor<MyBaseViewModel>));
}
```

p.s. if a  test is required for this PR to be accepted let me know and I'll look into writing a new test for it.

### DIFF
--- a/ReactiveUI.Tests/RxRouting.cs
+++ b/ReactiveUI.Tests/RxRouting.cs
@@ -34,6 +34,14 @@ namespace Foobar.ViewModels
             HostScreen = hostScreen;
         }
     }
+
+    [ViewResolutionTypeOverride(Type = typeof(BazViewModel))]
+    public class BazViewModel<T> : BazViewModel
+    {
+        public BazViewModel(IScreen hostScreen) : base(hostScreen)
+        {
+        }
+    }
 }
 
 namespace Foobar.Views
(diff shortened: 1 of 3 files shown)


paulcbetts:
I'm not sure that we need this because we have the `contract` parameter, does that work?

patsissons:
I couldn't think of a way to make the contract work elegantly. Do you have
an example I could take a look at?

moswald:
@patsissons:

```csharp
[ViewContract(Contract = "Detailed")]
class MyDetailedView : IViewFor<MyViewModel>

[ViewContract(Contract = "Summary")]
class MySummaryView : IViewFor<MyViewModel>
```

See [this completed PR](https://github.com/reactiveui/ReactiveUI/pull/879) for one way you might put it to use.

patsissons:
I'm not sure if this fully solves the issue I was facing way back then, but I don't really use RxUI these days so I can't really say for sure. That said I can probably close this PR now since it has become so stale.

patsissons:
closing this as it has become way too stale.

Issue #906, opened 2015-07-24
API for reactive commands can be improved


kwesolowski:
I file it as issue to discuss some proposal. We created (for inhouse use;)) helper class `ReactiveCommandEx`, with following methods:

```
public static ReactiveCommand<Unit> CreateAlwaysAvailable(Action action);
public static ReactiveCommand<Unit> CreateAlwaysAvailableAsyncTask(Func<object, Task> action)
public static ReactiveCommand<Unit> CreateAlwaysAvailableAsyncTask(Func<Task> action)
public static ReactiveCommand<Unit> CreateAsyncTask(IObservable<bool> canExecute, Func<Task> action)

public static ReactiveCommand<Unit> CreateAsyncTask(IObservable<bool> canExecute, Func<object, Task> action)
public static ReactiveCommand<Unit> Create(IObservable<bool> canExecuteObservable, Action<object> action)
public static ReactiveCommand<Unit> Create(IObservable<bool> canExecuteObservable, Action action)
```

Which are simply wrappers for `ReactiveCommand` methods, ensuring that:

1. `ReactiveCommand<Unit>` is consistently used as return type,
2. When command is always available client code is shorter (no Observable.Return(true) everywhere :)).
3. When `Func<Task>` throws exception instead of returning faulted task it is properly converted to observable and is pushed to `ThrownExceptions`.

Is it worth integrating such functionality into ReactiveUI? If so I will create PR.

mteper:
Could you give an example of before and after code?

kwesolowski:
@mteper Does are mostly additions wrapping existing functionality, which code (client or library) you would like to see?

mteper:
@kwesolowski just trying to understand the problem you are solving as it's not ringing a bell. So curious what your app code looks like without the helpers and then with.

kwesolowski:
AD1: It is just minor annoyance - `ReactiveCommand.Create` creates `ReactiveCommand<object>` so if you change implementation to CreateAsyncTask you have to change type. Guidelines state that VM state should be used anyway (not object passed by command)).

AD2. it simply allows to replace:
`ReactiveCommand.CreateAsyncTask(Observable.Return(true), _ => actionAsync());` with:
`ReactiveCommandEx.CreateAlwaysAvailableAsyncTask(_ => actionAsync());` (with code completion this is faster to type:)).

[...]

PR #910, opened 2015-07-29
Added test case illustrating problem with task failing "too early".

kwesolowski:
Below is test case illustrating problem described at number 3 in discussion:
https://github.com/reactiveui/ReactiveUI/issues/906

Commented out Task implementation work fine/passes, current implementation causes Task to become in "cannot execute" state without providing error to `ThrownExceptions`

Multiple existing methods/API in the world do parameter validation "before" actually starting asynchronous operations (sometime separate parameter validating wrapper is public, and actual `async` implementation is private for example).

I was going to dive deeper and fix it - but would like to know if you consider it "bug" or I am missing sth :)

### DIFF
--- a/ReactiveUI.Tests/ReactiveCommandTest.cs
+++ b/ReactiveUI.Tests/ReactiveCommandTest.cs
@@ -571,6 +571,29 @@ namespace ReactiveUI.Tests
         }
 
         [Fact]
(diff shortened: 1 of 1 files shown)


kentcb:
See comments in #906 - this is mostly irrelevant now with the new RxCommand.

kwesolowski:
Sure - I will check when i get chance to fiddle with RxUI 7 :)

PR #926, opened 2015-09-01
Avoid hash calculations where possible

kentcb:
This PR avoids potentially expensive hash calculations for items in a `ReactiveList`. Previously, the code always output each item's hash to the log at info level. This PR changes the level to debug and only performs the log call (and hence the hash calculation) if the log level is debug.


paulcbetts:
Just delete the line altogether

kentcb:
Fair enough - wasn't sure whether it was somewhat useful information to someone, somewhere :)

Done.

PR #931, opened 2015-09-10
Fix events package for XamForms

### LINKED ISSUE #881: ReactiveUI-Events package does not support Xamarin Forms portable
The reactiveui-events package does not install into a portable Xamarin Forms project:

> Could not install package 'reactiveui-events 6.5.0'. You are trying to install this package into a project that targets 'portable-net45+win+wp80+MonoTouch10+MonoAndroid10+xamarinmac20+xamarinios10', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

It seems to me it should be possible to support this, since Xamarin.Forms itself has portable support.

kentcb:
This PR fixes #881 by fixing the project file and adding the requisite file stubs for the XamForms events package.


kentcb:
gfdi, wrong base branch

PR #936, opened 2015-09-10
Ensure ObservableAsPropertyHelper provides initial value immediately

### LINKED ISSUE #815: ObservableAsPropertyHelper's latent provision of initial value is thwarting skipping of initial value
When using any scheduler other than immediate/current thread, `ToProperty` will fire its initial value latently, which prevents `ObservableForProperty(..., skipInitial: true)` working as expected. Instead of skipping the initial value, one still gets it.

Here's an example to show what I mean. I couldn't do this as a unit test because XUnit is just refusing to work for me at the moment:
```C#
public class VM : ReactiveObject
{
    private readonly ObservableAsPropertyHelper<bool> someProperty;

    public VM()
    {
        this.someProperty = Observable
            .Never<bool>()
            .ToProperty(this, x => x.SomeProperty, initialValue: false, scheduler: RxApp.MainThreadScheduler);
    }

    public bool SomeProperty => this.someProperty.Value;
}

public class AltView : UIViewController
{
    public AltView()
    {
        var vm = new VM();
        vm
            .ObservableForProperty(x => x.SomeProperty, skipInitial: true)
            .ObserveOn(RxApp.MainThreadScheduler)
            .Subscribe(x => Debug.WriteLine("WTF?", x.Value));
    }
}
```
This currently outputs "WTF?" when it shouldn't. Changing the scheduler passed into `ToProperty` to either immediate or current thread (the default) circumvents the problem, but obviously that's not always possible.

I believe I've tracked down the problem but wanted to run by the implications before formalising as a PR (mainly because I flat out can't run tests - which XUnit runner should I be using for RxUI? VS2013 with xUnit.net runner for Visual Studio 0.99.9 just isn't working for me, presumably because it's deprecated).

So anyway, `ObservableAsPropertyHelper` currently has this:
```C#
            bool firedInitial = false;
            subj.Subscribe(x => {
                // Suppress a non-change between initialValue and the first value
                // from a Subscribe
                if (firedInitial && EqualityComparer<T>.Default.Equals(x, _lastValue)) return;

                onChanging(x);
                _lastValue = x;
                onChanged(x);
                firedInitial = true;
            }, exSubject.OnNext);

            subj.OnNext(initialValue);
```
The problem is that the `ObservableForProperty` call has already gone off and evaluated the initial value by the time the initial value is fed into the `Subscribe` call. Thus, the observable created by `ObservableForProperty` sees two successive `false` values, only one of which is skipped. Ergo, WTF.

My thinking on this is: why does the initial value even *need* to go through the subject? I mean, it's an initial value - that implies it already applies and there's no "change" to that initial value. It should just be there already:
```C#
public ObservableAsPropertyHelper(
    IObservable<T> observable, 
    Action<T> onChanged, 
    Action<T> onChanging = null,
    T initialValue = default(T), 
    IScheduler scheduler = null)
{
    Contract.Requires(observable != null);
    Contract.Requires(onChanged != null);

    scheduler = scheduler ?? CurrentThreadScheduler.Instance;
    onChanging = onChanging ?? (_ => {});

    var subj = new ScheduledSubject<T>(scheduler);
    var exSubject = new ScheduledSubject<Exception>(CurrentThreadScheduler.Instance, RxApp.DefaultExceptionHandler);

    subj.Subscribe(x => {
        onChanging(x);
        _lastValue = x;
        onChanged(x);
    }, exSubject.OnNext);

    ThrownExceptions = exSubject;

    _lastValue = initialValue;
    _source = observable.DistinctUntilChanged().Multicast(subj);
    if (ModeDetector.InUnitTestRunner()) {
        _inner = _source.Connect();
    }
}
```
This certainly seems to work in my case, but I won't feel confident until I can run the tests. Also wanted to get input from those more familiar with the code base than I.

kentcb:
Fixes #815.

The initial value is no longer piped through the scheduled subject. Instead, it is prepended to the source observable. Consequently, it is immediately available.

### DIFF
--- a/ReactiveUI/ObservableAsPropertyHelper.cs
+++ b/ReactiveUI/ObservableAsPropertyHelper.cs
@@ -66,32 +66,23 @@ namespace ReactiveUI
             Contract.Requires(onChanged != null);
 
             scheduler = scheduler ?? CurrentThreadScheduler.Instance;
-            onChanging = onChanging ?? (_ => {});
-            _lastValue = initialValue;
+            onChanging = onChanging ?? (_ => { });
 
             var subj = new ScheduledSubject<T>(scheduler);
             var exSubject = new ScheduledSubject<Exception>(CurrentThreadScheduler.Instance, RxApp.DefaultExceptionHandler);
 
-            bool firedInitial = false;
             subj.Subscribe(x => {
-                // Suppress a non-change between initialValue and the first value
-                // from a Subscribe
-                if (firedInitial && EqualityComparer<T>.Default.Equals(x, _lastValue)) return;
-
                 onChanging(x);
                 _lastValue = x;
                 onChanged(x);
-                firedInitial = true;
             }, exSubject.OnNext);
 
             ThrownExceptions = exSubject;
 
-            // Fire off an initial RaisePropertyChanged to make sure bindings
-            // have a value
-            subj.OnNext(initialValue);
-            _source = observable.DistinctUntilChanged().Multicast(subj);
-
-            if (ModeDetector.InUnitTestRunner()) {
+            _lastValue = initialValue;
+            _source = observable.StartWith(initialValue).DistinctUntilChanged().Multicast(subj);
+            if (ModeDetector.InUnitTestRunner())
+            {
                 _inner = _source.Connect();
             }
         }
--- a/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
+++ b/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
@@ -34,6 +34,37 @@ namespace ReactiveUI.Tests
         }
 
         [Fact]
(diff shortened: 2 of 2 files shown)


ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -66,32 +66,23 @@ public sealed class ObservableAsPropertyHelper<T> : IHandleObservableErrors, IDi
...
-            // Fire off an initial RaisePropertyChanged to make sure bindings
-            // have a value
-            subj.OnNext(initialValue);
-            _source = observable.DistinctUntilChanged().Multicast(subj);
-
-            if (ModeDetector.InUnitTestRunner()) {
+            _lastValue = initialValue;
+            _source = observable.StartWith(initialValue).DistinctUntilChanged().Multicast(subj);
+            if (ModeDetector.InUnitTestRunner())
+            {
```
WilkaH: Mr Betts isn't going to like the new line here :)

PR #939, opened 2015-09-11
Add a reactive version of the Android.Support.V4.DialogFragment class

flagbug:


ReactiveUI.AndroidSupport/ReactiveDialogFragment.cs
```diff
@@ -0,0 +1,110 @@
...
+namespace ReactiveUI.AndroidSupport
+{
+    /// <summary>
+    /// This is a DialogFragment that is both a DialogFragment and has ReactiveObject powers 
+    /// (i.e. you can call RaiseAndSetIfChanged)
+    /// </summary>
+    public class ReactiveDialogFragment<TViewModel> : ReactiveDialogFragment, IViewFor<TViewModel>, ICanActivate
+        where TViewModel : class
+    {
+        protected ReactiveDialogFragment() { }
```
kentcb: Is there a reason the ctor is protected but the class is not abstract? Same question for the non-generic counterpart.

flagbug: Yeah, you're right, I'll fix that

flagbug: @kentcb I just looked at the other fragments (i.e `ReactiveUI.AndroidSupport.DialogFragment`) and they're also not marked as `abstract`

kentcb: Fair enough - broken windows theory and all that ;)

PR #349, opened 2013-07-18
Add ReactiveUserControl<ViewModel> helper class

bradphelan:
Fixes #348

PR #943, opened 2015-09-17
Add ReactiveUserControl for XAML platforms.

kentcb:
This PR adds a `ReactiveUserControl<TViewModel>` for XAML platforms. It is the simplest possible thing that works and is intended to supersede #349 in the absence of a reason for the complexity of that solution.

### DIFF
--- a/ReactiveUI/ReactiveUI_Net45.csproj
+++ b/ReactiveUI/ReactiveUI_Net45.csproj
@@ -82,6 +82,7 @@
     <Compile Include="ExpressionMixins.cs" />
     <Compile Include="ExpressionRewriter.cs" />
     <Compile Include="Platform\ComponentModelTypeConverter.cs" />
+    <Compile Include="Xaml\ReactiveUserControl.cs" />
     <Compile Include="Xaml\WpfAutoSuspendHelper.cs" />
     <Compile Include="Xaml\WpfDependencyObjectObservableForProperty.cs" />
     <Compile Include="Xaml\ActivationForViewFetcher.cs" />
--- a/ReactiveUI/ReactiveUI_WP8.csproj
+++ b/ReactiveUI/ReactiveUI_WP8.csproj
@@ -89,6 +89,7 @@
     <Compile Include="Xaml\DependencyObjectObservableForProperty.cs" />
     <Compile Include="Xaml\PhoneServiceStateDriver.cs" />
     <Compile Include="Xaml\PlatformOperations.cs" />
+    <Compile Include="Xaml\ReactiveUserControl.cs" />
     <Compile Include="Xaml\RoutedViewHost.cs" />
     <Compile Include="Xaml\TransitioningContentControl.cs" />
     <Compile Include="Xaml\ViewModelViewHost.cs" />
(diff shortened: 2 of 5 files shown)


bradphelan:
I gave a reason for the complexity of my solution. It has to do with isolating the DataContext from inside the usercontrol from the datacontext outside the usercontrol.

kentcb:
I get the isolation thing. What I don't get is why you're overriding the UC's DataContext from the outside. If you don't do that, there's no need for the added complexity.

bradphelan:
It's the other way round. If you set data context from the inside then when
you try to apply bindings from the outside your data context is the one
from the inside instead of the lexical xaml scope you would expect.

kentcb:
Maybe I'm being thick, but I'm not getting it. Why would you set the DC twice, whether it be from the inside or the outside? Can you give me a concrete example of where this would be necessary?

The fact is, I'm successfully using my solution in a non-trivial app and I've not run into any need for this.

bradphelan:
If you set the DataContext on a user control from inside it is exactly the
same as if you set the DataContext from outside. If you do that then
bindings don't work as the user might expect.

However this is only important if you would like to use DataContext inside
your user control. I suspect that you just use the ViewModel property.

Hard to write example at the moment as I'm on my mobile sorry.

kentcb:
I think I get what you're saying, but this counts as a big reason *not* to include such an implementation in RxUI because it is non-idiomatic. Using RxUI bindings has a lot of advantages over WPF bindings.

That said, I see no reason why you couldn't extend the `ReactiveUserControl<T>` provided by RxUI and have it bind the `ViewModel` to your `DataContext`.

bradphelan:
I don't use RXUI bindings. With Resharper it does static analysis on the XAML and allows me to do rename refactor through XAML to view model. In this case RXUI binding have no advantages.

kentcb:
Your solution is only useful if used with R#? Sounds like you're making my argument for me :)

But seriously, there's more advantages to RxUI bindings, including:

* they're all in one place! And this is consistent across all your views. No more digging through reams of XAML to tweak bindings
* simple and more powerful conversion of data than using value converters. Just provide a Func<TOld, TNew> and you can do conversions inline. If you need more power, you can write an `IBindingTypeConverter` and either manually pass it in, or register it so it is used automatically for relevant bindings
* memory management. It's super easy to drop all your bindings when your views are deactivated
* full control over signalling when the view property has changed, and thus that the VM property should be updated. No more picking between the update source triggers that WPF decides are appropriate - now you have full control

Anyway, if you're happy with what you're doing then by all means don't let me try and talk you out of it. I guess the point is that we should be pushing RxUI consumers along the idiomatic, pit of success, path.

bradphelan:
Rx binding don't work with data templates. So that makes them pretty much
useless.

bradphelan:
From the man himself.

http://stackoverflow.com/questions/25466260/bind-data-inside-listbox-using-code-behind-in-reactiveui

I asked this question years ago. I was impressed with the ability to do
type safe data bindings but without the ability to do data templates you
end up with a mix of type safe and un typesafe bindings. There is no
idomatic RXUI way to do bindings if you can't be consistent. The only way
to get it to work is to push all datatemplates into user-controls which is
generally not too practical.

kentcb:
Moving the XAML from inline into UserControls is exactly what I do and don't find it impractical. In fact, I find it results in better-factored, easier to understand XAML. The only downside is needing to explicitly bind the VM:

```XML
<DataTemplate>
    <views:SomeView ViewModel="{Binding}"/>
</DataTemplate>
```

A little sacrifice for a lot of gain.

PR #952, opened 2015-09-18
Add support for View activation.

### LINKED ISSUE #949: Add activation support for Xamarin Forms Views
The Xamarin Forms `ActivationForViewFetcher` currently only supports `Page` and `Cell` objects. Conspicuous in its absence is support for `View` objects. That's most likely because XF hasn't deemed it fit to expose lifecycle related events for views, so it's a bit harder than we would like.

This code seems to do the trick, but I just need to test more before committing to it.

```C#
using System;
using System.ComponentModel;
using System.Linq;
using System.Reactive;
using System.Reactive.Linq;
using System.Reflection;
using Xamarin.Forms;

namespace ReactiveUI.XamForms
{
    public class ActivationForViewFetcher : IActivationForViewFetcher
    {
        public int GetAffinityForView(Type view)
        {
            return
                (typeof(Page).GetTypeInfo().IsAssignableFrom(view.GetTypeInfo())) ||
                (typeof(View).GetTypeInfo().IsAssignableFrom(view.GetTypeInfo())) ||
                (typeof(Cell).GetTypeInfo().IsAssignableFrom(view.GetTypeInfo()))
                ? 10 : 0;
        }

        public IObservable<bool> GetActivationForView(IActivatable view)
        {
            var ret = Observable.Never<bool>();
            

            var page = view as Page;

            if (page != null) {
                ret = Observable.Merge(
                    Observable.FromEventPattern<EventHandler, EventArgs>(x => page.Appearing += x, x => page.Appearing -= x).Select(_ => true),
                    Observable.FromEventPattern<EventHandler, EventArgs>(x => page.Disappearing += x, x => page.Disappearing -= x).Select(_ => false));
            } else {
                var xfView = view as View;

                if (xfView != null) {
                    var propertyChanged = Observable.FromEventPattern<PropertyChangedEventHandler, PropertyChangedEventArgs>(
                        x => xfView.PropertyChanged += x,
                        x => xfView.PropertyChanged -= x);
                    var parentChanged = propertyChanged
                        .Where(x => x.EventArgs.PropertyName == "Parent")
                        .Select(_ => Unit.Default);

                    return parentChanged
                        .StartWith(Unit.Default)
                        .Select(_ => GetPageFor(xfView))
                        .Select(x =>
                            x == null ?
                                Observable.Return(false) :
                                Observable
                                    .Merge(
                                        Observable.FromEventPattern<EventHandler, EventArgs>(y => x.Appearing += y, y => x.Appearing -= y).Select(_ => true),
                                        Observable.FromEventPattern<EventHandler, EventArgs>(y => x.Disappearing += y, y => x.Disappearing -= y).Select(_ => false))
                                    .StartWith(true))
                        .Switch();
                } else {
                    var cell = view as Cell;

                    if (cell != null)
                    {
                        ret = Observable
                            .Merge(
                                Observable.FromEventPattern<EventHandler, EventArgs>(x => cell.Appearing += x, x => cell.Appearing -= x).Select(_ => true),
                                Observable.FromEventPattern<EventHandler, EventArgs>(x => cell.Disappearing += x, x => cell.Disappearing -= x).Select(_ => false));
                    }
                }
            }

            return ret.DistinctUntilChanged();
        }

        private static Page GetPageFor(Element element)
        {
            Page page = null;

            while (element != null)
            {
                page = element as Page;

                if (page != null)
                {
                    return page;
                }

                element = element.Parent;
            }

            return null;
        }
    }
}
```

kentcb:
Fixes #949 by adding support for `View` activation.

The implementation is a little more awkward than we would like, but that's due to Xamarin Forms not exposing any lifecycle events for `View` objects. Instead, we have to hook into the lifecycle events of the view's hosting `Page`.

### DIFF
--- a/ReactiveUI.XamForms/XamForms/ActivationForViewFetcher.cs
+++ b/ReactiveUI.XamForms/XamForms/ActivationForViewFetcher.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.ComponentModel;
-using System.Linq;
 using System.Reactive;
 using System.Reactive.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


ReactiveUI.XamForms/XamForms/ActivationForViewFetcher.cs
```diff
@@ -17,6 +14,7 @@ public int GetAffinityForView(Type view)
         {
             return
                 (typeof(Page).GetTypeInfo().IsAssignableFrom(view.GetTypeInfo())) ||
+                (typeof(View).GetTypeInfo().IsAssignableFrom(view.GetTypeInfo())) ||
```
gluck: Maybe splitting this type in 3 activators (one for each type) will avoid the growing complexity of the below method and clearly separate each behavior.

ReactiveUI.XamForms/XamForms/ActivationForViewFetcher.cs
```diff
@@ -32,17 +31,61 @@ public IObservable<bool> GetActivationForView(IActivatable view)
...
+
+                    return parentChanged
+                        .StartWith(Unit.Default)
+                        .Select(_ => GetPageFor(xfView))
+                        .Select(x =>
+                            x == null ?
+                                Observable.Return(false) :
+                                Observable
+                                    .Merge(
+                                        Observable.FromEventPattern<EventHandler, EventArgs>(y => x.Appearing += y, y => x.Appearing -= y).Select(_ => true),
```
gluck: Not really sure we care about performance here, but if we do, there's a `FromEvent` overload which performs much better than `FromEventPattern` as it uses no reflection ([benchmark](https://gist.github.com/gluck/da82c1b537e44b55daa8)), and isn't really more complex to write:

instead of:

    Observable.FromEventPattern<EventHandler, EventArgs>(y => x.Appearing += y, y => x.Appearing -= y).Select(_ => true)

You can write:

    Observable.FromEvent<EventHandler, bool>(a => (_, __) => a(true), y => x.Appearing += y, y => x.Appearing -= y)

kentcb: Cool - will do. Thanks for the tip.

ReactiveUI.XamForms/XamForms/ActivationForViewFetcher.cs
kentcb: Hmmm, I think I'll split into distinct methods for now. I don't really want to have one platform differ by having multiple activator implementations.

ReactiveUI.XamForms/XamForms/ActivationForViewFetcher.cs
kentcb: Actually, reading into this a bit more, I don't think I'm going to make this change because it goes against all the advice I can find. Basically, if the event conforms to the normal pattern then the advice is to use `FromEventPattern`. Granted, it may be a tad slower, but I think it better to follow the official guidance.

PR #956, opened 2015-09-22
Remove the legacy XAML-based platforms

flagbug:
- Windows Phone 8
- Windows Phone Silverlight
- WinRT 8.0

I didn't remove any legacy iOS platforms, because I've never developed for iOS, so I'm not sure what actually needs to be removed


jlaanstra:
I would like to keep Wp8.1 Silverlight and WinRT around. Since we don't have anything WP8.1 specific that probably means we can keep WP8 Silverlight?

ghuntley:
I'd prefer (personal bias) to keep WP8 (inc Silverlight) at this stage even if it means I'm the guy who adopts/feeds it. If it becomes a burden (ie. doesn't open in VS2016/VS2017) then :fire:

ghuntley:
I was doing Mango (7.2-7.8) development this morning which is definitely dead but that provides some perspective into the enterprise --> consumer lifecycle. Whilst WP8 (inc silverlight) open in Visual Studio and cause no strain on the team my preference is to keep it in, even if I have to coddle it.

flagbug:
> I was doing Mango (7.2-7.8) development this morning

Oh wow :smile: :man: :gun: 

>  Whilst WP8 (inc silverlight) open in Visual Studio and cause no strain on the team my preference is to keep it in, even if I have to coddle it.

Sounds reasonable. Maybe we can :fire: it in with the next version of VS

ghuntley:
Agreed

moswald:
Just FYI, the UAP branch has removed these projects (except WP8, since it still works). The current PR was created for `master`, but it should be abandoned and retargeted to `rxui7-master`.

ghuntley:
:gun: iOS legacy imho. I've started deleting it as part of #1043

PR #961, opened 2015-10-01
Feature/184

bradphelan:
Fixes a race condition which causes double subscribe to observable. See testcase


bradphelan:
@kentcb I blew away the old pull request. This one is based on rxui7-master and should be ready to go.

bradphelan:
fixes https://github.com/reactiveui/ReactiveUI/issues/814

ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -24,6 +25,7 @@ public sealed class ObservableAsPropertyHelper<T> : IHandleObservableErrors, IDi
         readonly IConnectableObservable<T> _source;
         IDisposable _inner;
                 
+        private int _activated;
```
flagbug: Why is this an `int` and not a `bool`?

kentcb: @flagbug that'd be because `Interlocked` is being used for thread-safe compare/exchange, and `Interlocked` does not support `bool`.

flagbug: @kentcb Oh, I only saw an overload with `T`on MSDN, but didn't see it was restricted to reference types, my fault!

kentcb:
Thanks @bradphelan

StefanDoubleU:
Any chance or releasing a version 6.5.0.1 with this hotfix? Would help a lot!

Thanks in advance
Stefan

Issue #920, opened 2015-08-14
Generalize UserError


kentcb:
The `UserError` class is more widely applicable than just handling errors. It could also be used to prompt the user for confirmations or warnings, or simply to give them an important piece of information. Whilst it's technically possible to use it for such scenarios, the name is off-putting.

Renaming `UserError` to `UserAlert` would result in an API more suggestive of its purpose. Other members (such as the `Throw` method) will need to be renamed too.

kentcb:
@paulcbetts before I put any effort into a PR, is this a direction you'd be happy with?

flagbug:
I think this is a good idea :+1: 

Currently I'm passing functions to my ViewModels for user confirmations like this: `public MyViewModel(Func<Task<bool>> confirmDeletion)` but that doesn't really feel right.

ghuntley:
I agree, `UserError` poorly explains what is possible and just how powerful it is.

PR #964, opened 2015-10-05
User interaction

kentcb:
# OPEN FOR DISCUSSION

This feature is intended to supplant the existing `UserError` feature (see #920 for original issue). The goals are as follows:

* Generalize the problem. Often we need to interact with the user in non-error scenarios and it felt awkward doing so via `UserError`
* Provide greater flexibility. The `UserError` infrastructure was very opinionated in what constituted an interaction. Stuff like `RecoveryOptionResult` couldn't be extended.
* Provide a non-global means of implementing interactions. Since we're no longer dealing with just errors, it makes a lot more sense that we might want to handle interactions locally rather than by some global handler.

## How it works

Developers use the `UserInteraction<TResult>` class to represent an interaction with a user. `TResult` here represents the type of answer expected back from the user. For example, if we're asking them a yes/no question we could use `UserInteraction<bool>`. If we're asking them for a filename, a `UserInteraction<FileInfo>` might be appropriate.

To obtain an answer from the user, developers have two options: handle the interaction locally or globally.

For local interactions, the typical approach would be:

```C#
// view model code
public UserInteraction<bool> Confirmation
{
    get { return this.confirmation; }
    private set { this.RaiseAndSetIfChanged(ref this.confirmation, value); }
}

private async Task DoSomethingDangerous()
{
    this.Confirmation = new UserInteraction<bool>();
    var proceed = await this.Confirmation.Raise();

    if (!proceed)
    {
        // user opted out
        return;
    }

    // do the dangerous thing
}


// view code
this
    .WhenAnyValue(x => x.ViewModel.Confirmation)
    .Where(x => x != null)
    .Subscribe(
        async x =>
        {
            var result = await ShowDialog("Confirm", "Are you sure", blabla);
            x.SetResult(result == DialogResult.OK);
        })
    .AddTo(disposables);
```

All we do is create the interaction instance and make sure the view can see it changing. The VM awaits the result of calling `Raise`, which essentially waits for an answer to the interaction. The view then looks out for those interaction instances and prompts the user accordingly. Importantly, it always provides a result via the `SetResult` method. If nothing provides a result, our VM will just wait indefinitely.

Typical global interaction usage looks like this:

```C#
// view model code
private async Task DoSomethingDangerous()
{
    var confirmation = new UserInteraction<bool>();
    var proceed = await confirmation.RaiseGlobal();

    if (!proceed)
    {
        // user opted out
        return;
    }

    // do the dangerous thing
}


// code elsewhere, probably in a view but possibly at a higher level of the application
UserInteraction
    .RegisterGlobalHandler<UserInteraction<bool>>(
        async interaction =>
        {
            var result = await ShowDialog("Confirm", "Are you sure", blabla);
            x.SetResult(result == DialogResult.OK);
        });
```
So for global interactions, the VM does not need to expose the interaction via its own API. Instead, it just awaits the result of a call to `RaiseGlobal`. A non-generic base class (`UserInteraction`) provides the necessary static APIs.

Raising errors globally works much the same as it did for `UserError`. Multiple handlers can be registered, and later registered handlers take precedence over earlier ones.

## What's "missing"?

* the ability to override handlers, but I'm not sure this is necessary. As per my tests one can just use the `RegisterGlobalHandler` method and dispose the result.
* recovery options. Again, not sure this is necessary. I've included a subclass of `UserInteraction<TResult>` called `UserErrorInteraction<TResult>`. Essentially, it adds an exception. I figure if people want recovery options and whatever else they should just define that stuff in their own application and subclass the interactions that RxUI provides.


flagbug:
> Provide a non-global means of implementing interactions. Since we're no longer dealing with just errors, it makes a lot more sense that we might want to handle interactions locally rather than by some global handler.

I wonder if we should/could do that for `UserError`s too, I really don't like how they're global.

Also I'm really :+1: :sparkling_heart: :sparkles: for this.

kentcb:
@flagbug: well, the idea will be to completely remove `UserError` because `UserInteraction` encompasses those scenarios too. I haven't done that yet, but it'll be a separate commit in this PR.

flagbug:
@kentcb Even better!

mteper:
A couple of comments:
 - naming-wise, I am partial to `Raise` over `Propagate`, anyone else have a preference?
 - just to confirm, one can have multiple global handlers defined, right? (one per the specific UserInteraction<T> type)

kentcb:
@mteper : multiple handlers are definitely supported.

As for naming, I'm not happy yet either. I was hoping for some symmetry between handling interactions locally versus handling them via the global handlers. For example:

* `Handle` / `HandleLocal`
* `Raise` / `RaiseLocal`
* `GetResult(bool propagate = true)`

I'm not sold on any of these APIs yet - what do you think? And does anyone else have any thoughts on naming?

mteper:
I would go with `Raise` and `RaiseGlobal` because "global" is easier to grok intuitively. People might get confused about what scope "local" refers to. Also, rather than passing a parameter to `GetResult`, I suggest `GetResult` and `PeekResult`.

mteper:
Alternatively to `RaiseGlobal`, hang a `Raise` method off the non-generic static `UserInteraction` class, thus giving symmetry to handling global interactions.

kentcb:
@mteper: please see updated naming in latest commit.

mteper:
@kentcb :cool: but what did you think about the idea of having `UserInteraction.Raise<T>(...)`? Perhaps then you could drop "Global" from `RegisterGlobalHandler`?

kentcb:
@mteper: I actually prefer it to be explicit. Otherwise, these very similar looking pieces of code do quite different things:

```C#
var interaction = ...;

// these do different things!
await interaction.Raise();
await UserInteraction.Raise(interaction);
```

mteper:
@kentcb yup, good point. Maybe putting "Global" into the static type would make it better?

```c-sharp
var interaction = ...;

// these do different things!
await interaction.Raise();
await GlobalUserInteraction.Raise(interaction);
```

mteper:
I think we are just dancing around the fact that a `UserInteraction` both carries data and has behavior. Would be simpler if it were just data.

kentcb:
I dunno man. The "behavior" is just to wait for data. It seems appropriate to me, but I'm not clear on what alternative you're suggesting. So it would look like this as it stands:

```C#
var interaction = ...;

// do one of these depending on requirements
await interaction.Raise();
await interaction.RaiseGlobal();
```

Are you suggesting something like this?

```C#
var interaction = ...;

// do one of these
await UserInteraction.Raise(interaction);
await UserInteraction.RaiseGlobal(interaction);
```

It seems to me that just moves the problem, and makes the client code more verbose and less discoverable.

mteper:
@kentcb yup, no argument, was just throwing a thought out there in case it sparked something. I'm good with what you've got.

kentcb:
Excellent! Thanks for the input @mteper. I like the API a lot more now.

moswald:
So here's my thoughts on the "local" version of `UserInteraction`: it turns a functional, reactive object into a procedural one. Rather than throwing an observable object over the wall for the View to modify and toss back, I prefer the following pattern.

```
public class ConfirmEventViewModel
{
    public ReactiveCommand<Unit> ConfirmCommand { get; protected set; }
    public ReactiveCommand<Unit> CancelCommand { get; protected set; }
}

{
	public ReactiveCommand<Unit> DoSomethingDangerousCommand { get; }
	
	Subject<ConfirmEventViewModel> _confirmDangerousEvents = new Subject<ConfirmEventViewModel>();
	public IObservable<ConfirmEventViewModel> ConfirmDangerousEvents => _confirmDangerousEvents.AsObservable();
	
    public MyViewModel()
    {
		DoSomethingDangerousCommand = ReactiveCommand.CreateAsyncTask(
			_ =>
			{
				_confirmDangerousEvents.OnNext(new ConfirmEventViewModel
				{
					ConfirmCommand = ReactiveCommand.CreateAsyncTask(/* do the dangerous thing */);
					CancelCommand = ReactiveCommand.CreateAsyncTask(/* don't do it/back out/whatever */);
				});
				
				return Task.CompletedTask;
			});
    }
}

// somewhere in the view code
this
	.WhenAnyObservable(x => x.ViewModel.ConfirmDangerousEvents)
	.Subscribe(confirm =>
	{
		var yesNo = /* prompt user */;
		
		if (yesNo)
		{
			confirm.ConfirmCommand.ExecuteAsyncTask();
		}
		else
		{
			confirm.CancelCommand.ExecuteAsyncTask();
		}
	});
```

mteper:
@moswald what do you see as pros of this approach? My gut reaction is that it's a lot of ceremony and the view is forced to do some extra work (executing commands), but I am not criticizing, just genuinely want to understand the motivating factors.

moswald:
This was a very short example, but it's functionally the same as what I'm using in projects today. I went this route because in most cases, my app doesn't present a dialog. Instead, more controls are usually added to the visual tree.

For example, instead of prompting "do you really want to delete this file", we instead add a text box that the user enters the file name into and then presses the confirm button. The confirm command is bound to the button press, like any other button in the app.

moswald:
Also, `ReactiveCommand`s are more powerful than simple booleans.

mteper:
@moswald Playing devil's advocate, aren't you pushing logic into the view that doesn't belong there? The proposed `UserInteraction<T>` enables `T` to be anything, so it's not just a boolean value that comes back to the view model. It can contain form text / data, it can contain action selection (i.e. the button that was pressed that gets passed back via, for example, an enum), etc. Then the view model gets to act on it, directly or by executing commands. Feels like better encapsulation, better testability, no?

moswald:
I don't think so. The view is free to present the confirmation however it wants, and the vm has all the logic. The `ConfirmEventViewModel` in my example can hold presentation details if required.

mteper:
@moswald true, but in that example the VM is not in control of logic flow. The VM constructs commands and exposes them, but the view gets the decide which command to execute depending on user response. It's subtle, but I think important distinction. I would argue that `if (yesno) {} else` block belongs in the VM.

moswald:
I don't agree, since the VM is doing nothing more than providing the confirmation and cancellation commands, as it would normally do. The view is then free to present the commands however it wants. Normally, I'd bind it to a command-enabled control, like `Button`. In fact, I've got a `PopupObserver` class that uses `ViewModelViewHost` to visualize certain prompts that I *do* want to be a popup. The `IViewFor<SomeDerivedEventViewModel>` uses `BindCommand` on the confirm and cancel buttons.

moswald:
To summarize brief Slack discussion - we're splitting hairs and getting off-track. :)

I am not a fan of the "local" version. Views and ViewModels are already in tight communication, and having a procedural method of prompting the user built right into the framework will probably do newcomers (and those of us with more experience!) a disservice.

PR #968, opened 2015-10-12
Support ancestral parent changes in View activation.

kentcb:
The code I wrote previously to detect parent changes for a `View` so that activation can be hooked is insufficient. We need to monitor all ancestors.


kentcb:
Ended up just using `IsVisible` to detect `View` activation. This saves from having to construct absurdly elaborate and fragile recursive observables monitoring each view in the visual tree.

PR #969, opened 2015-10-14
[RFC] Add the "subscribeImmediately" parameter to the ObservableAsPropertyHelper class

flagbug:
When this parameter is set to true, the ObservableAsPropertyHelper
immediately subscribes to the source sequence.

### DIFF
--- a/ReactiveUI/ObservableAsPropertyHelper.cs
+++ b/ReactiveUI/ObservableAsPropertyHelper.cs
@@ -34,14 +34,22 @@ namespace ReactiveUI
         /// changes, typically this will call the ViewModel's
         /// RaisePropertyChanged method.</param>
         /// <param name="initialValue">The initial value of the property.</param>
+        /// <param name="subscribeImmediately">
+        /// A value indicating, whether the <see cref="ObservableAsPropertyHelper{T}"/> 
+        /// should immediately subscribe to the source <paramref name="observable"/>, or not.
+        /// 
+        /// If this parameter is <c>false</c>, the <see cref="ObservableAsPropertyHelper{T}"/> 
+        /// subscribes to the <paramref name="observable"/> when the first <see cref="Value"/> is requested.
+        /// </param>
         /// <param name="scheduler">The scheduler that the notifications will be
         /// provided on - this should normally be a Dispatcher-based scheduler
         /// (and is by default)</param>
         public ObservableAsPropertyHelper(
             IObservable<T> observable, 
             Action<T> onChanged, 
-            T initialValue = default(T), 
-            IScheduler scheduler = null) : this(observable, onChanged, null, initialValue, scheduler) {}
+            T initialValue = default(T),
+            bool subscribeImmediately = false,
+            IScheduler scheduler = null) : this(observable, onChanged, null, initialValue, subscribeImmediately, scheduler) {}
 
         /// <summary>
         /// Constructs an ObservableAsPropertyHelper object.
@@ -54,6 +62,13 @@ namespace ReactiveUI
         /// changes, typically this will call the ViewModel's
         /// RaisePropertyChanging method.</param>
         /// <param name="initialValue">The initial value of the property.</param>
+        /// <param name="subscribeImmediately">
+        /// A value indicating, whether the <see cref="ObservableAsPropertyHelper{T}"/> 
+        /// should immediately subscribe to the source <paramref name="observable"/>, or not.
+        /// 
+        /// If this parameter is <c>false</c>, the <see cref="ObservableAsPropertyHelper{T}"/> 
+        /// subscribes to the <paramref name="observable"/> when the first <see cref="Value"/> is requested.
+        /// </param>
         /// <param name="scheduler">The scheduler that the notifications will be
         /// provided on - this should normally be a Dispatcher-based scheduler
         /// (and is by default)</param>
@@ -61,7 +76,8 @@ namespace ReactiveUI
             IObservable<T> observable, 
             Action<T> onChanged, 
             Action<T> onChanging = null,
-            T initialValue = default(T), 
+            T initialValue = default(T),
+            bool subscribeImmediately = false,
             IScheduler scheduler = null)
         {
             Contract.Requires(observable != null);
@@ -83,9 +99,10 @@ namespace ReactiveUI
 
             _lastValue = initialValue;
             _source = observable.StartWith(initialValue).DistinctUntilChanged().Multicast(subj);
-            if (ModeDetector.InUnitTestRunner())
+            if (subscribeImmediately || ModeDetector.InUnitTestRunner())
             {
                 _inner = _source.Connect();
+                _activated = 1;
             }
         }
 
@@ -125,7 +142,7 @@ namespace ReactiveUI
         /// (and is by default)</param>
         public static ObservableAsPropertyHelper<T> Default(T initialValue = default(T), IScheduler scheduler = null)
         {
-            return new ObservableAsPropertyHelper<T>(Observable.Never<T>(), _ => {}, initialValue, scheduler);
+            return new ObservableAsPropertyHelper<T>(Observable.Never<T>(), _ => {}, initialValue, false, scheduler);
         }
     }
 
@@ -136,6 +153,7 @@ namespace ReactiveUI
                 IObservable<TRet> observable,
                 Expression<Func<TObj, TRet>> property,
                 TRet initialValue = default(TRet),
+                bool subscribeImmediately = false,
                 IScheduler scheduler = null)
             where TObj : IReactiveObject
         {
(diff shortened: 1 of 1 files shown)


ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -83,9 +99,10 @@ public sealed class ObservableAsPropertyHelper<T> : IHandleObservableErrors, IDi
 
             _lastValue = initialValue;
             _source = observable.StartWith(initialValue).DistinctUntilChanged().Multicast(subj);
-            if (ModeDetector.InUnitTestRunner())
+            if (subscribeImmediately || ModeDetector.InUnitTestRunner())
             {
                 _inner = _source.Connect();
+                _activated = 1;
```
flagbug: @kentcb We can add this line, even if we don't merge this PR, so connect isn't called a second time in `Value`

ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -34,14 +34,22 @@ public sealed class ObservableAsPropertyHelper<T> : IHandleObservableErrors, IDi
...
         /// <param name="scheduler">The scheduler that the notifications will be
         /// provided on - this should normally be a Dispatcher-based scheduler
         /// (and is by default)</param>
         public ObservableAsPropertyHelper(
             IObservable<T> observable, 
             Action<T> onChanged, 
-            T initialValue = default(T), 
-            IScheduler scheduler = null) : this(observable, onChanged, null, initialValue, scheduler) {}
+            T initialValue = default(T),
+            bool subscribeImmediately = false,
```
WilkaH: Inserting `subscribeImmediately` here, rather than adding at the end does mean it'll break any code expecting this position to be a scheduler. I know breaking changes are acceptable in a major version bump, but is passing true here going to be more likely than wanting to specify a scheduler?

It might be worth adding this ctor as an overload, so we keep the scheduler at position 4 as well as adding this new flag.

The same comments apply to `ToProperty(..)`

flagbug: I didn't add it as last parameter, because in Rx, the `IScheduler` is always the last parameter, but you're probably right, that we could add it as overload

WilkaH: Good point. We should stick to `IScheduler` being the last parameter,

kentcb:
I know it's a big, scary breaking change and all, but my inclination would be to default `subscribeImmediately` to `true` (or - even better - rename the parameter to `delaySubscription` or `deferSubscription` and default it to `false`). That way, people aren't trolled by the default, and if/when they discover performance issues with OAPH they can dig deeper and move away from the default behavior. The pit of success with OAPH is way too narrow atm. This PR will _allow_ people to widen the pit, but I'd rather just have the pit wider by default.

moswald:
I would agree with that. `deferSubscription` would be my choice, since "delay" indicates you're waiting on some sort of timespan to pass.

flagbug:
The reason I think that delaying the subscription by default is good, is because I think that the constructor of an object shouldn't fire off any work on other threads.

Take the following class for example:

```cs
{
    private readonly ObservableAsPropertyHelper<int> myProperty;

    public MyViewModel()
    {
        this.myProperty = CalculateAsync().ToProperty(this, x => x.MyProperty);
    }

    public int MyProperty => this.myProperty.Value;

    public IObservable<int> CalculateAsync()
    {
        // Calculate something expensive here
    }
}
```

If `myProperty` would subscribe to the source immediately, the construction of `MyViewModel` now fire of some potentially expensive work on another thread, while the current behavior would subscribe when accessing `MyProperty` the first time

kentcb:
Isn't this a case of catering the esoteric scenario at the expense of the common?

I don't think it's a stretch to say that _most_ OAPH usage will simply be massaging existing properties into new ones. And to have those scenarios constantly require that people remember to specify `deferSubscription = false` seems backwards to me. After all, the people who are kicking off expensive work in a ctor are likely to notice and immediately remember that they should defer that particular subscription (or, if it's the first time they've come up against the problem, look into OAPH a bit and find the solution).

flagbug:
Yeah, you're probably right that my use case isn't very common.

Maybe @paulcbetts also has some input on this?

shiftkey:
cc @niik for feels on this change

kentcb:
Doesn't seem to be any further feedback on this and I would :heart: for this to get into the 7 branch ASAP so people can start banging on it.

@flagbug could you please change the argument name to `deferSubscription` and default it to `false`?

flagbug:
@kentcb Sure! I'll open a new PR for that change, because I deleted my fork where this PR is based :see_no_evil:

shiftkey:
Just did a quick scan of GitHub Desktop's code - we have 66 `.ToProperty()` methods, and only 10 of those set `scheduler: Scheduler.Immediate` to address this issue. I can look into why we've done that where we have, but I still favour lazy for our scenarios - even when in tests. 

Not sure how flipping the default behaviour here from lazy to eager will impact us, but I suspect it will. I don't want to hold this up this PR from being merged (I like this being exposed directly, and it's more clearer than using `Scheduler.Immediate`), but I'm curious about whether this is the right defaults (because I'm a fan of being lazy in general when programming).

kentcb:
Isn't scheduling a different issue entirely? The purpose of the `deferSubscription` parameter is to facilitate a lazy subscription, whereas scheduling just dictates where the target property will be scheduled to update. Scheduling is of no consequence if the property is lazy and nothing ever connects to it.

kentcb:
@shiftkey also, allow me to flip the question around a little: what negative consequences could there be from making the default behavior a hot observable? To me, this aligns the feature far more with people's expectations. We've seen that time and again through support questions and (certainly for me anyway) our own experiences using OAPH.

The only possible argument I see against making it hot is performance, and I would consider that a poor argument because:

1. If you really want lazy, you can still get it simply by passing `deferSubscription: true`
2. Why are people defining these properties if nothing is actually observing it? Surely that's the exception rather than the norm.
3. Performance at the expense of usability is a poor trade-off, imo, especially when that performance can be reinstated per (1)

shiftkey:
@kentcb I shall check the history on these changes to understand why we've done the `Scheduler.Immediate` thing. I believe this is the same issue, but perhaps viewed from two different perspectives...

shiftkey:
@kentcb 

> To me, this aligns the feature far more with people's expectations. We've seen that time and again through support questions and (certainly for me anyway) our own experiences using OAPH.

I'm probably missing some anecdotes on this, as aside from this `Scheduler.Immediate` thing our usage of OAPH has been largely conventional. Do you have a good example of newbies running into this so I can get some extra context?

> If you really want lazy, you can still get it simply by passing `deferSubscription: true`

This is true - and I'm not afraid of writing our own extension methods over this to ensure we do it by default.

> Why are people defining these properties if nothing is actually observing it? Surely that's the exception rather than the norm.

Perhaps this is the issue at hand. Something definitely necessitated us doing this, which made us feel bad at the time. I'll let you know what I find out.

> Performance at the expense of usability is a poor trade-off, imo, especially when that performance can be reinstated per (1)

For most of the usages as I see it we're not talking about _expensive_ operations. So it's gotta be something else.

flagbug:
One common example that just came to my mind is the Android `RecyclerView`. RecyclerView list items are virtualized, so only items that come into the view are actually used. This means that with deferred OAPH subscription, most of the ViewModel properties are never subscribed to, because the ViewModel never actually comes into view. 

With the new behavior of immediate subscription, all the properties are loaded at once, as soon as the ViewModels for the `RecyclerView` are created, which can potentially bog down performance with a lot of items (e.g if the ViewModels also load images)

kentcb:
@flagbug: interesting. My approach to such a scenario would be to only load the image if the VM is activated.

Here are some other examples:

* testing. Often I've run into issues testing because an assertion is being made against a property that is using OAPH. Of course, if nothing is subscribed to it then the test will fail, but it's not always simple to track down
* non-binding scenarios. For example, a VM exposes a property `A` that the view binds to. It derives a property `B` from `A` using OAPH (to do some further munging of the value). It wants to pass the value of `B` into the service layer whenever a command executes. But nothing is subscribed to `B`, so the value is unset.
* non-RxUI binding scenarios. For example, your VM uses OAPH and your iOS version uses RxUI binding. All is well. Then you do a WPF port (or whatever) and for perf reasons you have to use native bindings for a particular view. Things don't work because there are no subscriptions.

flagbug:
I've created the new PR at https://github.com/reactiveui/ReactiveUI/pull/1061

@kentcb Don't get me wrong, I don't like the current behavior, especially the switch where in unit tests, OAPH immediately subscribes to the source, which makes unit tests behave *different* than production code.

I just wanted to show a case where the current behavior is beneficial, if you have a VM with 10 properties times a list with 100 items, that's at least 1000 calls to `Subscribe`, which is suuuuuuuuper slow, especially on Android. So I guess what I want to say is that we have to make sure that people are aware of this and aren't blaming RxUI on the slowness, since they now can use the `deferSubscription` parameter

flagbug:
Closing in favor of https://github.com/reactiveui/ReactiveUI/pull/1061

PR #971, opened 2015-10-24
Reactive Command Rewrite

kentcb:
## What is this?

This PR re-imagines and re-writes `ReactiveCommand` and friends. There are breaking changes, but they were necessary to improve the overall experience of using the API.

### No interfaces

No longer are there any interfaces for reactive commands. The `IReactiveCommand` and `IReactiveCommand<T>` of today's code base is causing people to think they should be mocking commands. Moreover, it suggests there's some kind of valid reason why you might want to substitute implementations of your own. These interfaces are essentially causing people to fall into a pit, and not one of the success variety.

### Improved APIs

The parameter taken by a command is now strongly typed, and consumers will see only an `ExecuteAsync` method which optionally takes a parameter. There is no `Execute(object)` method exposed (except see below regarding `ICommand`).

### Explicit `ICommand` implementation

Reactive commands now implement `ICommand` explicitly. This means client code will need to type the reactive command as `ICommand` in order to see members like `Execute(object)` and `CanExecuteChanged`. This is a good thing, because these members are not asynchronous or reactive in nature.

### Behavioral properties

Relevant properties, such as `IsExecuting` and `CanExecute` are now behavioral in nature. That is, subscribers will always receive the latest value upon subscription.

### The type hierarchy

image

I get that some people may find it too complex, and in fact I'd rather it was simplified myself. However, I've done the simplest thing I can get away with that doesn't impose too much on general usage.

### ICommand.Execute implementation

The implementation throws if the provided parameter is of a type other than `TParam`. It could be more lenient and pass through `default(TParam)` whenever it gets an object of any other type. However, it seems to me that failing early (with a clear message, which is on my list of things to tidy up) is better.

### ToCommand is no more

The `ToCommand` extension method has been removed. It doesn't really offer a lot any more because creating a command requires the execution logic. Thus, we'd need a whole bunch of `ToCommand` overloads to support what is achievable through `ReactiveCommand.CreateXxx`.


flagbug:
> No ICommand implementation

This would mean that XAML command bindings don't work anymore, right?

kentcb:
@flagbug: that depends on what you mean by "don't work". If you're set on using XAML bindings rather than RxUI bindings, you can just use a call to `ToPlatform` to surface the reactive command as an `ICommand`. This is much the same approach used by Splat with things like `Color` and `ToNative`.

Of course, if you're using RxUI bindings then `ToPlatform` will be called on your behalf where it makes sense to do so. At least, that is my intention - that part of the code isn't present in this PR yet.

The idea is to alleviate the core implementation from the need to implement a platform-specific abstraction (`ICommand`) because it makes things awkward. It assumes synchronous commands and untyped parameters, neither of which gel well with the stronger contract provided by the reactive commands. And `ICommand` means nothing to iOS/Android/Winforms and probably more platforms in the future.

flagbug:
@kentcb But how do I cann `ToPlaftorm` when I bind my command like this?

`<Button Command={Binding MyCommand} />`

kentcb:
@flagbug: either in your VM (if you don't mind tying it to `ICommand`) or in a simple converter.

moswald:
I'm not sold on the separating of sync- vs. async-command. Is there really a need for that? A `ReactiveCommand` should be async. The `Execute` method was there because of `ICommand`, which you're doing away with anyway.

If you're going to break the world, then let's Do It Right: synchronous methods are Out, asynchronous methods are In.

moswald:
I've never liked the idea of `ReactiveCommand` tracking the number of "in-flight" operations. It looks like it's complicating the code for little real benefit. If the user wants to ignore `IsExecuting` and start a new operation, let them track their own limit.

kentcb:
@moswald: good points, both.

The in-flight stuff was in the old code base and for some reason I got lost in trying to fix it rather than taking a step back and questioning its utility. You're right - I think we can totally lose that. If people need something like it they can always roll their own.

As for sync vs async, I think you're probably right there, too. I guess my only concern is that people will see `ExecuteAsync` even when the operation executes synchronous. It might be a little bit off-putting for people unfamiliar with it. For example, the routing infrastructure will be affected:

```C#
// old (param is loosely typed as object)
this.hostScreen.Router.Navigate.Execute(someVM);

// new (param is strongly typed as IRoutableViewModel)
this.hostScreen.Router.Navigate.ExecuteAsync(someVM);
```

Probably not a big thing to get hung up on?

paulcbetts:
> No interfaces

A+ Best Idea I've Ever Heard

> More granular types and improved APIs

I don't think it's a good idea to split these back up again. Nobody understood the split when it was split up, and the fact is, almost every interesting Command is going to have an async component to it. Async should be the default, and sync is the boring one. 

> No ICommand implementation

This is gonna hurt, a _ton_ of people still use XAML binding and they will be none-too-pleased that their bindings are all broken now (being able to call ToPlatform means everyone has to do a bunch of shit work making dummy versions of their properties)

> Typed command parameters

Into this, but you can only do this because you're :fire:ing ICommand

> Behavioral properties

A+

> Support for multiple in-flight commands

Nobody ever used this or if they did, they did it by-accident. Limiting to 1x/time is Hard to do when multiple in-flight commands were enabled, and it was the thing that 99% of people want. Supporting multiple in-flight commands with the current setup is easy:

```cs
myCoolCommand.SelectMany(x => DoAGoogleSearch(x)).Subscribe();
```

Constraining asynchrony is *not* because you have to reach into the internals of Command (or you create this circular dependency)

flagbug:
Could we implement `ICommand` as explicit interface, so the `Execue(object)` method doesn't get exposed, but XAML binding can still bind to the `ReactiveCommand`?

mteper:
:+1: to :fire: `ICommand`, typed `ReactiveCommand` FTW. I appreciate what @paulcbetts is saying about keeping it backwards compatible for XAML folks, so if there is a way to do that (e.g. what @flagbug suggested), that would be win-win.

Async-only RxCmd is not really that painful, and having a single contract makes things less confusing, so :+1: for that too.

paulcbetts:
> Could we implement ICommand as explicit interface, so the Execute(object) method doesn't get exposed, but XAML binding can still bind to the ReactiveCommand?

This might work, would have to look into it more

kentcb:
Thanks for the feedback all.

An explicit implementation of `ICommand` was actually my backup plan if people didn't like the split as much as I do, and that seems to be the case so I'll give it a go. I'll do it in a partial class so the split is at least maintained lexically if not in the binary itself. I'm pretty sure we can keep the typed param too - it just involves casting the parameter passed to `ICommand.Execute`.

> being able to call ToPlatform means everyone has to do a bunch of shit work making dummy versions of their properties

My inclination would actually be to implement my own `CommandBindingMarkupExtension` which I could use like this:

```XML
<Button Command="{xxx:CommandBinding TheCommand}"/>
```

I know not all XAML platforms are quite this flexible, but the ones I like are :)

> I don't think it's a good idea to split these back up again

I don't think we have much choice because, for example, the binding infrastructure relies on being able to ascertain `CanExecute` in a non-generic context. The old implementation relied on the non-generic interfaces for this, so we need to some equivalent. Also, I think it's actually quite useful for folks to be able to treat heterogeneous reactive commands as a composite.

I don't think we'll need quite as many as we have now that we're ditching synchronous commands. And the base types are abstract, so that at least avoids people trying to instantiate them.

Alright, I make these changes soonish and ping the Slack channel again. Cheers.

kentcb:
All: I've just updated this PR with the changes discussed. Before I go and replace the old implementation with the new, could you please eyeball this and voice any concerns. Thanks.

mteper:
Looks nice and clean, :+1:

moswald:
Yeah, I think it looks good.

onovotny:
I would NOT like this change without `ICommand`. Custom markup extensions are not supported in WinStore apps so this hoses UWP. Furthermore `x:Bind` really needs the type to be assignable.

paulcbetts:
@onovotny afaik this does have an explicit ICommand implementation

onovotny:
Long as `ICommand` exists then I'm :+1:. I was never a huge fan of the RxUI 6 ReactiveCommand, so a rethink is welcome by me.

moswald:
I just did a quick test with UWP - `{x:Bind ViewModel.SomeProperty}` works fine, although Intellisense complains that `ViewModel` is ambiguous. I would want to fix that, but don't have a good idea other than modifying `ReactiveUserControl` to have a `BindingRoot` property like this:

```
public TViewModel BindingRoot => ViewModel;
```

moswald:
Additionally, `CommandParameter` is supported and is even type-safe!

The following are all _compiler errors_ (!):
* your `ReactiveCommand` expects a parameter, but you don't supply `CommandParameter` in your xaml.
* your `ReactiveCommand` _doesn't_ expect a parameter, but you _do_ supply a `CommandParameter`.
* your `ReactiveCommand` expects a parameter of type `int`, but your `CommandParameter` value is a `string` (probably very likely).

flagbug:
@moswald That type-safety only works for UWP apps with `x:Bind`, right?

moswald:
That's all I checked, I forgot to try the old-style `Binding`.  :sob: 

@kentcb these are things that should be mentioned in the docs you're writing.

mteper:
Question: if CreateAsync[Task/Observable] are just wrappers around create, that (I assume) are there to ease migration, can we mark them Obsolete so as to nudge people to new API?

kentcb:
@mteper they're not just wrappers around `Create`. Originally, I _did_ want to have a single `Create` method (and my initial PR reflected this desire) that is called regardless of what you're creating with (`Action`, `Func<TResult>`, `Action<TParam>`, `Func<IObservable<TResult>>` etc). However, C# is unfortunately unable to disambiguate calls. eg:

```cs
// should this create via a Func<TResult> or a Func<IObservable<TResult>>?
ReactiveCommand.Create(() => Observable.Return(42));
```

Hence, we need to have distinct names to suit the different scenarios. And since those names were already established, I went with them to reduce migration pain.

mteper:
@kentcb I see, makes sense. So do we still need all three (`Create`, `CreateAsyncTask`, `CreateAsyncObservable`) or could we collapse down to two, `Create` and `CreateAsyncObservable`, with the latter serving the `IObservable` scenario?

FWIW, the current names (`CreateAsyncTask`, `CreateAsyncObservable`) have always been a bit confusing to me because they imply that we are creating a Task or an Observable, which is not the case. I think something like `CreateWithTask` / `CreateWithObservable` would have been clearer.

I am not sure if there is value in renaming / obsoleting, but if others have also been tripped up by this (especially when first ramping up on RxUI), now perhaps would be a good time to do it.

kentcb:
I tend to agree about the names being a little confusing. I like `CreateWithXxx`, or perhaps `CreateAsyncWithXxx`. Let's float it in Slack and see how well it's received.

shiftkey:
Great work @kentcb, glad to see this finally land!

PR #974, opened 2015-10-28
Forward exceptions on creating execute observable to ThrownExceptions

### LINKED ISSUE #973: Exception swallowed in command if during executeAsync
Consider this code:
```csharp
var command = ReactiveCommand.CreateAsyncObservable(_ => ThrowImmediately());
// ...
private IObservable<Unit> ThrowImmediately()
{
    throw new InvalidOperationException();
}
```
`ReactiveCommand.ExecuteAsync` creates an observable using:
```csharp
var ret = Observable.Create<T>(subj => {
    if (Interlocked.Increment(ref inflightCount) == 1) {
        isExecuting.OnNext(true);
    }
    // ...
    var disp = executeAsync(parameter)
        .ObserveOn(scheduler)... // ...
```
The framework calls `command.Execute`. The code for this is currently:
```csharp
ExecuteAsync(parameter).Catch(Observable.Empty<T>()).Subscribe();
```
So it seems this happens:
- the framework calls `ReactiveCommand.Execute`
- `ReactiveCommand.ExecuteAsync` sets `IsExecuting = true`, disabling the command.  It returns an observable containing the exception.
- The exception is swallowed by the catch, and as the disposable that would set `IsExecuting = false` is never returned the command is stuck as disabled in the UI.

It seems this may relate to the changes in #686.  Is this a bug?  What is the expected behaviour?

csmager:
If the `executeAsync` func throws an exception while creating the observable (i.e. synchronously), then this exception is lost by frameworks relying on `ICommand.Execute` - this swallows any exceptions returned from `ReactiveCommand.ExecuteAsync`.

This change ensures these exceptions are forwarded to `ThrownExceptions`, which would seem to be the expected behaviour.

Fixes #973

### DIFF
--- a/ReactiveUI.Tests/ReactiveCommandTest.cs
+++ b/ReactiveUI.Tests/ReactiveCommandTest.cs
@@ -184,32 +184,101 @@ namespace ReactiveUI.Tests
         [Fact]
         public async Task ExecuteAsyncThrowsExceptionOnError()
         {
-            var command = ReactiveCommand.CreateAsyncObservable(_ =>
-                Observable.Throw<Unit>(new Exception("Aieeeee!")));
+            var exception = new Exception("Aieeeee!");
 
-            var exceptions = command.ThrownExceptions.CreateCollection();
+            var command = ReactiveCommand.CreateAsyncObservable(_ => throwSync(exception));
 
-            bool failed = false;
-            try {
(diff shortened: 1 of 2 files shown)


flagbug:
:+1: LGTM

ghuntley:
Thanks for the contribution @csmager

Issue #1005, opened 2015-12-25
#yourfirstpr - Update our template and remove thousands of warnings from our build logs.


ghuntley:
Instead of fixing this issue immediately, we are leaving this [issue open and up for grabs](http://up-for-grabs.net) for someone [who has never contributed to open source before](http://www.firsttimersonly.com/). We will not accept solutions to this issue from non-first-timers to this project or open-source, [we will check](http://firstpr.me/).

# Problem

We lack XML documentation on our ReactiveUI-Events package which results in an excessive amount of warnings when building ReactiveUI on our continuous integration environment.

It really doesn't make sense to go back and add the missing XML documentation as these projects are automatically generated by a [moustache template](https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI.Events/Events.mustache) and a [console application](https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI.Events/EventBuilder.cs).

Additionally it does not make sense to add dummy XML documentation so the approach we are proposing is update the template that generates the projects to [supress the compiler warning](https://msdn.microsoft.com/en-us/library/edzzzth4%28v=vs.90%29.aspx?f=255&MSPPError=-2147217396).

# Suggested Resolution

On https://github.com/reactiveui/ReactiveUI/blob/master/ReactiveUI.Events/Events.mustache#L14

At the top of the file you will find the following section

	{{#Namespaces}}
	namespace {{Name}}

Change it to be

	#pragma warning disable 1591
	{{#Namespaces}}
	namespace {{Name}}

At the bottom of the file you will find the following section

	{{/Types}}
	}
	{{/DelegateNamespaces}}

Change it to be

	{{/Types}}
	}
	#pragma warning restore 1591
	{{/DelegateNamespaces}}

Submit your first pull-request and when it gets merged, we will send a tweet to  [@yourfirstpr](https://twitter.com/yourfirstpr) in celebration :)

# What you'll be fixing

In our build logs #1006 we have thousands of these annoying and erroneous warning messages:
	
	Events.cs(2506,16): warning CS1591: Missing XML comment for publicly visible type or member 'TimelineEvents.TimelineEvents(Timeline)' [C:\projects\reactiveui\ReactiveUI.Events\ReactiveUI.Events_Net45.csproj]
[...]

PR #1007, opened 2015-12-26
Fixing Issue #1005 as per Dads Suggestion

oliverhuntley:
Fixing issue https://github.com/reactiveui/ReactiveUI/issues/1005 
Dad said i should Give a pull request a go...


oliverhuntley:
@ghuntley Did I do you proud?

PR #1008, opened 2015-12-26
Portable PlatformRegistrations should work from unit tests

OlexaLe:
When use Xamarin Studio on OS X it is not possible to use ReactiveUI in Unit Testing Library projects. I am able to add "reactiveui-testing" nuget package, but during ReactiveUI initialisation it uses PlatformRegistrations from PortableLibraryStubs and crashes. As we do not need any additional registrations during unit tests I propose to allow ReactiveUI work when NUnit in loaded. This assumption smells like a hack, but at least this allows us to use ReactiveUI in our production process. Thanks!

### DIFF
--- a/ReactiveUI/PortableLibraryStubs.cs
+++ b/ReactiveUI/PortableLibraryStubs.cs
@@ -1,8 +1,6 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using System.Reflection;
 
 namespace ReactiveUI
 {
@@ -10,7 +8,16 @@ namespace ReactiveUI
     {
         public void Register(Action<Func<object>, Type> registerFunction)
         {
+            if (IsRunningFromNUnit()) return;
             throw new Exception("You are referencing the Portable version of ReactiveUI in an App. Reference the platform-specific version.");
+        }
+
+        private bool IsRunningFromNUnit()
+        {
+            var currentdomain = typeof(string).GetTypeInfo().Assembly.GetType("System.AppDomain").GetRuntimeProperty("CurrentDomain").GetMethod.Invoke(null, new object[] { });
+            var getassemblies = currentdomain.GetType().GetRuntimeMethod("GetAssemblies", new Type[]{ });
+            var assemblies = getassemblies.Invoke(currentdomain, new object[]{ }) as Assembly[];
+            return assemblies.Any(assembly => assembly.FullName.ToLowerInvariant().StartsWith("nunit.framework"));
         }
     }
 }
--- a/ReactiveUI/PortableLibraryStubs.cs
+++ b/ReactiveUI/PortableLibraryStubs.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Linq;
-using System.Reflection;
 
 namespace ReactiveUI
 {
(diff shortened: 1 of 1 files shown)


ReactiveUI/PortableLibraryStubs.cs
```diff
@@ -10,7 +8,16 @@ public class PlatformRegistrations : IWantsToRegisterStuff
     {
         public void Register(Action<Func<object>, Type> registerFunction)
         {
+            if (IsRunningFromNUnit()) return;
```
flagbug: Use `Splat.ModeDetector.IsInUnitTestRunner()` for this

OlexaLe: Unfortunately it does not work on my test project. According to Splat source code https://github.com/paulcbetts/splat/blob/master/Splat/ModeDetector.cs there is no way on how to detect a unit test, but my solution works.

ghuntley:
Can you please share with us some of your unit tests? What are you testing?

OlexaLe:
Sure. Basically it hangs on any ReactiveUI-specific test. Here is one:
   ```c#
[TestFixture]
    public class Test
    {
        SampleViewModel _viewModel;

        [TestCase(null, false)]
        [TestCase("", false)]
        [TestCase("abc", true)]
        public void TestCanExecute(string paramValue, bool expectedCanExecute)
        {
            _viewModel = new SampleViewModel();
            _viewModel.SomeParameter = paramValue;
            Assert.AreEqual(expectedCanExecute, _viewModel.DoSomething.CanExecute(null));
        }
    }

    public class SampleViewModel : ReactiveObject
    {
        public ReactiveCommand<object> DoSomething { get; }

        private string _someParameter;

        public string SomeParameter
        {
            get { return _someParameter; }
            set { this.RaiseAndSetIfChanged(ref _someParameter, value); }
        }

        public SampleViewModel()
        {
            var canExecute = this.WhenAny(vm => vm.SomeParameter, x => !string.IsNullOrWhiteSpace(x.Value));
            DoSomething = ReactiveCommand.Create(canExecute);
        }
    }
```
Thanks!

**Comment by kentcb** [comment]
See #807 and also http://kent-boogaart.com/blog/using-the-visual-studio-test-runner-for-mobile-development/.

File: ReactiveUI/PortableLibraryStubs.cs [hunk 2]
```diff
@@ -10,7 +8,16 @@ public class PlatformRegistrations : IWantsToRegisterStuff
...
             throw new Exception("You are referencing the Portable version of ReactiveUI in an App. Reference the platform-specific version.");
         }
+
+        private bool IsRunningFromNUnit()
+        {
+            var currentdomain = typeof(string).GetTypeInfo().Assembly.GetType("System.AppDomain").GetRuntimeProperty("CurrentDomain").GetMethod.Invoke(null, new object[] { });
```
shiftkey: While it sounds like you're getting a bunch of feedback from the core team already, I thought I'd share some history on this.

In the early days, Splat did something not unlike what you're looking to bring in: https://github.com/paulcbetts/splat/blob/227ee96227df6fc542eea4c1b6eb1ffdcb7edd49/Splat/PlatformModeDetector.cs

This quickly devolved into supporting a lot of edge case scenarios - you can see all the test runners that were supported at the time there.

My advice here would be to wire up your own rules within the context of your app - it looks like you've got the rules sorted, it just needs to be wired up to Splat (and thus into ReactiveUI):

```csharp
public class NUnitModeDetector : IModeDetector
{
    // cache this computed value so that it's not re-evaluated
    Lazy<bool> checkForNUnit = new Lazy<bool>(IsRunningFromNUnit);
    public bool? InUnitTestRunner() { return checkForNUnit.Value; }
    public  bool? InDesignMode() { return null; }
    private bool IsRunningFromNUnit() {
        // this code goes here
    }
}

// elsewhere, in your ReactiveUI setup code
ModeDetector.OverrideModeDetector(new NUnitModeDetector());
```

Apologies if this code doesn't compile, but hopefully you get the idea...

cc @ghuntley remind me to document this somewhere when I'm back from holidays

OlexaLe:
First of all I would thank you all guys for your help. It is really nice to see that ReactiveUI community is so active.

Combining @shiftkey and @flagbug advices I was able to make it work. Still need that condition in PlatformRegistrations Portable implementation, but I've replaced my ugly IsRunningFromNUnit() method with Splat.ModeDetector.InUnitTestRunner() and now able to override ModeDetector for unit tests.

ghuntley:
This is resolved in the upcoming v8 release. We have removed all WPF references from the main `reactiveui` package.

PR #1013, opened 2016-01-03
Have a lot of Views? Hate registering each one? This PR is for you!

moswald:
The many lines of:

```csharp
resolver.Register(() => new FooView(), IViewFor<FooViewModel>);
resolver.Register(() => new BarView(), IViewFor<BarViewModel>);
... <etc> ...
```

Can be reduced to: `resolver.RegisterViewsForViewModels(typeof(FooView).Assembly)`

### DIFF
--- a/.hgeol
+++ b/.hgeol
@@ -1,0 +1,2 @@
+[repository]
+native = LF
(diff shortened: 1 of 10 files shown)


martingemme:
Does this means we have to name our ViewModel exactly <Predicate>ViewModel = <Pedicate>View?

PR #1015, opened 2016-01-04
#1014 resolve base interface types correctly, and add unit tests to cover more cases

### LINKED ISSUE #1014: DefaultViewLocator doesn't resolve IViewFor<IFooBarViewModel>
According to the comments in DefaultViewLocator, it looks for IViewFor<IFooBarViewModel> for a ViewModel of type FooBarViewModel. This doesn't appear to be the case, so the test below fails

```C#
[Fact]
public void ResolveByInterfaceType()
{
    var resolver = new ModernDependencyResolver();

    resolver.InitializeSplat();
    resolver.Register(() => new FooBarView(), typeof(IViewFor<IFooBarViewModel>));

    using (resolver.WithResolver()) {
        var fixture = new DefaultViewLocator();
        var vm = new FooBarViewModel(null);

        var result = fixture.ResolveView(vm);
        this.Log().Info(result.GetType().FullName);
        Assert.IsType<FooBarView>(result);
    }
}

namespace Foobar.ViewModels
{
    public interface IFooBarViewModel : IRoutableViewModel {}

    public class FooBarViewModel : ReactiveObject, IFooBarViewModel 
    {
        public string UrlPathSegment { get { return "foo"; } }
        public IScreen HostScreen { get; private set; }

        public FooBarViewModel(IScreen hostScreen)
        {
            HostScreen = hostScreen;
        }
    }
}

namespace Foobar.Views
{
    using ViewModels;

    public class FooBarView : IViewFor<IFooBarViewModel> 
    {
        object IViewFor.ViewModel { get { return ViewModel; } set { ViewModel = (IFooBarViewModel) value; } }
        public IFooBarViewModel ViewModel { get; set; }
    }
}

```
I've had a crack at fixing it, but the only way I can see is to attempt to resolve an IViewFor for every interface the ViewModel implements and that seems a little ugly (it might be reasonable to restrict that to interfaces that implement IRoutableViewModel which is a little better though?).

olevett:
Fixes #1014

### DIFF
--- a/ReactiveUI.Tests/RxRouting.cs
+++ b/ReactiveUI.Tests/RxRouting.cs
@@ -1,17 +1,14 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using ReactiveUI;
-using ReactiveUI.Routing;
-using Xunit;
+﻿using ReactiveUI;
 using Splat;
+using Xunit;
 
 namespace Foobar.ViewModels
 {
     public interface IFooBarViewModel : IRoutableViewModel {}
 
     public interface IBazViewModel : IRoutableViewModel {}
+
+    public interface IQuxViewModel : IRoutableViewModel { }
 
     public class FooBarViewModel : ReactiveObject, IFooBarViewModel 
     {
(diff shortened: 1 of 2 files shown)


ghuntley:
Thanks for the high quality bug report @olevett!

We have been talking in slack about it, specifically the fact you submitted a failing test as part of the initial bug report and that you proceeded to resolve the issue via a PR.

You should come and join us in Slack, see the README.md for details - for now I've sent you an invitation to join the ReactiveUI community which when accepted will provide the ability to wear our badge on your profile.

PR #1017, opened 2016-01-09
BindTo helpful error

mattrrichard:
If you `observable.BindTo(null, x => x.SomeProperty)`, the binding will fail when it tries to execute with the somewhat cryptic error message "Non-static method requires a target" and a stack trace that doesn't point you to where you Did It Wrong (tm). 

I have `BindTo` throw an `ArgumentNullException` if its `target` paramter is `null` with what I hope is a helpful error message. 

I'm pretty new to the framework myself so please let me know if you'd like me to change anything.

### DIFF
--- a/ReactiveUI/PropertyBinding.cs
+++ b/ReactiveUI/PropertyBinding.cs
@@ -906,6 +906,11 @@ namespace ReactiveUI
             object conversionHint = null,
             IBindingTypeConverter vmToViewConverterOverride = null)
         {
+
+            if (target == null) {
+                throw new ArgumentNullException("target", "Can't bind to a property on null.  To fix this, you probably want BindTo(this, ...) isntead. ");
+            }
+
             var viewExpression = Reflection.Rewrite(property.Body);
 
             var ret = evalBindingHooks(This, target, null, viewExpression, BindingDirection.OneWay);
--- a/ReactiveUI/PropertyBinding.cs
+++ b/ReactiveUI/PropertyBinding.cs
@@ -908,7 +908,7 @@ namespace ReactiveUI
         {
 
             if (target == null) {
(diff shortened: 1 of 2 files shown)


ReactiveUI/PropertyBinding.cs
```diff
@@ -906,6 +906,11 @@ public class PropertyBinderImplementation : IPropertyBinderImplementation
             object conversionHint = null,
             IBindingTypeConverter vmToViewConverterOverride = null)
         {
+
+            if (target == null) {
+                throw new ArgumentNullException("target", "Can't bind to a property on null.  To fix this, you probably want BindTo(this, ...) isntead. ");
```
kentcb: Hmmm, I think simply throwing an `ArgumentNullException` for `target` will suffice. Trying to be more helpful here feels a little misleading/confusing.

mattrrichard:
In particular, I wasn't positive where the best place to _put_ the check I added.  It's in `PropertyBinderImplementation.BindTo` currently which is sort of the "middle" of the call chain. I can see reasons to put it both "higher"  (e.g. at `BindingMixins.BindTo) and "lower" (e.g. at `PropertyBinderImplementation.bindToDirect`).

kentcb:
In this case I think lower is fine because the parameter name remains the same. That way we can avoid checking it twice.

mattrrichard:
Would you go all the way down or leave it where it is now?

kentcb:
I _think_ where it is now is :+1: because it's the final public API. From there it just calls internal/private members...?

mattrrichard:
Next lower is an internal method, yes.  I'll leave it where is.  Thanks!

kentcb:
Also, you'll have to target rxui7-master branch for this to be merged. I'm afraid that means closing this PR and re-opening another :(

mattrrichard:
d'oh

PR #1018, opened 2016-01-09
BindTo helpful error when binding to null.

mattrrichard:
If you `observable.BindTo(null, x => x.SomeProperty)`, the binding will fail when it tries to execute with the somewhat cryptic error message "Non-static method requires a target" and a stack trace that doesn't point you to where you Did It Wrong (tm).  Instead, I have `BindTo` throw an `ArgumentNullException` if its `target` paramter is `null`.

See #1017 for previous discussion.


kentcb:
Thanks Matt!

PR #999, opened 2015-12-18
ActOnEveryObject can use IReactiveCollection<T>

arkein:
Fix for #989

Issue #989, opened 2015-11-27
ActOnEveryObject can use IReactiveCollection<T> instead of ReactiveList<T>


yar-shukan:

flagbug:
@yar-shukan Can you submit a pull request for this?

yar-shukan:
Sure, will do.

PR #1021, opened 2016-01-11
ActOnEveryObject<T> can accept IReadOnlyReactiveCollection<T>

arkein:
Enhancement for #989 
(this should target the proper branch now, previous PR was #999)

### DIFF
--- a/ReactiveUI/AutoPersistHelper.cs
+++ b/ReactiveUI/AutoPersistHelper.cs
@@ -106,7 +106,7 @@ namespace ReactiveUI
         /// if an object is constantly changing, it is possible that it will never
         /// be saved.</param>
         /// <returns>A Disposable to disable automatic persistence.</returns>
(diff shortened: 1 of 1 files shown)


kentcb:
I think you might have missed my comment on the previous PR. `IReadOnlyReactiveCollection` extends `IReactiveCollection`, so you've actually _narrowed_ the utility of this method by taking `IReadOnlyReactiveCollection` rather than just `IReactiveCollection`. Do you agree @flagbug ?

flagbug:
@kentcb is right, I thought the implementation was the other way around, so extending `IReactiveCollection` is the right way to go

shiftkey:
:thumbsup: `IReactiveCollection`

arkein:
Read, facepalmed, fixed. I guess there's no more room left to do another mistake.

kentcb:
Thanks @arkein!

PR #1022, opened 2016-01-13
Added ViewAttributes to Xam.Mac projects (projects were not compiling)

mteper:


ghuntley:
Confirmed fixes Xam.Mac - LGTM. 

iOS still don't build because of the same issue but I'll fix that in the morning.

PR #1032, opened 2016-01-16
Reactive Xamarin.Forms pages implementations

michaelstonis:
Includes additional Xamarin.Forms pages setup for Reactive UI. 

Also, I bumped up the version of Xamarin.Forms referenced, but this can be excluded. There have been some API changes since 1.5, but nothing too major.


ghuntley:
LGTM.

kentcb:
Thanks @michaelstonis . Unfortunately, the XamForms events project file includes a path to the XF binary under the packages directory (which includes the version). Could you please update that reference to use the new version too? Otherwise, XF events won't build.

kentcb:
This is the line I was talking about, but also looks like it's necessary to upgrade the XF packages for that project.

michaelstonis:
Upgraded. Totally missed the events package. It had been broken for so long... Excited to have it back.

kentcb:
Thanks @michaelstonis !

PR #1042, opened 2016-01-18
Rxui5 validation

ryanramadhan:


moswald:
At first glance, this looks great. I'll review it more in depth later today. However, I want to get the `uap` branch merged into `rxui7-master` before this one. I think the resulting merge conflicts will end up simpler when applied in that order.

PR #1045, opened 2016-01-19
Add Universal Windows Platform (UWP) support.

moswald:
Recreated from #829 to target `rxui7-master`.

This PR adds UWP support to RxUI.

## TODO:

- [x] Add activation logic on the `Loading` event instead of the `Loaded` event.
- [x] Create an `ObservableForProperty` class that uses the new dependency property change notification APIs.
- [x] Fix NuGet scripts.
- [x] Figure out the `WINRT` vs `NETFX_CORE` stuff.

~~- [ ] Fix runtime directives (.rd.xml) files.~~
The Dot Net Native toolchain has _serious_ trouble with RxUI, and no amount of runtime directives will get it working. There's just too many layers of reflection for it to deal with. We'll have to note this in the documentation, and work toward a better fix as time goes on.

Note: the rd.xml files still need to be checked as the tools will look for them.


ghuntley:
Matt - what needs fixing in the NuGet scripts? Don't spend too much time on them as I'll be rewriting most of that junk as part of CI.

moswald:
All of it? As in, UWP NuGet packages are not built. @onovotny has volunteered to take a look at it.

moswald:
If you're interested in dropping the NuGet package requirement because it may get reworked later, that's fine with me. I want to get this branch merged ASAP.

onovotny:
My changes should be complete now. I've updated the `MakeRelease.ps1` script to download the latest v3 `nuget.exe` and package the `uap10.0` stuff. Still needs testing/validation, but the guts should be there.

moswald:
I have validated that `MakeRelease.ps1` builds the `uap10.0` packages (after a couple of minor changes - mainly `System.Net.WebClient.DownloadFile` wouldn't write to the current folder without a full path). 

Can someone who knows better than I run it and make sure the folder structure/contents within the `.nupkg` is correct?

PR #1055, opened 2016-01-25
Improve interactions implementation

kentcb:
This PR improves the interactions implementation by:

* generalizing the name to "interactions" rather than "user interactions"
* splitting the responsibility for the distribution of interactions into a separate, non-static, class
* lots of smaller improvements to round edges in the API

See the documentation for this feature [here](http://docs.reactiveui.net/en/user-guide/interactions/index.html).

TODO:
- [x] Decide on naming


shiftkey:
:eyes:

kentcb:
Hmmm, weird. git appears to be picking up my work email sometimes, other times not :/

moswald:
I dumped this into Slack, but since there's a little pushback on the name `InteractionData`, what about `InteractionPayload` instead?

mteper:
So, I was thinking about interactions, and in particular about the long-winded declaration statement that we currently have (`Interaction<InteractionData<blah>>`) , and like Kent, was trying to categorize the moving pieces in my head to better frame the problem.

For reference, this is Kent's categorization lifted from the Slack RxUI channel:
> 1. An abstract base class for all individual interactions (currently `InteractionData`). Has an `IsHandled` property.
> 2. A class for individual interactions with a specific result type (currently `InteractionData<TResult>`. Extends 1 and adds `GetResult` and `SetResult` methods. Application developers are encouraged to subclass this class if they need to add extra data to interactions.
> 3. A class that provides the coordination and distribution mechanism for interactions (currently `Interaction`). Has `RegisterHandler` methods (for views) as well as `Handle` method (for view models)

I came up with a slightly different set:
 a. interaction “broker” -- this matches Kent's no. 3, the class that provides coordination
 b. request data -- information that the ViewModel sends along for the ride to interaction handler(s)
 c. result of the interaction -- something concrete that the ViewModel expects back from the interaction
 d. metadata about the handling of the interaction (e.g. was it handled? how? by whom? etc.)

What we have today combines items (b)-(d) into one class, `InteractionData<TResult>`, where only the result is variable, and user is expected to subclass it if they need something different (e.g. sending an Exception along for the ride to handler(s) or getting more context than just `IsHandled`). It feels to me, however, that there's another way to dice this up:

 - `TInput`: the input to the interaction (for example, this could be a `string` or `Exception` or custom class)
 - `TResult`: the result coming back from the interaction (for example, a `FileInfo` from file picker). I was initially thinking that the interaction handling state would have to be a separate, third, generic type, but as I thought through actual scenarios, I concluded we don't need it to be separate. Anything the developer wants to send back, they can include in the result. The only thing the interaction needs them to say is whether the given handler has completed handling of this interaction. Which brings us to the last new component,
 - `InteractionResponse<TResult>`: the type that is returned from the interaction handlers. Now that we've separated out the interaction input, this type can be fixed and no longer needs subclassing.
 - `Interaction<TInput, TResult>`: this is the interaction itself. Maybe we can even throw in a couple of simpler concrete implementations,
  - `InputInteraction<TInput>` that maps to `Interaction<TInput, object>` where we don't care about the result.
  - `ResultInteraction<TResult>` that maps to `Interaction<object, TResult>` where we don't care about input to the interaction and just want the result back.

I see the following benefits to this approach:
 - we no longer need to argue over naming of `InteractionData` :)
 - we have a more concise declaration for the interaction itself
 - we no longer require the user to subclass anything to implement interactions, ever
 - we are clear and explicit about what goes in and what comes out

To wrap it all up, here's a usage example:
```cs
{
  private string myFile = "My.precious";

  public ReactiveCommand<Unit, Unit> UpdateFileContents;
  public Interaction<string, FileInfo> PromptForMissingFile => new Interaction<String, FileInfo>();

  public MyViewModel()
  {
    UpdateFileContents = ReactiveCommand.Create(async () =>
        {
          // we expect my file to always exist, but just in case, we sanity check and
          // prompt the user if it's missing
          if (!File.Exists(myFile))
          {
            var response = await PromptForMissingFile.Handle("Darn! There was supposed to be this file, but it's missing, please tell me where it went!");
            if (response.IsHandled) myFile = response.Result.FullName;
            else { /* beats me, this is just an example! */ }
          }

          // go on with updating the file
        });
  }
}

public class View
{
  // wherever you are doing view setup
  {
    ViewModel.PromptForMissingFile.RegisterHandler(input =>
       {
          // throw up a file selector with my input message and capture the result
          var fileInfo = // ...

          return new InteractionResponse(fileInfo, /* handled */ true);
       }
  }
}
```

P.S. I am still partial to `interaction.Invoke()` over `interaction.Handle()`, but only a little bit. I understand the explanation Kent put forward and I can learn to live with it if I have to. :)

moswald:
:+1: Definitely an improvement in my mind.

mteper:
Oops, realized there was a problem with the View part of my example, in that the handler really should not be returning the response (this will make it harder to coordinate multiple handlers and will make it harder for a handler to choose to do nothing). Instead, perhaps this would be more sensible:

```cs
public class View
{
  // wherever you are doing view setup
  {
    ViewModel.PromptForMissingFile.RegisterHandler(input, context =>
       {
          // throw up a file selector with my input message and capture the result
          var fileInfo = // ...

          context.SetResult(fileInfo);
          // or, if you *did not* want to mark interaction handled, you would do this
          // context.SetResult(fileInfo, handled: false);
       }
  }
}
```

kentcb:
@mteper Thanks for the suggestion. I like the idea of formalizing the input and output (though I prefer request/response terminology). I'm not a huge fan of forcing handlers to return something. This, in particular, makes no sense to me:

```cs
context.SetResult(fileInfo, handled: false);
```

Setting a result _is_ handling the interaction, so this is a contradictory API. However, I noticed you introduced another concept in your most recent post: the context. To me, it would make sense for the context to include a `Request` property and a `SetResponse` method. If handlers don't want to handle the interaction, they simply don't call the `SetResponse` method. Is this what you intended to convey? If so, I believe this means we'd have:

```
Interaction<TRequest, TResponse>
 + RegisterHandler(Func<InteractionContext<TRequest, TResponse>, IObservable<Unit>>) // plus other overloads
 + Handle(TRequest input) : IObservable<TResponse>

InteractionContext<TRequest, TResponse>
+ IsHandled : bool
+ Request : TRequest (get)
+ SetResponse(TResponse)
+ GetResponse() : TResponse

UnhandledInteractionException<TRequest, TResponse>
+ Interaction : Interaction<TRequest, TResponse>
+ Request : TRequest
```

And usage would resemble:

```cs
// view model
private readonly Interaction<string, bool> confirmDelete = new Interaction<string, bool>();

public Interaction<string, bool> ConfirmDelete => this.confirmDelete;

private async Task DeleteFileAsync()
{
    var fileName = ...;

    // this will throw if nothing handles the interaction
    var confirm = await this.confirmDelete.Handle(fileName);

    if (confirm)
    {
        // delete the file
    }
}

// view
this
    .ViewModel
    .ConfirmDelete
    .RegisterHandler(
        async context =>
        {
            var result = await this.DisplayMessageAsync("Confirm Delete", $"Delete file '{context.Request}'?", "YES", "NO");

            // handler could simply not make this call if it does not wish to handle the interaction
            context.SetResponse(result);
        });
```

Does this sound good to you?

mteper:
@kentcb,

### Terminology
My thinking started out with "request" / "response", but I don't think these terms quite fit. In particular the thing going in is not a request per se, it's just data, it's "input". The thing coming out is a little different in what I had in mind vs. your post above. I was (perhaps mistakenly) assuming that you wanted the response to carry back some interaction processing information (e.g. last value of IsHandled) back to the caller, and so I pictured the interaction returning an instance of `InteractionResponse<TResult>`. If that's not the case (and, again, I was just trying to fit into my perception of what you had -- I don't have a compelling reason for things to be that way) then we can just return the value. If we do so, then back to terminology, that's again more of a "result" than a "response". In other words, just data.

That said, I won't be terribly sad whether these guys are named Input/Result or Request/Response. :smile: 

### Handling
This might have been another misunderstanding on my part. I was assuming that you wanted to support a situation where one handler could do something and return without marking the interaction handled and pass control over to the next handler in the chain. I am OK not doing things that way, too. However, there *is* one sticking point. In the event that we don't care about the result of the interaction, we just want to communicate something forward, we still need a way for the handler to set the interaction as handled, right? That could be done either via `context.SetResult(null);` or `context.SetHandled()`. I kind of like the explicitness of the latter, but I am also concerned that having both methods might confuse people into thinking they need to both set a result *and* then call `SetHandled`. We could perhaps get around that explicitly with renaming `SetResult` to `SetHandledWithResult`?

### Exception
When does the exception get raised? Is it when there are no handlers registered or when no handler sets the result / handled flag?

kentcb:
@mteper 

## Terminology

I know what you mean. My problem with input/output was the other side of the equation. Handlers would need to call `SetOutput`, which feels a bit wrong to me. But whatever, I'll run with Input/Output and see how it pans out :+1: 

## Handling

Handlers are not in control of invoking the next handler - the interaction itself does this. The implementation that comes with RxUI will invoke handlers in reverse order and cease when any handler handles the interaction. But subclasses of `Interaction<TInput, TOutput>` will be able to change that behavior if desired.

## Exception

The exception is raised if the interaction goes unhandled. That could be because no handlers are registered for it, or none of the registered handlers provided an output.

mteper:
@kentcb 

> Handlers are not in control of invoking the next handler

Now I am confused. As long as mere execution of a handler is not equivalent to "handling" the interaction, a handler is, in fact, able to exercise control. When we say that a handler is responsible for explicitly saying that interaction is complete/handled, and allow it to do nothing, we enable it to cause the execution to flow to next handler, no?

kentcb:
@mteper sure, but I said the handler does not invoke the next handler, and that's true. It merely returns without setting an output for the interaction.

> we just want to communicate something forward

What is the something? If the output of the interaction is unimportant, you'd just use `Unit` as `TOutput` and therefore the handler would do this to mark the interaction as handled:

```cs
context.SetOutput(Unit.Default);
```

kentcb:
FYI, updated to formalize the input and output types. Docs also updated.

mteper:
:+1:

Very last API question: is there a good reason for `GetOutput()` to not be a read only property?

kentcb:
@mteper yes, because .NET guidelines stipulate that property getters should never throw.

mteper:
Huh, cool, wasn't aware. In that case, you've got my vote to :ship:

pmbanka:
Just my $0.02:

* I really like the new API, because it makes the feature look similar to the `ReactiveCommand<TIn, TOut>`. Definitely a step forward into correct direction.
* :+1: for `TRequest`/`TResponse`. In 99% of cases, a *real user* will be requested to provide some response, so the names are quite a natural match.
* :+1: to `Invoke` instead of `Handle`. In 99% of cases, this will be invoking some message box/other visible UI element prompting *real user* for a response. I want this `Interaction` with the user to be `Invoke`d, instead of `Handle`d (handled by *whom*?). (*disclaimer - I'm not a native speaker*)
* :+1:  for burning `InteractionResponse<TResult>` and throwing exception instead, as @kentcb explained. Also, I like the idea and naming of `context.SetResult(Unit.Default);`
* :fire:ing `Global` case is even more sensible now

Issue #1043, opened 2016-01-18
Decouple ReactiveUI and ReactiveUI events


ghuntley:
# The Plan

The package `reactiveui-events` is used in the platform specific unit test runner apps. Cannot progress CI builds until this dependency is unlinked. Here's the plan:

- [x] local `reactiveui-events` removed from all solutions.
- [x] eventbuilder xplat.
- [x] all `msbuild/mdtool` prebuild steps :fire:
- [x] :fire: any notion of VS or XS project (ie.. iOS64.csproj and iOS64_XS.csproj)
- [x] :fire: iOS legacy (non unified)
- [x] Rename iOS64 to iOS
- [x] :fire: Mac legacy (non unified)
- [x] :fire: WinRT 8.0
- [x] Build solution for `reactiveui-events` created.
- [x] Solution building for Xamarin iOS (latest)
- [x] Solution building for Xamarin Android (latest)
- [x] Solution building for Xamarin Mac (latest)
- [x] Solution building for Xamarin Forms (latest)
- [x] Solution building for Windows Forms
- [x] Solution building for Windows Store (WPA/RT)
- [x] Solution building for Windows Presentation Framework
- [x] Solution building for Windows Universal
- [x] Solution building for Windows Phone 8.x (WPA)
- [x] Solution building for Windows Phone 8.x (Silverlight)
- [x] Publish packages to MyGet feed.
- [x] Update all platform test runner applications to use the package MyGet feed.
- [x] Always ensure latest version of Xamarin iOS is used.
- [x] Always ensure latest version of Xamarin Android is used.
- [x] Always ensure latest version of Xamarin Mac is used.
- [x] Always ensure latest version of Xamarin Forms is used.
- [x] Create monthly job to republish the events on a schedule OR find a way to cause dependency trigger builds (ie.. new XF package = new Events build)
- [x] Start publishing/promoting packages to NuGet.
- [x] Resolve https://github.com/reactiveui/ReactiveUI/issues/1011 by excluding Windows metadata files.

# Background

The package `ReactiveUI-Events` (EventBuilder) does not depend on RxUI at all and RxUI does not depend on it either.  **edit** it's used in the platform specific unit test runner (i.e. iOS.app) **edit**

EventBuilder is essentially a giant xplat Observable from EventHandler helper library at the core, depending on which platform it is executed on it will generate a different library / set of Observable from EventHandlers.

[...]

PR #1056, opened 2016-01-26
eventbuilder.exe is now xplat (tm) and always uses the latest SDK's.

ghuntley:
# Don't merge yet. 

Progresses issue #1043 

Next steps is
* :fire: all `$platform_XS.csproj` conventions
* :fire: all prebuild mdtool/msbuild conventions
* Configure EventBuilder (this) into CI
* Create ReactiveUI-Events packages
* Progress iOS CI story.

### DIFF
--- a/EventBuilder.sln
+++ b/EventBuilder.sln
@@ -1,0 +1,22 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 14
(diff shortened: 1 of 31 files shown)


ghuntley:
Through QA completed on generating events. All platforms generate the same output as the old implementation regardless of Mac|Windows.

PR #1057, opened 2016-01-29
A Rx implementation of the Android service bind.

aalmada:
Service is automatically unbound when connection is disposed.
Sends a null value when the service bind is unexpectedly lost. A new binder is sent when bind restored.
Supports local, Messenger and AIDL binds.
Supports any type of binders returned by the services. Uses generics so it's strongly-typed.


flagbug:
Is there any specific reason you added this to the `ReactiveUI.AndroidSupport` package, instead of the normal `ReactiveUI` package? I'm just wondering, since this code doesn't depend on the Android support library

aalmada:
You're right. No support dependency.
I didn't know where it would fit in the ReactiveUI package.

flagbug:
This folder should be a good place: https://github.com/reactiveui/ReactiveUI/tree/5af96bab3e0d2814b035c1548cf99d5564ba93d1/ReactiveUI/Android

aalmada:
Moved to the suggested place...

flagbug:
After some thinking, I'm wondering if ReactiveUI-Events would be a good place for this @ghuntley

aalmada:
I never used ReactiveUI-Events but, from my understanding, it converts .NET events into IObservables. These PRs I recently submitted don't use events. They wrap the Android/Java pattern of overriding base classes or interfaces.
You may be thinking about extending the purpose of ReactiveUI-Events but then these PRs include extension methods for specific Xamarin for Android classes and not for developer classes. Does this still fit in?

ghuntley:
Correct. Events just autogenerates event handlers into observables.
Extending the autogeneration to also include things like this isn't a bad
idea but no plans at this time (up for grabs). @flagbug If the PR quality
is good then :shipit:

PR #1059, opened 2016-01-29
A Rx implementation of the Android shared preferences changed.

aalmada:
Get the changes in a observable sequence of keys for the changed
preferences.

### DIFF
--- a/ReactiveUI/Android/SharedPreferencesExtensions.cs
+++ b/ReactiveUI/Android/SharedPreferencesExtensions.cs
@@ -3,7 +3,7 @@ using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using Android.Content;
 
-namespace ReactiveUI.Android.Support
+namespace ReactiveUI
 {
     public static class SharedPreferencesExtensions
     {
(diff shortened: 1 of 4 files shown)


ReactiveUI/Android/SharedPreferencesExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+                sharedPreferences.RegisterOnSharedPreferenceChangeListener(listener);
+                return Disposable.Create (() => sharedPreferences.UnregisterOnSharedPreferenceChangeListener (listener));
+            });
+        }
+
+        /// <summary>
+        /// Private implementation of ISharedPreferencesOnSharedPreferenceChangeListener
+        /// </summary>
+        class OnSharedPreferenceChangeListener
+            : Java.Lang.Object
```
flagbug: I'm curious, why does this inherit from  `Java.Lang.Object`?

aalmada:
Removing the Java.Lang.Object inheritance results in the folowing compiler errors:

> Error CS0535: 'ISharedPreferencesExtensions.OnSharedPreferenceChangeListener' does not implement interface member 'IJavaObject.Handle' (CS0535) 
> 
> Error CS0535: 'ISharedPreferencesExtensions.OnSharedPreferenceChangeListener' does not implement interface member 'IDisposable.Dispose()' (CS0535) 
> 

Implementing IDisposable is easy but I'm not sure about IJavaObject. The Xamarin version of Java.Lang.Object takes care of them both.

This is used in a private class so no unwanted methods and properties are exposed to the developers.

PR #1067, opened 2016-02-05
Add Android reactive preferences

aalmada:
PreferenceActivity and PreferenceFragment are used to show a hierarchy of preferences to the user. These reactive overrides allows a ReactiveUI view model to be rendered as an Android preferences View or Fragment.
To create ReactivePreferenceActivity and ReactivePreferenceFragment classes I copied ReactiveActivity and ReactiveFragment source and then replaced the names. I don't see a away to avoid code duplication.


flagbug:
Thanks @aalmada !

aalmada:
@flagbug You're welcome.
It's a small contribution to a great project. Thanks to all the team.

PR #1069, opened 2016-02-06
Add Ensure for more consistent guard code

### LINKED ISSUE #1036: Move away from code contracts, and define internal argument helper
There are a few different places ([example](https://github.com/reactiveui/ReactiveUI/blob/rxui7-master/ReactiveUI/IReactiveObject.cs#L50)) where code contracts are being used, but mostly it's just throwing exceptions ([example](https://github.com/reactiveui/ReactiveUI/blob/rxui7-master/ReactiveUI/CommandBinding.cs#L311)).

I think it would be good to have a discussion to decide which we should use, and put it in writing in the contribution guide.
### UPDATE

So the consensus (from here and Slack discussion) is:
- move away from contracts (change existing usage to inline checks)
- define our own helper to deal with checking arguments
- update the contributions docs so we educate our community

kentcb:
Fixes #1036 by adding an internal `Ensure` class. This class contains static methods to validate arguments and check conditions.
- I'm not super happy with the method names `ConditionSupported` and `ConditionValid` - ideas welcome
- I may not have always chosen the best-fit exception type - any suggested tweaks welcome
- I did not obsessively replace every `throw` with `Ensure` - it wasn't always a good fit and ain't nobody got time for that
- The tests all pass but there's a chance I stuffed up one of the `Ensure` calls and no test exercises it


flagbug:
I'm not very happy about this change, for the sole reason of memory allocation. ReactiveUI is a library used on mobile devices , where memory bandwidth is precious.

While things like 

```
if (binder == null) {
               throw new Exception(String.Format("Couldn't find a Command Binder for {0}", type.FullName));
          }
```

don't allocate any additional memory

```
 Ensure.ConditionValid(
                binder != null,
              "Couldn't find a Command Binder for {0}",
              type.FullName);
```

allocates the string ` "Couldn't find a Command Binder for {0}"`, regardless if the condition is valid or not, potentially resulting in many additional allocations

kentcb:
@flagbug my understanding of this (which is limited in the context of Xamarin's application of IL) is that string interning essentially makes this into a non-issue. This is the reason I added overloads to avoid calls to `String.Format` - all arguments can be passed as constants, thus avoiding allocations.

Are you saying that the allocation will occur every time the `ldstr` occurs? Because that would shock me.

kentcb:
Any more thoughts, @flagbug?

flagbug:
Hm, you're probably right!

kentcb:
Closing this out because I don't think there's much appetite for this right now. TBH, I'd rather redo it later once the shift has been made to netstandard+SDK csproj+cleaner file names etc.

PR #1074, opened 2016-02-24
Don't double-count when navigating back from external trigger.

kentcb:
When the user navigates back externally (e.g. by pressing back in the action bar), make sure we don't double count. That is, make sure we don't _also_ call `PopAsync` because obviously the user will navigate back two pages instead of one and the nav stack will be out of sync with the visible screen.

### DIFF
--- a/ReactiveUI.XamForms/XamForms/RoutedViewHost.cs
+++ b/ReactiveUI.XamForms/XamForms/RoutedViewHost.cs
@@ -41,7 +41,7 @@ namespace ReactiveUI.XamForms
                 var currentCount = previousCount.Skip(1);
 
                 d (Observable.Zip(previousCount, currentCount, (previous, current) => new { Delta = previous - current, Current = current })
-				    .Where(_ => !userInstigated)
+                    .Where(_ => !userInstigated)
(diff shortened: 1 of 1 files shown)


ReactiveUI.XamForms/XamForms/RoutedViewHost.cs
```diff
@@ -40,6 +41,7 @@ public RoutedViewHost()
                 var currentCount = previousCount.Skip(1);
 
                 d (Observable.Zip(previousCount, currentCount, (previous, current) => new { Delta = previous - current, Current = current })
+				    .Where(_ => !userInstigated)
```
shiftkey: Tabs vs spaces? This looks all sad and alone when compared to the next line...

shiftkey:
Aside from the indenting thing, this all seems :thumbsup: from here

ReactiveUI.XamForms/XamForms/RoutedViewHost.cs
kentcb: Damn it! That's what I get for using XS on this occasion :)

Issue #796, opened 2015-01-23
AutoSuspendHelper not using IActivatedEventArgs


rddesmit:
Hi,

The AutoSuspendHelper suspend helper in WinRT is using a concrete implementation of the IActivatedEventArgs, namely LaunchActivatedEventArgs.

    class AutoSuspendHelper
    {
        readonly ReplaySubject<LaunchActivatedEventArgs> _launched = new ReplaySubject<LaunchActivatedEventArgs>(1);
        void OnLaunched(LaunchActivatedEventArgs args)
        {
            _launched.OnNext(args);
        }
    }

This is fine if you call the AutoSuspendHelper from the OnLaunched method in App.xaml.cs. But this won't work if the app is activated in any other way. This is so because the OnActivated method will be called with IActivatedEventArgs that can be of any subtype.

        protected override void OnActivated(IActivatedEventArgs e)
        {
            //Wont work
            new AutoSuspendHelper(this).OnLaunched(e);
        }

Note that if an app is launched by a protocol, and the previous appstate was not running, the app will be launched for the first time. But only the OnActivated method will be called.

So can u let the AutoSuspendHelper use IActivatedEventArgs instead of the concrete implementation LaunchActivatedEventArgs? I could not find any reason why you won't use the interface, concrete fields of the LaunchActivatedEventArgs aren't used.

Thanks.

rddesmit:
I just saw a previous ongoing issue regarding to this one (https://github.com/reactiveui/ReactiveUI/issues/793), so i'll close this one.

PR #1076, opened 2016-02-29
Just use IActivatedEventArgs inside WINRTAutoSuspendApplication inste…

PureWeen:
Can you use IActivatedEventArgs inside WINRTAutoSuspendApplication instead of LaunchActivatedEventArgs so it can be called from OnActivated. In scenarios like where you open app from a protocol OnLaunched won't be called but OnActivated will be...  Changes are pretty unexciting... Just changing the two places where it's LaunchActivatedEventArgs to IActivatedEventArgs ... The only things accessed in the class are defined at the interface level so no special business.

I realize this issue was closed but it seemed more valid then the issue that was referenced in that issue
https://github.com/reactiveui/ReactiveUI/issues/796

### DIFF
--- a/ReactiveUI/Xaml/WinRTAutoSuspendApplication.cs
+++ b/ReactiveUI/Xaml/WinRTAutoSuspendApplication.cs
@@ -15,23 +15,23 @@ namespace ReactiveUI
 {
     public class AutoSuspendHelper : IEnableLogger
     {
-        readonly ReplaySubject<IActivatedEventArgs> _launched = new ReplaySubject<IActivatedEventArgs>(1);
+        readonly ReplaySubject<IActivatedEventArgs> _activated = new ReplaySubject<IActivatedEventArgs>(1);
 
         public AutoSuspendHelper(Application app)
         {
             Reflection.ThrowIfMethodsNotOverloaded("AutoSuspendHelper", app, "OnLaunched");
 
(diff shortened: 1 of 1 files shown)


ghuntley:
Yo @reactiveui/reviewers-uwp could you please review above.

moswald:
So, technically this should be ok, however I have no experience with the auto suspend helper stuff. I've never gotten a good tutorial on what it does, or how it works. I assume this won't break anything, but... 

¯\_(ツ)_/¯

ReactiveUI/Xaml/WinRTAutoSuspendApplication.cs
```diff
@@ -15,7 +15,7 @@ namespace ReactiveUI
 {
     public class AutoSuspendHelper : IEnableLogger
     {
-        readonly ReplaySubject<LaunchActivatedEventArgs> _launched = new ReplaySubject<LaunchActivatedEventArgs>(1);
```
macqm: Should _launched be renamed to _activated?

flagbug:
@PureWeen Can you open a new PR that targets `rxui7-master` please? All new changes go onto this branch

PureWeen:
https://github.com/reactiveui/ReactiveUI/pull/1099

flagbug:
Thanks @PureWeen !

Issue #963, opened 2015-10-04
ReactiveCommand exceptions & View deactivation


flagbug:
Let's take the following piece of code:

``` cs
public class MyViewModel : ReactiveObject
{
    public ReactiveCommand<Unit> TheCommand { get; }

    public MyViewModel()
    {
        TheCommand = ReactiveCommand.CreateAsyncTask(_ => IWillThrowSoon());

        TheCommand.ThrownExceptions
           .SelectMany(ex => UserError.Throw("Message", ex))
           .Subscribe();
    }

    private async Task IWillThrowSoon()
    {
        await Task.Delay(5000);

        throw new Exception("Explosion!");
    }
}
```

In my view I'm registering the `UserError` to catch the exception that the command throws, so I can display the message to the user.

The problem: When my view deactivates, I unregister the `UserError`. If that happens before the command finishes, the ReactiveUI/Rx reroutes the `UserError` exception, because it has no subscribers and the app explodes.

My current solution looks something like this:

``` cs
public class MyViewModel : ReactiveObject, ISupportsActivation
{
    public ReactiveCommand<Unit> TheCommand { get; }

    public ViewModelActivator Activator { get; }

    public MyViewModel()
    {
        this.Activator = new ViewModelActivator();

        var deactivationSignal = new Subject<Unit>();

        this.WhenActivated(d =>
        {
            d(Disposable.Create(() =>
            {
                deactivationSignal.OnNext(Unit.Default);
            }));
        });

        TheCommand = ReactiveCommand.CreateAsyncTask(_ => IWillThrowSoon().ToObservable()
            .TakeUntil(deactivationSignal).DefaultIfEmpty(Unit.Default)));

        TheCommand.ThrownExceptions
           .SelectMany(ex => UserError.Throw("Message", ex))
           .Subscribe();
    }

    private async Task IWillThrowSoon()
    {
        await Task.Delay(5000);

        throw new Exception("Explosion!");
    }
}
```

But this doesn't feel pretty.

/cc @shiftkey With whom I discussed this today

mteper:
Fwiw, in our app we solve this by having the user error subscription in the main view model that hangs around for the lifetime of the app.

kentcb:
Hmmm. It seems to me that RxUI is working as expected here and perhaps you'll need to design around it. You could either take @mteper's suggestion and register the handler at a higher level, or ensure your view cannot be navigated away from until command execution completes.

flagbug:
[...]

PR #1079, opened 2016-03-02
Adds a Deactivated observable to ViewModelActivator, addresses #963.

mteper:
Per @flagbug's suggestion.

### DIFF
--- a/ReactiveUI/Activation.cs
+++ b/ReactiveUI/Activation.cs
@@ -4,10 +4,9 @@ using System.Linq;
 using System.Reactive;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using System.Reflection;
-using System.Text;
 using System.Threading;
-using System.Threading.Tasks;
 using Splat;
 
 namespace ReactiveUI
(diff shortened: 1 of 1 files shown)


ReactiveUI/Activation.cs
```diff
@@ -37,9 +36,12 @@ namespace ReactiveUI
     public sealed class ViewModelActivator
     {
         readonly List<Func<IEnumerable<IDisposable>>> blocks;
+        readonly Subject<Unit> deactivated = new Subject<Unit>();
```
kentcb: For consistency, can we move object creation to the ctor, since that is where `blocks` is instantiated?

ReactiveUI/Activation.cs
```diff
@@ -37,9 +36,12 @@ namespace ReactiveUI
...
         IDisposable activationHandle = Disposable.Empty;
         int refCount = 0;
 
+        public IObservable<Unit> Deactivated { get { return deactivated; } }
```
kentcb: :+1: can we have an `Activated` counterpart too? Seems like we should make it easy to incorporate activation into a reactive pipeline too.

kentcb:
Thanks @mteper!

PR #1084, opened 2016-03-14
Make ToProperty work properly on Indexers

timotei:
In order for Data Bindings to work properly, we need to signal the right property name, in the case of indexers.


kentcb:
Thanks @timotei !

PR #1091, opened 2016-03-25
WhenActivated and CompositeDisposable enhancements

### LINKED ISSUE #1063: Some form of `CompositeDisposable` for `this.WhenActivated` would be a nice addition.
Kent:
> I use `WhenActivated` these days (in iOS and XF): https://github.com/kentcb/WorkoutWotch/blob/master/Src/WorkoutWotch.UI/ExerciseProgramView.xaml.cs#L21
> `AddTo` is still super useful because the code is far more readable. I have a custom `WhenActivated` (https://github.com/kentcb/WorkoutWotch/blob/master/Src/WorkoutWotch.Utility/ReactiveUI/WhenActivatedExtensions.cs) that gives you a `CompositeDisposable` instead of an `Action<IDisposable>`, which means I can do:
```this.WhenActivated(
    disposables =>
    {
        this.Bind(…)
            .AddTo(disposables);
    });
```

Geoff: 
> `AddTo` makes it super explicit that resource will be disposed but it's not really needed right? It's being explicit to help the reader of the code understand what will happen?

Kent: 
> Exactly. I could also do it using the built-in `WhenActivated` like this:
```this.WhenActivated(
    d =>
    {
        d(this.Bind(…));
    });
```
I just find `AddTo` improves readability. And, frankly, makes it easier to write too.

Dennis:
> I've created an extension method called `DisposeWith(CompositeDisposable)` for this, I too think that this makes it easier to read/write

JCM:
> In a similar vain, we wrap the WhenActivated at a lower level and have a ILifetime construct instead and invoke our own virtual methods of Activated through which the ILifetime is passed in.

kentcb:
Fixes #1063 by adding `WhenActivated` overloads that work with `CompositeDisposable`, as well as a `DisposeWith` extension method to add a disposable to a `CompositeDisposable`.

Facilitates code that looks like this:

```cs
this
     .WhenActivated(
         disposables =>
         {
            this
                .Bind(this.ViewModel, x => x.SomeProperty, x => x.someControl.SomeProperty)
                .DisposeWith(disposables);
            this
                .BindCommand(this.ViewModel, x => x.SomeCommand, x => x.someButton)
                .DisposeWith(disposables);
         });
```

### DIFF
--- a/ReactiveUI/AddTo.cs
+++ b/ReactiveUI/AddTo.cs
@@ -1,0 +1,27 @@
+﻿namespace System.Reactive.Disposables
+{
+    public static class CompositeDisposableExtensions
+    {
+        /// <summary>
+        /// Adds a disposable to the specified <see cref="CompositeDisposable"/>.
+        /// </summary>
+        /// <typeparam name="T">
+        /// The type of the disposable.
+        /// </typeparam>
+        /// <param name="this">
+        /// The disposable.
+        /// </param>
+        /// <param name="compositeDisposable">
+        /// The <see cref="CompositeDisposable"/> to which <paramref name="this"/> will be added.
+        /// </param>
+        /// <returns>
+        /// The disposable.
+        /// </returns>
+        public static T AddTo<T>(this T @this, CompositeDisposable compositeDisposable)
+            where T : IDisposable
+        {
+            compositeDisposable.Add(@this);
+            return @this;
+        }
+    }
+}
--- a/ReactiveUI/Activation.cs
+++ b/ReactiveUI/Activation.cs
@@ -138,6 +138,25 @@ namespace ReactiveUI
 
         /// <summary>
         /// WhenActivated allows you to register a Func to be called when a
+        /// ViewModel's View is Activated.
+        /// </summary>
+        /// <param name="This">Object that supports activation.</param>
+        /// <param name="block">The method to be called when the corresponding
+        /// View is activated. The Action parameter (usually called 'disposables') allows
+        /// you to collate all the disposables to be cleaned up during deactivation.</param>
+        public static void WhenActivated(this ISupportsActivation This, Action<CompositeDisposable> block)
+        {
+            This
+                .WhenActivated(
+                    () => {
+                        var d = new CompositeDisposable();
+                        block(d);
+                        return new[] { d };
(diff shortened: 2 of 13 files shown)


flagbug:
Just mentioning this in here, because I'm using this name for my extension method: How about `DisposeWith` instead of `AddTo`? But I've no hard feelings about that name.

I'd name the file that contains the `CompositeDisposableExtensions` class like the class, not the method it contains

ReactiveUI/AddTo.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace System.Reactive.Disposables
+{
+    public static class CompositeDisposableExtensions
```
gluck: current naming policy for extension methods in RxUI is FooMixin(s) for extension methods relating to type Foo. Maybe this one (as well as the file) should be named DisposableMixin(s) ?

ReactiveUI/AddTo.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <param name="this">
+        /// The disposable.
+        /// </param>
+        /// <param name="compositeDisposable">
+        /// The <see cref="CompositeDisposable"/> to which <paramref name="this"/> will be added.
+        /// </param>
+        /// <returns>
+        /// The disposable.
+        /// </returns>
+        public static T AddTo<T>(this T @this, CompositeDisposable compositeDisposable)
```
gluck: what do you need to return the disposable for ? I'd except that once you put the disposable instance into the composite, re-using it for something else is probably not what you'd want ?

gluck: *expect :smile:

flagbug: @gluck I think there's a mistake here, the function returns the original value, not the disposable, it's just the comment that is wrong

gluck: The original value is the disposable (this), just not the composite one, my remark is that the original value is unlikely to be of any use once it's put into the composite, hence not returning anything would make the user write the most sensible thing after calling that one, which is `;` :smile:

michaelstonis:
I am also in favor of DisposeWith for naming

gluck:
LGTM !

PR #1105, opened 2016-05-01
Add ReactiveAppCompatActivity

YvesDelcoigne:
Add ReactiveAppCompatActivity. This class is actually a copy of ReactiveActionBarActivity, but extends from AppCompatActivity, the new standard for Android compat activities.

### DIFF
--- a/ReactiveUI.AndroidSupport/ReactiveAppCompatActivity.cs
+++ b/ReactiveUI.AndroidSupport/ReactiveAppCompatActivity.cs
@@ -1,0 +1,170 @@
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Reactive;
+using System.Reactive.Concurrency;
+using System.Reactive.Disposables;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+using System.Reactive.Threading.Tasks;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Android.App;
+using Android.Content;
+using Android.OS;
+using Android.Runtime;
+using Android.Support.V7.App;
+using Android.Views;
+using Android.Widget;
+using Splat;
+
+namespace ReactiveUI.AndroidSupport
+{
+    /// <summary>
+    /// This is an Activity that is both an Activity and has ReactiveObject powers 
+    /// (i.e. you can call RaiseAndSetIfChanged)
+    /// </summary>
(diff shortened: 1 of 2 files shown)


moswald:
pinging @reactiveui/reviewers-android

kentcb:
Seems to me there are more support base classes missing, such as one for fragments.

ReactiveUI.AndroidSupport/ReactiveAppCompatActivity.cs
```diff
@@ -0,0 +1,170 @@
...
+using Android.App;
+using Android.Content;
+using Android.OS;
+using Android.Runtime;
+using Android.Support.V7.App;
+using Android.Views;
+using Android.Widget;
+using Splat;
+
+namespace ReactiveUI.AndroidSupport
```
kentcb: Hmmm, shouldn't this be `Android.Support`? Don't want to set a bad precedent here.

ReactiveUI.AndroidSupport/ReactiveAppCompatActivity.cs
```diff
@@ -0,0 +1,170 @@
...
+            get { return activityResult.AsObservable(); }
+        }
+
+        protected override void OnActivityResult(int requestCode, Result resultCode, Intent data)
+        {
+            base.OnActivityResult(requestCode, resultCode, data);
+            activityResult.OnNext(Tuple.Create(requestCode, resultCode, data));
+        }
+
+        public Task<Tuple<Result, Intent>> StartActivityForResultAsync(Intent intent, int requestCode)
```
kentcb: I realise you've probably just copied the code from the non-support implementation, but would be interested to know why we're using TPL here (and below) instead of Rx.

ReactiveUI.AndroidSupport/ReactiveAppCompatActivity.cs
YvesDelcoigne: That was something I thought as well, but I didn't want to break with the namespaces of the other classes implemented.

ReactiveUI.AndroidSupport/ReactiveAppCompatActivity.cs
YvesDelcoigne: This is a blind copy from ReactiveActionBarActivity (which is also from the support library, but deprecated), so maybe this is a question for the author of that class? Valid question though.

YvesDelcoigne:
@kentcb Most of the support classes were already here, such as a support class for Fragment, DialogFragment, RecyclerViewAdapter... Only AppCompatActivity was missing, hence this class.

kentcb:
@YvesDelcoigne ah yes, I forgot the Android support stuff is in a separate binary/NuGet.

kentcb:
Had all sorts of trouble trying to build this in XS, but it builds OK in VS.

Thanks @YvesDelcoigne !

PR #1110, opened 2016-05-13
WaitForScheduler Exception handling

MistyKuu:
Replace catching Exception into InvalidOperationException. We want to be explicit because i.e when System.Reactive.Windows.Threading is not attached then FileNotFoundException is thrown and causes using CurrentThreadScheduler instead of DispatcherScheduler. It may cause a lot of silent exceptions being thrown and usage of wrong dispatcher


paulcbetts:
What does this do on:

* Android
* iOS
* OS X

MistyKuu:
There is no impact for other platforms. The only place where WaitForDispatcherScheduler is being used is Registrations.cs and is platform dependent based on directive

paulcbetts:
:+1:

PR #1115, opened 2016-05-27
Add support for ICanActivate.

kentcb:
A change in XF behavior triggered this. Basically, views that are hidden by a dialog are no longer considered active, so their `OnDisappearing` triggers. This was causing problems for me because my main view (which extends `RoutedViewHost`) was disappearing and then reappearing.

Rather than attempt to fix the routing infrastructure (which I don't really want to touch ever again) I decided to make the XF activation-for-view-fetcher support `ICanActivate`. This allows me to circumvent the problem by implementing said interface on my main view such that it _never_ deactivates.


kentcb:
yo @reactiveui/reviewers-xf - anyone feel like eye-balling this and maybe clicking the scary, green button? Just trying not to merge my own shit and all.

Issue #1164, opened 2016-09-22
BindCommand using the Expression withParameter version with a ViewModel causes a exception


glennawatson:
**_Note_: for support questions, please ask on StackOverflow: https://stackoverflow.com/questions/tagged/reactiveui . This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a _feature_ or report a _bug_?**
bug

**What is the current behavior?**
Use the Expression<Func<TViewModel, TParam>> with the parameter "withParam" overload of BindCommand, and don't initially have a view model that is created (eg initially null), then it will cause a exception inside the Reactive WhenAny mixin. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Create a View class and derive off IViewFor<ViewModel>
2. Don't initially set your ViewModel to a value.
3. Use the withParameter Expression<Func<TViewModel, TParam>> to set a parameter for the command
4. Launch your application with the View shown, it will cause a exception in the ReactiveUI WhenAny mixin.

**What is the expected behavior?**
To be able to handle ViewModel's that aren't initially set. Eg potential Master/Child relationships. 

**What is the motivation / use case for changing the behavior?**
Not to have the exception caused. To provide a interface the users expects.

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the latest stable and snapshot (http://docs.reactiveui.net/en/contributing/snapshot/index.html) versions.**
I tested mainly with ReactiveUI 7 with WPF. 

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
See pull request.

glennawatson:
This is still an issue. 

As per the conversation with Geoffrey Huntley on Slack:

you still have the same issue, the issue is if your "ViewModel" doesn't exist yet.

Glenn Watson [1 hour ago] 
public static IReactiveBinding<TView, TViewModel, TProp> BindCommand<TView, TViewModel, TProp, TControl, TParam>(
                this ICommandBinderImplementation This,
                TViewModel viewModel,
                TView view,
                Expression<Func<TViewModel, TProp>> propertyName,
                Expression<Func<TView, TControl>> controlName,
                Expression<Func<TViewModel, TParam>> withParameter,
                string toEvent = null)
            where TViewModel : class
            where TView : class, IViewFor<TViewModel>
            where TProp : ICommand
        {
[...]

PR #1165, opened 2016-09-22
Fixed when using withParameter with the 'Expression<Func<TViewModel, …

glennawatson:
It's a bug fix for: #1164 

#1164

Don't believe so, just uses the existing view model When Any Value overloads used by the vmProperty field. This allows the ViewModel to be null initially.

**Other information**:

…TParam>> withParameter' overload when having initially a null ViewModel property

### DIFF
--- a/src/ReactiveUI/CommandBinding.cs
+++ b/src/ReactiveUI/CommandBinding.cs
@@ -286,7 +286,10 @@ namespace ReactiveUI
             where TView : class, IViewFor<TViewModel>
             where TProp : ICommand
         {
-            return This.BindCommand(viewModel, view, propertyName, controlName, view.ViewModel.WhenAnyValue(withParameter), toEvent);
+            var paramExpression = Reflection.Rewrite(withParameter.Body);
+            var param = Reflection.ViewModelWhenAnyValue(viewModel, view, paramExpression);
+
+            return This.BindCommand(viewModel, view, propertyName, controlName, param, toEvent);
         }
     }
 


ghuntley:
Thanks for the PR Glenn. @reactiveui/reviewers-core can you please review.

kentcb:
@glennawatson can you provide an example of code that previously failed that will now work with this PR applied?

I'm a little unclear from reading the PR description. It makes it sound like it's merely fixing existing behavior, but in fact the PR adds a whole new `BindCommand` overload. My understanding is that you've added the overload to facilitate delayed evaluation of the VM. Is that correct? Regardless, can you please provide some sample before/after code.

glennawatson:
Sorry been away of late. Will get you a example in the next day or so.

kentcb:
@glennawatson ping

glennawatson:
Hi, 

Not going to be able to get back to you guys for at least a week. Been in the hospital for a bit.

The main issue was using the Expression<Func<>> form, it didn't take into account changing values

ex thisValue => thisValue.VM.Blah

If VM wasn't set initially it wouldn't trigger properly.

So I changed it to use the same mechanism you use in the other code to handle that case.

glennawatson:
If you look internally also, I added a new private method but changed the public facing method to use the better approach of using the new method that uses the following: 

var withParamExpression = Reflection.Rewrite(withParameter.Body);
 var withParamFunc = Reflection.ViewModelWhenAnyValue(viewModel, view, withParamExpression);

glennawatson:
Anyway I'm not going to be able to help change the code sample for the next week or so due to technical restrictions, but hope that explanation helps.

kentcb:
Thanks @glennawatson! I wish you a swift recovery.

If I get a chance I'll take a look at getting this merged for the next release.

glennawatson:
I signed the CLA btw, hasn't been updated here though since I did it over a week ago.

glennawatson:
I updated with the latest develop to pass the tests.

glennawatson:
Closing due to CLA bot issues.

PR #1189, opened 2016-11-01
Add WhenAnyObservable overloads that combine latest

kentcb:

Feature.

The `WhenAnyObservable` operator does not work in the same fashion as other `WhenAny*` operators. All observables must be the same type, and the resulting observable sequence is the result of a `Merge` of those observables. Consider this:

```C#
public IObservable<bool> First { ... }

public IObservable<bool> Second { ... }

public IObservable<string> Third { ... }

// elsewhere
this
    .WhenAnyObservable(x => x.First, x => x.Second)
    .Select( /* er, umm, how do I distinguish whether the value came from First or Second? */ )

// somewhere else
this
    .WhenAnyObservable(x => x.First, x => x.Third)  // oops, this is not even possible
```

This PR adds overloads of `WhenAnyObservable` that work in the same fashion as other `WhenAny*` operators. Each observable can be of a different type, and a selector must be provided to combine the results. In our example above, it would look like this:

```C#
this
    .WhenAnyObservable(x => x.First, x => x.Second, (first, second) => /* do whatever here to select an appropriate value */)

this
    .WhenAnyObservable(x => x.First, x => x.Third, (first, third) => /* do whatever here to select an appropriate value */)
```

No.

### DIFF
--- a/src/ReactiveUI.Tests/ReactiveNotifyPropertyChangedMixinTest.cs
+++ b/src/ReactiveUI.Tests/ReactiveNotifyPropertyChangedMixinTest.cs
@@ -11,6 +11,7 @@ using System.Windows;
 using ReactiveUI.Testing;
 using Xunit;
 using Microsoft.Reactive.Testing;
+using System.Reactive;
 
 #if !MONO
 using System.Windows.Controls;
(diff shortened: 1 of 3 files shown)


ghuntley:
LGTM

PR #1193, opened 2016-11-07
Release ReactiveUI V7

ghuntley:
When this pull-request is merged into master it will trigger the CI server to create a draft release containing release notes which were automatically generated from issues assigned to the 7.0.0 milestone. Afterwards the release master should adjust the release notes as desired and click publish release. Publishing a release stamps a 7.0.0 tag and triggers the final build which is then uploaded to NuGet.org / attached to the releasenotes.


ghuntley:
image

Dear @kentcb; I agree that Adelaide does qualify as a real city. Just not as cool as Sydney.

Issue #806, opened 2015-02-11
Issues with ReactiveLists and Xamarin Forms ListView showing new items


PureWeen:
I have a Xamarin Forms project with a ListView bound to a ReactiveList...  After a web call I saturate the list with items... On Android it works great and the items show up but on iOS it doesn't work.... 

I'm pretty positive it has something to do with the thread the items are being added from because if I add an item in the constructor of the view it  shows up but then no subsequent items show up...

I've tried a bunch of scenarios but here's one I thought was interesting:
If I create an ObservableCollection, bind that to the ListView, then subscribe to the CollectionChangedEvent on the ReactiveList, then add items to that ObservableCollection as items are added to the ReactiveList then items show up..... If I change that same ObservableCollection to a ReactiveList then the items no longer show up

Hopefully that makes sense...

It's actually kind of frustrating because I've been trying to create the problem in a smaller project to send you but for some reason in that project everything works super :-(

I've tried as many different permutations of awaits/tasks/ObserveOns that I can but each project doesn't want to budge with how it works/doesn't work on iOS

If that all doesn't trigger anything I can probably take the defunct project and strip things away until I hopefully have something reproduceable... But I was just curious if something in there sounded familiar...

detroitpro:
:+1: I am running into the same problem.

chillbythewater:
The problem is with the ListView not refreshing the layout (at least in my case). The items are added to the ListView but they don't appear until you do something like rotate the device. See: http://forums.xamarin.com/discussion/18631/listview-binding-to-observablecollection-does-not-update-gui
I tested this using a ReactiveList instead of an ObservableCollection and sure enough there are the missing items.

PureWeen:
Alright I think I was able to create an extremely contrived example that seems to demonstrate the behavior (I realize the clearing of the list makes no sense but it's needed to recreate the behavior in this App)

https://github.com/PureWeen/XamarinFormsiOSListViewReactiveList

--code behind adding to the Lists
https://github.com/PureWeen/XamarinFormsiOSListViewReactiveList/blob/master/TheApp/TheApp/Page.xaml.cs

So when I run this project on iOS the ListView bound to the ReactiveList stops displaying items that have been added after the 4th item .... When I run this same project on Android the ListView keeps displaying added items.
```
 //uncomment this to use an ObservableCollection instead and the items will show up in the list
            //without a problem
            //ObservableCollection<Monkey> Monkies = new ObservableCollection<Monkey>();
[...]

PR #1206, opened 2016-11-24
Split ReactiveList into two different implementations

bratsche:

Bug fix

This makes ReactiveList configurable so that it can work with controls that implement their own weak reference event managers, such as Xamarin.Forms.

https://github.com/reactiveui/ReactiveUI/issues/806

Be configurable for platforms that need it.

No

**Other information**:

This is not a working solution as-is. If you notice, in `ReactiveUI.XamForms.Registrations.Register()` is calls `ReactiveListBase.ConfigureStrongReferences()`. I put that there as a hack thinking that it would be called when the applications starts, but it's not.  So for this solution to work we need one of:

1. Find out if there is a place that's called on startup (it looks like there's not though)
2. Make the `ConfigureStrongReferences()` API public, which sucks.
3. Add some kind of `InitReactiveUI()` public API that all applications should call, and have it perform platform-specific initializations like this. This would do nothing for every other API right now, and then on Xamarin.Forms it would call `ConfigureStrongReferences()`.
4. Something else?

### DIFF
--- a/src/ReactiveUI/ReactiveList.cs
+++ b/src/ReactiveUI/ReactiveList.cs
@@ -189,7 +189,7 @@ namespace ReactiveUI
     }
 #endif
 
-    public class ReactiveList
+    public static class ReactiveList
     {
         internal static bool useStrongRefs = false;
 
--- a/src/ReactiveUI/ReactiveList.cs
+++ b/src/ReactiveUI/ReactiveList.cs
@@ -19,92 +19,234 @@ using Splat;
 
 namespace ReactiveUI
 {
+    interface IReactiveListImpl<T>
+    {
+        void CollectionChangingAddHandler(ReactiveList<T> source, NotifyCollectionChangedEventHandler handler);
+        void CollectionChangingRemoveHandler(ReactiveList<T> source, NotifyCollectionChangedEventHandler handler);
+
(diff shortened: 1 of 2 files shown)


oysteinkrog:
Please don't take this to be rude, but have you tried using DynamicData instead?
I (and I know many others) think that ReactiveList should be removed in favor of it.

src/ReactiveUI/ReactiveList.cs
```diff
@@ -19,92 +19,233 @@
...
+        }
+
+        public void RaisePropertyChanged(ReactiveList<T> source, PropertyChangedEventArgs e)
+        {
+            PropertyChangedEventManager.DeliverEvent(source, e);
+        }
+    }
+#endif
+
+    public class ReactiveList
```
shiftkey: Make this `static` to ensure nothing accidental creeps in

escamoteur:
@bratsche Has the problem how to configure now solved?

bratsche:
@escamoteur No, I was hoping maybe to get some feedback on that part of it.

For my own app I did end up switching to DynamicData so I'm not feeling any pressure on this.

bratsche:
It sounds like @PureWeen has sent a PR to Xamarin.Forms that fixes this, so closing.

PR #1212, opened 2016-12-08
Updatable list reactivelistadapter

michaelstonis:
This change adds the ability to swap out the backing list for a ReactiveListAdapter. Sometimes your ViewModel is not ready when you are setting up the adapter or maybe you need to swap out your backing list, this new constructor allows for the ability to replace the list implementation and keep everything working as expected.

Currently, you must assign the backing list into the constructor and you cannot reassign it.

There is a new constructor which allows for the replacement of the backing list in a ReactiveListAdapter.

I don't believe so.

**Other information**:
In terms of docs and tests, I have a sample Android app that I tested it out with, but I am not 100% sure what the best way to test this generically would be...

### DIFF
--- a/src/ReactiveUI/Android/ReactiveListAdapter.cs
+++ b/src/ReactiveUI/Android/ReactiveListAdapter.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Reactive;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Threading;
@@ -44,7 +45,11 @@ namespace ReactiveUI
                 .Subscribe(backingList => {
                     this.list = backingList;
 
-                    _backingListUpdatesObservable.Disposable = this.list.Changed.Subscribe(_ => NotifyDataSetChanged());
+                    _backingListUpdatesObservable.Disposable = 
+                        this.list.Changed
+                            .Select(_ => Unit.Default)
+                            .StartWith(Unit.Default)
+                            .Subscribe(_ => NotifyDataSetChanged());
                 });
         }
 
--- a/src/ReactiveUI/Android/ReactiveListAdapter.cs
+++ b/src/ReactiveUI/Android/ReactiveListAdapter.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Reactive.Disposables;
+using System.Reactive.Linq;
 using System.Threading;
 using Android.Views;
 using Android.Widget;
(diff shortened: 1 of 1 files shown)


ghuntley:
Thanks for the PR @michaelstonis . It's a 👍 from me. Can one of our @reactiveui/reviewers-android reviewers review as well. Looking to get this in `ReactiveUI v7.1` in the next couple of days.

kentcb:
I haven't done a whole lot of Android, so please take that into account here. I'm wondering why this is necessary - instead of having an adapter that changes its backing list based on an observable, could you not just create and assign the adapter based on the observable instead? Certainly that's the approach I took when I did Android development with RxUI. Something like:
```C#
// in view creation logic
this
    .WhenActivated(
        disposables =>
        {
            this
                .WhenAnyValue(x => x.ViewModel.Items)
                .Where(items => items != null)
                .Select(this.CreateAdapterFor)
                .Subscribe(
                    adapter =>
                    {
                        recyclerView.SetAdapter(adapter);
                        adapter.NotifyDataSetChanged();
                    })
                .DisposeWith(disposables);
        });

private ItemsAdapter CreateAdapter(IReadOnlyList<Item> items)
{
    ...
}
```

michaelstonis:
The code above is also how we code up these types of scenarios on Android today. While _it works_, it does require that you do a bit of work and have to setup the adapter again, etc. I feel like there could be a better way though

The intent for this pull is to address two things. Maybe it is bigger than just this pull and I can update it accordingly as well based on feedback, but here is my thought process.

First, the iOS implementation of ReactiveCollectionViewSource<TSource> which is the closest equivalent that we have for Android [allows updating the Data property](https://github.com/reactiveui/ReactiveUI/blob/develop/src/ReactiveUI/Cocoa/ReactiveCollectionViewSource.cs#L156). When this data object gets updated, it will refresh the contents of the container.The change I made was to allow the list to not be read-only. Really, I was just looking for consistency between platforms FWIW. In thinking through it more, if we were to allow the list to be updated and be an externally available property, that would probably solve the same issue or allow for the same kind of updating without a full adapter creation.

The second change that I made was to introduce a constructor that takes an IObservable of backing data. I like this kind of pattern, as it works well when you have late bound data or ViewModels that get replaced. It just becomes an easy way to _set it and forget it_, similar to what we get from WhenAnyValue. It hides some of the implementation away from devs which can help for consistency across the app, but it does not preclude devs from controlling it with the same amount of granularity that they have today.

With all of that being said, what if I broke this into two PRs.
1. Expose the list object as a property in Android ReactiveListAdapter and allow it to be set externally.
2. Modify constructors for both the Android ReactiveListAdapter and mac/iOS ReactiveCollectionViewSource to allow for passing in an observable of backing data.

Thoughts?

Qonstrukt:
I'm also bugged with the backingList being a private field. You can't reach it in your own adapter implementations, while there's often the need for direct access to the list in the state that the adapter is currently using.

Stuff like `GetItemViewType` is impossible to use without exposing your own backing list property or the ViewModel itself. While there's a perfectly good actual representation of the list that's in sync with the state of the adapter.

Besides that I agree for consistency across platforms. Being able to set up the binding to a Data property via the Activity/Fragment/Controller like on iOS would be much appreciated. 👍

PR #1217, opened 2016-12-12
feature: allow WireUpControls to find View members

Qonstrukt:
This will allow `WireUpControls()` to wire up View properties, and not just subclasses of it.

`WireUpControls()` doesn't currently wire up View properties, just subclasses.

`WireUpControls()` now also wires up View properties.

It shouldn't, unless you were expecting `null` values on your View properties. If you were working around this with manual `FindViewById` calls you're assigning the same view twice to the same property, worst-case. (Considering you weren't using duplicate names for views or controls in the first place.)

**Other information**:

### DIFF
--- a/src/ReactiveUI/Android/ControlFetcherMixin.cs
+++ b/src/ReactiveUI/Android/ControlFetcherMixin.cs
@@ -78,7 +78,7 @@ namespace ReactiveUI
         public static void WireUpControls(this ILayoutViewHost This)
         {
             var members = This.GetType().GetRuntimeProperties()
-                .Where(m => m.PropertyType.IsSubclassOf(typeof(View)) || m.PropertyType == typeof(View));
+                .Where(m => typeof(View).IsAssignableFrom(m.PropertyType));
 
             members.ToList().ForEach(m => {
                 try {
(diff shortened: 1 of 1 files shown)


ghuntley:
Thanks for the PR @Qonstrukt. It's a 👍 from me. Can one of our @reactiveui/reviewers-android reviewers review as well. Looking to get this in `ReactiveUI v7.1` in the next couple of days.

src/ReactiveUI/Android/ControlFetcherMixin.cs
```diff
@@ -78,7 +78,7 @@ public static T GetControl<T>(this Fragment This, [CallerMemberName]string prope
         public static void WireUpControls(this ILayoutViewHost This)
         {
             var members = This.GetType().GetRuntimeProperties()
-                .Where(m => m.PropertyType.IsSubclassOf(typeof(View)));
```
kentcb: Can you please change all these to use `IsAssignableFrom`:
```C#
.Where(m => typeof(View).IsAssignableFrom(m.PropertyType)))
```

Qonstrukt: Sure, that should be no problem. I choose for this solution because of a post I saw with some good points on SO: http://stackoverflow.com/a/2742288.
But I don't think the false positives are a problem here because we're not using value-type comparisons.

kentcb:
lgtm. Thanks @Qonstrukt! 👍

PR #1218, opened 2016-12-19
fix: subscription leakage in ReactiveCommand

Hinidu:
Hi! I've described my issue today on the slack channel but nobody replies to me so I found the problem myself :smile: That's my original message:

[SlackMessage]
Hello everyone! I have a problem: the `ViewModel` is leaking if I use any property of the `Model` inside `canExecute` argument of any `ReactiveCommand`. The `Model` lives for the entire life of the application but `ViewModel` should be removed when I close its tab. I use RxUI 7.0. The example:
```c#
var canDeepCopy = this.WhenAnyValue(vm => vm.Model.Id).Select(id => id != Constants.NewEntityId);
this.DeepCopy = ReactiveCommand.Create(this.HandleDeepCopy, canDeepCopy);
```
If I replace `canDeepCopy` value by `Observable.Return(true)` `ViewModel` is not leaking anymore. I even tried to dispose `DeepCopy` command but it didn't give any results. By the way I have another command in this `ViewModel` which doesn't use the `Model` inside its `canExecute` and this command doesn't cause the leakage of the `ViewModel`.
My application is written in WPF.
I can't find any information whether I should dispose reactive commands or not.
https://docs.reactiveui.net/en/design-guidelines/use-this-on-left-of-whenany.html Here I found that there can be the problems if lifetime of the object on which we call `WhenAny` is longer than lifetime of our object. But there is no info about whether it cause the problems if I use such long-life dependency inside the expression argument of `WhenAny`.
[/SlackMessage]

The actual problem was the leakage of subscription to `canExecute`. There are two subscriptions and I don't know why do we need the second one so I just removed it. But the first one which do the actual work is not disposed so I fixed it. I've tested these changes in my project and everything works fine now.

### DIFF
--- a/src/ReactiveUI.Tests/ReactiveCommandTest.cs
+++ b/src/ReactiveUI.Tests/ReactiveCommandTest.cs
@@ -101,6 +101,19 @@ namespace ReactiveUI.Tests
         }
 
         [Fact]
+        public void CanExecuteIsUnsubscribedAfterCommandDisposal()
+        {
+            var canExecuteSubject = new Subject<bool>();
+            var fixture = ReactiveCommand.Create(() => Observable.Return(Unit.Default), canExecuteSubject);
+
+            Assert.True(canExecuteSubject.HasObservers);
+
+            fixture.Dispose();
+
+            Assert.False(canExecuteSubject.HasObservers);
+        }
+
+        [Fact]
         public void CanExecuteTicksFailuresThroughThrownExceptions()
         {
             var canExecuteSubject = new Subject<bool>();
--- a/src/ReactiveUI/ReactiveCommand.cs
+++ b/src/ReactiveUI/ReactiveCommand.cs
@@ -776,11 +776,9 @@ namespace ReactiveUI
 
             this.exceptions = new ScheduledSubject<Exception>(outputScheduler, RxApp.DefaultExceptionHandler);
 
-            this
+            this.canExecuteSubscription = this
                 .canExecute
                 .Subscribe(_ => this.OnCanExecuteChanged());
-
-            this.canExecuteSubscription = this.canExecute.Subscribe();
         }
 
         /// <inheritdoc/>


ghuntley:
Suggested reviewer(s): @kentcb

Hinidu:
By the way in addition to the proposed change to fix my issue I've disposed the command when View was deactivated. Without disposing the ViewModel was leaking as before.

shiftkey:
Yep, I'm not sure there's a good reason for the dual-subscription here.

Tracing back through the history, [initially the disposable was created with some callback logic](https://github.com/reactiveui/ReactiveUI/commit/40d092fc2471f7c7b8402c98adb0a836f95ccf1b#diff-fc7e25a0fa28cd6af8970be934705c3dR80). Then [it was moved out into en empty subscription](https://github.com/reactiveui/ReactiveUI/commit/721262c8b91641869623ee05031015e49465d76d#diff-fc7e25a0fa28cd6af8970be934705c3dL101) as part of introducing more functionality, before it became the current `ReactiveCommand` behaviour in v7.

It really only makes sense to have one subscription here (it seems important). I'll let @kentcb confirm this but i'm :thumbsup: on this change...

Hinidu:
@shiftkey thank you for review! By the way commits referenced by you contains only one subscription. So the second one was added later.

Hinidu:
The second subscription which wasn't disposed was added here https://github.com/reactiveui/ReactiveUI/commit/d3091fe1d9ba0cfd8c6f9b6de3028d4bdb7fdbae.
Perhaps we could add a test that checks that `canExecute` is properly unsubscribed after the `ReactiveCommand` disposal.

Hinidu:
I've committed the test to assure that `canExecute` is unsubscribed after `ReactiveCommand` disposal.

Hinidu:
I've checked that the previous version is failing this test.

Hinidu:
@kentcb ping

kentcb:
@Hinidu am aware, and planning to look at this next week when I'm no longer on holidays.

Hinidu:
@kentcb Thanks! Our holidays in Russia will start tomorrow but I will try to respond if you will have any questions.

kentcb:
Definitely an oops, this one. Thanks for chasing it down @Hinidu and for submitting a great PR 👍

ghuntley:
Thanks for your patience @Hinidu. This has been merged in and is now available from our MyGet alpha channel. We are looking at doing a new release in a couple days time and this will be included within - it will either be `7.0.1` or `7.1.0` depending on how some other reviews progress.

Hinidu:
Thanks @kentcb abd @ghuntley! I'm on holidays until next monday so I can wait :-)

PR #1171, opened 2016-10-02
#1014 Fix DefaultViewLocator

ChrisPulman:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
#1014 bug in DefaultViewLocator

**What is the current behavior? (You can also link to an open issue here)**
Fails to resolve IViewFor<INTERFACE> i.e. A concrete class with an interface registered in Splat
MainViewModel as IMainViewModel register a view as IViewFor<IMainViewModel>
the DefaultViewLocator fails to resolve the view
**What is the new behavior (if this is a feature change)?**
Resolves IVeiwFor<INTERFACE>

**Does this PR introduce a breaking change?**
No, unless desired operation is to fail when using interfaces as IViewFor registration

**Please check if the PR fulfills these requirements**
- [ ] The commit follows our guidelines: https://github.com/reactiveui/reactiveui#contribute
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Other information**:

PR #1230, opened 2017-01-17
fix: Resolve views matching IViewFor<IViewModel>

### LINKED ISSUE #1014: DefaultViewLocator doesn't resolve IViewFor<IFooBarViewModel>

olevett:
include type registrations matching IViewFor<IViewModel> in view location

Fixes #1014, #1171

Bug fix

Currently, IViewFor<IViewModel> type registrations are not resolved, even though the comments suggest they should be

Resolving IViewFor<IViewModel> style registrations.

No

**Other information**:
This has come up a couple of times, including #1015, where a fix very like this was reviewed and merged, but I think it coincided with the RXUI7 changes, so got missed in a merge or something similar. The original issue was #1014  and there's a related pull request under #1171.

I've not added additional documentation as I believe the relevant docs (e.g. the comments) are 'correct' in terms of the functionality after this change.


ChrisPulman:
I think that there is an issue with this pull request due to the test not using an interface to resolve by it is instead using the concrete Class.
Please see my PR #1014 which includes the following (NOTE the use of IBazViewModel ):
```
public class RxRoutingTests : IEnableLogger 
     { 
        public void ResolveExplicitViewType()
             using (resolver.WithResolver()) { 
                 var fixture = new DefaultViewLocator(); 
                 IBazViewModel vm = new BazViewModel(null); 
                 var result = fixture.ResolveView(vm);  
                 this.Log().Info(result.GetType().FullName);  
                 Assert.True(result is BazView);  
             }  
         }  
     }

olevett:
I've possibly misunderstood here, but the original issue I raised was that if you register an `IViewFor<IViewModel>` then it can't be resolved for a concrete view instance. In the case this is used, we have an instance of a concrete ViewModel, and are trying to find the corresponding view (so, need to find what interfaces the ViewModel implements and then resolve a view for that). 

That's what the test I've written is trying to cover anyway...

ChrisPulman:
@olevett @ghuntley 
To fix this issue the ResolveView function needs to resolve the view by passing an INTERFACE not from a concrete instance of the interface. This is to allow us to use IViewFor<INTERFACE> in our view, navigate using the INTERFACE and then the IOC container does the rest, in this case Splat. 
This then allows us to swap out the actual concrete instance at will i.e. During test, during debug, during release, or just different use case scenario's.
I hope that this helps, perhaps you can combine the code I added in PR #1171 in particular the attemptToResolveView, and the two interfaceifyTypeName methods, this may result in a better PR especially if your addition resolves another issue.

olevett:
@ChrisPulman 
Right, I see now, I think we have two subtly different issues, e.g. I've been registering views for the interface, and then letting ReactiveUI do it's resolve around the concrete class it has, rather than forcing it to use the VM interface  (IViewModel lets say). Whereas it sounds like you're trying to resolve from the interface not the concrete class? I think we have two subtly different issues, I can bring in your fix and we can add a couple more tests to cover it as well, if that sounds good?

ChrisPulman:
@olevett thanks, out of interest does your implementation also resolve `ResolveView<T>("contract")` , this is also part of the view resolution, although I am not certain of the use case scenario.
Perhaps we should also add some tests for this to complete the testing of this function.
Perhaps @ghuntley can assist us with the use case scenario for `ResolveView<T>("contract")` as IViewFor doesn't take any arguments for the contract as far as I am aware, but I can see the benefit if you can get a view and attach it to a particular instance of a ViewModel

olevett:
From what I remember, and I'll do some digging to check, but I think you sometimes get things like "Landscape" and "Portrait" coming in in a couple of places (and I think it can be set in the RoutedViewHost with ViewContractObservable on platforms that support it).

I guess in terms of testing, it should end up hitting the DependencyResolver, and not really anywhere else. I'll have a think about what testing there makes sense.

kentcb:
Superseded by #1282.

PR #1240, opened 2017-01-25
Remove superfluous AsObservable calls from ReactiveCommand

### LINKED ISSUE #1236: perf: remove superfluous AsObservable calls in ReactiveCommand

Code issue.

`ReactiveCommand` includes several properties that return observables, but calls `AsObservable` on those return values.

There is no need to call `AsObservable` - it only adds to the performance cost by creating another object on top of the already-usable observable.

Better code and performance.

7

kentcb:

Correctness/performance improvement. Fixes #1236.

`ReactiveCommand` includes a couple of unnecessary calls to `AsObservable`.

Remove the unnecessary calls.

No.


kentcb:
@reactiveui/reviewers-core any of you fine people want to 👀 this?

mteper:
LGTM. I imagine this was probably discussed in Slack, but if I were to be picky, I would say the one thing this bug description is missing is a bit of history -- how did those calls come to be in the first place?

kentcb:
@mteper it's hard to say for sure due to the mangled history, but I believe this was just a vestige of the code's evolution. That is, at one point I had written it to use subjects to back `CanExecute` and `IsExecuting`. Consequently, I did the right thing and called `AsObservable` on those subjects. When I improved the code to remove the subjects, I neglected to remove the `AsObservable` calls.

How do you feel about merging this, following the process set out by @ghuntley?

mteper:
I've limited time and I'll need to set up a dev environment. If this can wait a week or two, I'll gladly figure out where the process is documented and follow it. If this sounds like I'll be the bottleneck, then I'll bow out.

Issue #1234, opened 2017-01-19
WireUpControls throws an exception


orzech85:
WireUpControls throws an exception:
`System.MissingFieldException: Failed to wire up the Property CurrentFocus to a View in your layout with a corresponding identifier`

But I have no view called `CurrentFocus`.

I think this commit caused it:
https://github.com/reactiveui/ReactiveUI/commit/276f96302a0ed9e444884dfa40f1cd5691ffd725

Example project with issue demo:
https://github.com/orzech85/ReactiveExample/blob/master/ReactiveUI.Droid/MainActivity.cs#L20

Qonstrukt:
See #1231.

Issue #1231, opened 2017-01-17
MissingFieldException in fragment and activity


binoypatel:
Hi

After upgrading to rx 7.1, I am getting following exception when WireUpControls called:

1. In fragment:
System.MissingFieldException: Failed to wire up the Property View to a View in your layout with a corresponding identifier

2. In activity:
System.MissingFieldException: Failed to wire up the Property CurrentFocus to a View in your layout with a corresponding identifier

As you can see that "View" and "CurrentFocus" both are type of Android.Views.View properties and I believe it should not be in the collection in ControlFetcherMixin.WireUpControls. You can reproduce this issue in Playground-Android. Let me know if I am doing something wrong? I have attached MainActivity.cs from Playground-Android for your reference:
[MainActivity.txt](https://github.com/reactiveui/ReactiveUI/files/712522/MainActivity.txt)

Thanks,
Binoy

kentcb:
Seems likely that this was brought about by #1217. Perhaps @Qonstrukt has some thoughts on this?

Qonstrukt:
I'll try and reproduce this today.

Qonstrukt:
So, this bug got introduced by #1217 but I'd like some ideas on how we might fix this, because I see a couple of options.
We could just revert #1217 of course, but this issue also happens with descendants when you use properties for dynamically created Views for example. 
I'd say we need a way to have this fail silently, or just throw a warning like we do with POCO properties.

@reactiveui/reviewers-android Do you guys think it's necessary to add a parameter to WireUpControls to triggers warnings instead of exceptions, or could we make this the default behaviour? Or do you think we shouldn't touch this at all and just revert #1217?

binoypatel:
I could not quite understand the changes made in #1217 but here is what I suggest:
```c#
public static void WireUpControls(this Activity This, bool viewWithsubClass = false)
{
     IEnumerable<PropertyInfo> members;
     if (viewWithsubClass)
     {
         members = This.GetType().GetRuntimeProperties()
         .Where(m => typeof(View).IsAssignableFrom(m.PropertyType));
      }
     else
     {
         members = This.GetType().GetRuntimeProperties()
         .Where(m => m.PropertyType.IsSubclassOf(typeof(View)));
     }...
```
reset of the method is same. We can do the same for all methods. 

@Qonstrukt Can you please share some code where you need IsAssignableFrom(), that will help me understand the benefit of it?

Kind regards,
Binoy

[...]

Issue #1227, opened 2017-01-12
ReactiveViewHost KeyNotFoundException after update to version 7.1


bidy:
I'm doing this in the Xamarin.Android Activity:
`var adapter = new ReactiveListAdapter<TreeItemViewModel>(ViewModel.TreeItemViewModels, (viewModel, parent) => new ImageItemView(viewModel, this, parent));`

the ImageItemView is like this:

`public class ImageItemView : ReactiveViewHost<TreeItemViewModel>
    {
        public ImageItemView(TreeItemViewModel viewModel, Context ctx, ViewGroup parent)
            : base(ctx, Resource.Layout.drawer_list_item, parent)
        {
            ViewModel = viewModel;

            this.OneWayBind(ViewModel, vm => vm.Title, v => v.Text1.Text);
            this.WhenAnyValue(vm => vm.ViewModel.Icon)
                .Subscribe(x => ImgIcon.SetImageDrawable(x.ToNative()));
        }

        public ImageView ImgIcon { get; private set; }
        public TextView Text1 { get; private set; }
    }`

This worked well in ReactiveUI Version 6. Now with version 7.1 I get this exception:
System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.

The exception occurs when calling the base constructor of ImageItemView.

The views / Layouts / xaml haven't been changed and worked with the previous version.

ghuntley:
Can you please provide the entire stacktrace?

bidy:
Call stack:
 	0xFFFFFFFFFFFFFFFF in System.Diagnostics.Debugger.Mono_UnhandledException_internal	C#
 	0x1 in System.Diagnostics.Debugger.Mono_UnhandledException at /Users/builder/data/lanes/3511/501e63ce/source/mono/mcs/class/corlib/System.Diagnostics/Debugger.cs:122,-1	C#
 	0x33 in object.b972e2c9-785a-4893-8f94-61ba0a6f3aa9	C#
 	0x48 in ReactiveUI.ControlFetcherMixin.	C#
 	0x42 in System.Collections.Generic.List<System.Reflection.PropertyInfo>.ForEach at /Users/builder/data/lanes/3511/501e63ce/source/mono/mcs/class/referencesource/mscorlib/system/collections/generic/list.cs:564,-1	C#
 	0x54 in ReactiveUI.ControlFetcherMixin.WireUpControls	C#
[...]

PR #1250, opened 2017-01-29
feat: more flexible WireUpControls implementation 

Qonstrukt:
This is an improvement upon `WireUpControls` and #1217, but this time without introducing the problems from #1234, #1231 and #1227.

`WireUpControls` doesn't allow you to specify how it'll find members currently. It'll always find all properties which are subclasses of View, which isn't always what you want. 

For example, when you have a `Toolbar` property in your Fragment which you initialise dynamically in `OnViewCreated`, you cannot use `WireUpControls` anymore because it's unable to wire up the property with a view in your layout. This counts for the fragment itself or any subclasses thereof.

This PR allows you to switch from just implicit resolving, to two additional ways to explicitly wire up properties to views in your layout, by using opt-in or opt-out.

When using **implicit**, it'll default to the way `WireUpControls` has always worked.

When using **explicit opt-in**, you should use the `WireUpResource` attribute on your properties to explicitly define you want it wired up.

When using **explicit opt-out**, it introduces the behaviour of #1217 with the added bonus of opting out certain properties with the `IgnoreResource` attribute. 
This is currently not very useful in activities and fragments, because it'll also introduce the problems with #1217, but it can certainly be handy in views for example.

In contrary to #1217 it shouldn't. Unfortunately there doesn't seem to be any testing set-up for Android currently (amongst other platforms), but I will be using this in production in several ways in the upcoming week.

**Other information**:
- I'd like feedback from @reactiveui/reviewers-android, @binoypatel and @makon on this. Am I missing something? Is the naming alright? The obvious.
- I'll be testing this thoroughly in the next couple of days.
- I'll whip up a page in the docs after I've had the green light on this.


kentcb:
@orzech85 @binoypatel @bidy are you guys able to pull this down and verify it resolves your issue?

binoypatel:
@kentcb Sure I will review this and get back to you by this week

makon:
At first sorry, for the late response. Unfortunately I am quite busy this week...

Thank you for pointing out the issue with the Toolbar. I haven't thought about that but you are totally right. The current solution does not fit this cases properly. It seems like we identified two separate issues with the current implementation

- The need to wire up controls of type `View`
- The need to exclude properties which are programmatically created

Please correct me if I am missing some other point.

I really like the idea of having the possibility to define which properties should be included or excluded. It is definitely an advantage compared to the current behavior. I was thinking about a similar solution but without using attributes for it. Something like the following pseudo code:

```
// excluding properties adresing issue #2
this.WireUpControls(options => options.Exclude(x => x.Toolbar)
                                      .Exclude(x => x.CurrentFocus));

// include properties adresing issue #2
this.WireUpControlsExplicit(options => options.Include(x => x.Toolbar));

// determining the resolve strategy adresing issue #1
this.WireUpControls(Action<ExcludeOptions> options, ResolveStrategy strategy = ResolveStrategy.IsSubclassOfView);

this.WireUpControlsExplicit(Action<IncludeOptions> options, ResolveStrategy strategy = ResolveStrategy.IsSubclassOfView);

public enum ResolveStrategy
{
    IsSubclassOfView,
    IsAssignableFromView
}
```

This solution would have the advantage to be able to include or exclude properties from base types which are defined in third part libraries where I cannot add any attribute (e. g. Activity, Fragment, etc.). It would also provide the flexibility to exclude a property just in a given scenario. I am not an RxUI expert but I have a slight feeling that it would better fit to the remaining API because I would not expect the need to use Attributes with RxUI. But this is a highly personal impression.

What do you think about this? Do you think it is worth the overhead on work? It would be great if you could give me some feedback. If there is some interest I could give it a shot. But as I already mentioned your solution is totally an improvement and I am also fine with it.

Qonstrukt:
Thank you for your insights!

The idea you've come up with is actually quite compelling, I agree it fits better in a Rx context. However I don't think I would apply it much to my use cases. Having to exclude certain Views every time (like `CurrentFocus`) gets cumbersome quick, so I would quickly fall back to the `Include` strategy. And then your solution would require me to define a query inside my code blocks again which feels nearly the same as using `FindViewById`. 

Having the possibility to use attributes keeps such housekeeping rules from my code which I prefer personally, but I'd really like to hear other people's opinions about this.

I'm going to rename `ResolveMembers` to `ResolveStrategy` btw, as that's a much better name. 😄

Qonstrukt:
We're a week down the road and I've pushed two improvements;
- It's now possible to define a custom resource name so your property names don't necessarily need to correspond with the resource names.
- I've added the possibility to use a `WireUpResource` attribute with the Implicit resolve strategy, so you can add `View` resolving on a per-property basis.

Besides that, I've renamed `ResolveMembers` to `ResolveStrategy` as I think it's a much better name, credit to @makon.

Is there anyone in @reactiveui/reviewers-android group that wants to share their opinion?

Qonstrukt:
Another week, and I've used my latest implementation in two big projects without problems. I'll write up some draft docs to explain how it works exactly.

kentcb:
@qonstrukt are you happy/keen for this to be merged and included in the next release?

Qonstrukt:
Personally I am, this time I've also made it with 100% backward compatibility in mind. So it should be 'safe".

kentcb:
Excellent! Thanks @Qonstrukt!

PR #1261, opened 2017-02-03
feature: allow interaction handlers of any observable type

### LINKED ISSUE #1260: Make Rx-based interactions agnostic about observable type

Design improvement.

Interaction handlers registered via `RegisterHandler(Func<InteractionContext<TInput, TOutput>, IObservable<Unit>> handler)` are required to return `IObservable<Unit>`. It's easy to miss this and return an observable of a different type when calling `RegisterHandler`:

```C#
someInteraction.RegisterHandler(context => someObservableThatIsNotUnit);
```
This will still compile, but resolve to the wrong `RegisterHandler` overload (the one that takes `Action<InteractionContext<TInput, TOutput>>`.

It would be better if `RegisterHandler` took `IObservable<TDontCare>`. The only slight issue with doing so is that every handler received will have to be converted to a signal before it can be stored in the list of handlers. This means that if the observable is already of type `Unit` you will incur a very slight performance hit that you wouldn't otherwise. But I'm only mentioning it for completeness - clearly correctness and falling into a pit of success is far more important.

Just allow any kind of observable to resolve to the correct `RegisterHandler` overload, much the same as any kind of `Task` resolves to the corrent `RegisterHandler` overload.

Pit of success, framework design.

7

kentcb:
Fixes #1260.

### DIFF
--- a/src/ReactiveUI/Interactions.cs
+++ b/src/ReactiveUI/Interactions.cs
@@ -200,14 +200,16 @@ namespace ReactiveUI
         /// <returns>
         /// A disposable which, when disposed, will unregister the handler.
         /// </returns>
-        public IDisposable RegisterHandler(Func<InteractionContext<TInput, TOutput>, IObservable<Unit>> handler)
+        public IDisposable RegisterHandler<TDontCare>(Func<InteractionContext<TInput, TOutput>, IObservable<TDontCare>> handler)
         {
             if (handler == null) {
                 throw new ArgumentNullException("handler");
             }
 
-            this.AddHandler(handler);
-            return Disposable.Create(() => this.RemoveHandler(handler));
+            Func<InteractionContext<TInput, TOutput>, IObservable<Unit>> unitHandler = context => handler(context).Select(_ => Unit.Default);
+
+            this.AddHandler(unitHandler);
+            return Disposable.Create(() => this.RemoveHandler(unitHandler));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/ReactiveUI/Interactions.cs
```diff
@@ -200,14 +200,16 @@ public IDisposable RegisterHandler(Func<InteractionContext<TInput, TOutput>, Tas
         /// <returns>
         /// A disposable which, when disposed, will unregister the handler.
         /// </returns>
-        public IDisposable RegisterHandler(Func<InteractionContext<TInput, TOutput>, IObservable<Unit>> handler)
+        public IDisposable RegisterHandler<TDontCare>(Func<InteractionContext<TInput, TOutput>, IObservable<TDontCare>> handler)
```
ghuntley: Should we keep `public IDisposable RegisterHandler(Func<InteractionContext<TInput, TOutput>, IObservable<Unit>> handler)` as an alias method to the new signature but annotate it with `[Obsolete("Please use RegisterHandler<T> instead?")]` as a migration path? I want to stop making hard public API contract changes. Your thoughts?

kentcb:
@ghuntley it's not a breaking change in the sense that if people were accidentally calling the non-observable version because their observable was not `Unit`, this change will actually _help_ them.

kentcb:
@ghuntley not sure which docs you mean? Nothing has changed from a consumer point of view apart from the API being more forgiving of the observable type. But the docs don't talk about that any way.

ghuntley:
Apologies, an assumption on my part that documentation would need updating. If they don't, they don't. I've started thinking about making this a standard question as merging pull requests - does the doc need updating and if so please do it. I'm keen to merge the documentation repo back into the main RxUI repo at some stage this year so documentation gets versioned along side the code.

PR #1266, opened 2017-02-07
Comment tidying

olevett:
Updating some comments

Some comments either are missing/don't make sense.

I've tidied up a few comments...

No

**Other information**:
There are many more places where the comments don't line up...

### DIFF
--- a/src/ReactiveUI/CommandBinding.cs
+++ b/src/ReactiveUI/CommandBinding.cs
@@ -9,6 +9,9 @@ using Splat;
 
 namespace ReactiveUI
 {
+    /// <summary>
+    /// Various helpers to bind View controls and ViewModel commands together
+    /// </summary>
     public static class CommandBinder
     {
         static ICommandBinderImplementation binderImplementation;
@@ -30,7 +33,7 @@ namespace ReactiveUI
         /// <param name="view">The View</param>
         /// <param name="viewModel">The View model</param>
         /// <param name="controlName">The name of the control on the view</param>
-        /// <param name="propertyName">The ViewModel command to Bind.</param>
+        /// <param name="propertyName">The ViewModel command to bind</param>
         /// <param name="withParameter">The ViewModel property to pass as the
         /// param of the ICommand</param>
         /// <param name="toEvent">If specified, bind to the specific event
@@ -57,6 +60,7 @@ namespace ReactiveUI
         /// the binding</returns>
         /// <param name="view">The View</param>
         /// <param name="viewModel">The View model</param>
+        /// <param name="propertyName">The ViewModel command to bind</param>
         /// <param name="controlName">The name of the control on the view</param>
         /// <param name="toEvent">If specified, bind to the specific event
         /// instead of the default.</param>
@@ -81,6 +85,7 @@ namespace ReactiveUI
         /// the binding</returns>
         /// <param name="view">The View</param>
         /// <param name="viewModel">The View model</param>
+        /// <param name="propertyName">The ViewModel command to bind</param>
         /// <param name="controlName">The name of the control on the view</param>
         /// <param name="withParameter">The ViewModel property to pass as the
         /// param of the ICommand</param>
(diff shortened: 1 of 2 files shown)


src/ReactiveUI/CommandBinding.cs
```diff
@@ -30,7 +33,7 @@ static CommandBinder()
         /// <param name="view">The View</param>
         /// <param name="viewModel">The View model</param>
         /// <param name="controlName">The name of the control on the view</param>
```
olevett: There's inconsistency on what "controlName" is called, e.g. here we have {{controlName}}  but on line 189 we have a {{controlProperty}} - I think it would be worth being consistent...

src/ReactiveUI/CommandBinding.cs
```diff
@@ -9,6 +9,9 @@
 
 namespace ReactiveUI
 {
+    /// <summary>
+    /// Various helpers to bind View controls and ViewModel commands together
+    /// </summary>
     public static class CommandBinder
     {
         static ICommandBinderImplementation binderImplementation;
```
olevett: There were more major things I felt like doing, but stopped myself...

This interface is internal, and only used by the concrete implementation, I'm not sure what benefit this split into a fair number of chunky classes that only call each other and all coexist in the one file is. I'd be tempted to combine into the one static class at this level and that should hopefully reduce the clutter/make it easier to follow.

src/ReactiveUI/ReactiveBinding.cs
```diff
@@ -76,17 +72,14 @@ internal class ReactiveBinding<TView, TViewModel, TValue> : IReactiveBinding<TVi
     {
         private IDisposable bindingDisposable;
 
-        /// <summary>
```
olevett: This comment doesn't really add anything, and we're in an internal class so not sure it really needs it - the same could be set for the property comments below, but at least they're not actually wrong...

olevett:
@ghuntley not entirely sure what's going on here, AppVeyor has failed on OpenCover but I can't see anything obviously wrong...

ghuntley:
image

https://ci.appveyor.com/project/ghuntley/reactiveui/build/505.Branch.pull/1266/merge.Sha.4a1025b956ebfb171a559c7de8e3c475412ff663#L2668

One of the tests is failing.

src/ReactiveUI/CommandBinding.cs
ghuntley: >  I think it would be worth being consistent...
Agreed and support this. Grep the build log for ` warning CS1573` for hints of inconsistency that also needs cleaning up

image

olevett:
@ghuntley  managed to miss that test failure completely, apologies!

kentcb:
This PR seems to include a lot more than the title suggests, and I have to say it makes me feel a little uneasy. Not saying it won't get accepted, but I think there's considerably less chance than when a PR is more focused in what it's doing. I don't suppose you'd consider separate PRs?

olevett:
@kentcb apologies, was trying to reduce the number of PRs for stuff that seemed fairly trivial...will happily split into two though.

src/ReactiveUI/CommandBinding.cs
olevett: In a separate commit/PR I'll have a look at lining up controlProperty/controlName.

As a general stylistic point, are we aiming for extension methods to be on a parameter called `This`, `@this` or on something like these ones like `view`?

ghuntley: re: seperate commit for controlProperty/controlName 👍 

re: extension method styling for xmldoc

@kentcb and @paulcbetts any thoughts/desires on this one?

ghuntley:
@olevett PR's as thin highly targeted slices that get merged/integrated often; even if it takes a tiny bit more time. Progress should be incremental, in general long-running layers are discouraged 👍  Don't apologise about the broken test failure; obviously it'll need fixing up but I'm actually really happy to see. It's positive feedback that the efforts in improving integration testing is starting to pay off.

PR #1268, opened 2017-02-08
style: tidy up and expand comments in Activation

olevett:
More comment tidying, in Activation this time

**Other information**:

### DIFF
--- a/src/ReactiveUI/Activation.cs
+++ b/src/ReactiveUI/Activation.cs
@@ -54,6 +54,9 @@ namespace ReactiveUI
         /// <value>The deactivated.</value>
         public IObservable<Unit> Deactivated { get { return deactivated; } }
 
+        /// <summary>
+        /// Constructs a new ViewModelActivator
+        /// </summary>
         public ViewModelActivator()
         {
             blocks = new List<Func<IEnumerable<IDisposable>>>();
@@ -87,8 +90,10 @@ namespace ReactiveUI
         /// This method is called by the framework when the corresponding View
         /// is deactivated.
         /// </summary>
-        /// <param name="ignoreRefCount">Force the VM to be deactivated, even
-        /// if more than one person called Activate.</param>
+        /// <param name="ignoreRefCount">
+        /// Force the VM to be deactivated, even
+        /// if more than one person called Activate.
+        /// </param>
         public void Deactivate(bool ignoreRefCount = false)
         {
             if (Interlocked.Decrement(ref refCount) == 0 || ignoreRefCount) {
@@ -98,6 +103,9 @@ namespace ReactiveUI
         }
     }
 
+    /// <summary>
+    /// A set of extension methods to help wire up View and ViewModel activation
+    /// </summary>
     public static class ViewForMixins
     {
         static ViewForMixins()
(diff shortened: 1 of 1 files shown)


src/ReactiveUI/Activation.cs
```diff
@@ -54,6 +54,9 @@ public sealed class ViewModelActivator
         /// <value>The deactivated.</value>
         public IObservable<Unit> Deactivated { get { return deactivated; } }
 
+        /// <summary>
```
olevett: Comments like this always feel...pointless, happy to remove it if wanted but thought since I'm here I may as well fix the compile time warning...

src/ReactiveUI/Activation.cs
```diff
@@ -315,16 +343,29 @@ static IDisposable handleViewModelActivation(IViewFor view, IObservable<bool> ac
 
     /// <summary>
     /// This class implements View Activation for classes that explicitly describe
-    /// their activation via ICanActivate. This class is used by the framework.
+    /// their activation via <see cref="ICanActivate"/>. This class is used by the framework.
     /// </summary>
     public class CanActivateViewFetcher : IActivationForViewFetcher
```
olevett: This feels like it should be in internal class, it's whole purpose in life is to be something used by the framework rather than end consumers, but that's definitely a breaking change/needs some discussion...

src/ReactiveUI/Activation.cs
ghuntley: Voted pointless, the return type is `Unit`

image

See above for another line that can be 🔥

src/ReactiveUI/Activation.cs
ghuntley: Hmm, I don't know.  

My perspective on breaking stuff has changed since starting work on video courseware; there has to be an incredibly good reason to do it as it wastes the time of consumers and my time if the courseware needs to be updated. @kentcb probably has similar feelings as he's going through this process right now with authoring a book.

I had a quick look at the GitHub for Desktop aka Visual Studio plugin source code and it has taken no references to this class. It's one of the larger open-source projects that uses RxUI.

https://github.com/github/VisualStudio/search?utf8=%E2%9C%93&q=CanActivateViewFetcher

src/ReactiveUI/Activation.cs
```diff
@@ -87,8 +90,10 @@ public IDisposable Activate()
         /// This method is called by the framework when the corresponding View
         /// is deactivated.
         /// </summary>
-        /// <param name="ignoreRefCount">Force the VM to be deactivated, even
-        /// if more than one person called Activate.</param>
+        /// <param name="ignoreRefCount">
+        /// Force the VM to be deactivated, even
+        /// if more than one person called Activate.
+        /// </param>
```
ghuntley: This could be merged onto a single line.

src/ReactiveUI/Activation.cs
olevett: In terms of compiler warnings (can't remember exactly what you get for missing a returns...), would you prefer empty comments or nothing at all?

src/ReactiveUI/Activation.cs
olevett: I'm wary of breaking things, but the pre-existing comment is along the lines of "used by the framework" and it's explicitly registered by ReactiveUI, so think its both fairly low risk and fairly low impact....

kentcb:
Thanks @olevett!

Issue #1256, opened 2017-02-01
XF: auto-assign ViewModel from BindingContext


kentcb:
**Do you want to request a *feature* or report a *bug*?**

Feature/enhancement.

**What is the current behavior?**

The `ViewModel` property for reactive views is not auto-populated from the `BindingContext`. This means that developers are required to manually propagate the property. For example:

```XML
        <ListView.ItemTemplate>
            <DataTemplate>
                <local:MyReactiveView ViewModel="{Binding .}"/>
            </DataTemplate>
        </ListView.ItemTemplate>
```

This is annoying and is an easy thing to miss.

**What is the expected behavior?**

It would be nice if the base reactive view classes could override `OnBindingContextChanged` and auto-assign to the `ViewModel` property where relevant:

```
protected override OnBindingContextChanged()
{
    this.ViewModel = this.BindingContext as TViewModel;
}
```

However, I think this code will need to be slightly smarter. If the `BindingContext` is not a `TViewModel` (or is `null`) then it shouldn't overwrite whatever is assigned to `ViewModel`, just in case the two differ.

**What is the motivation / use case for changing the behavior?**

Reduce friction.

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the latest stable and snapshot (http://docs.reactiveui.net/en/contributing/snapshot/index.html) versions.**

7

FrancoisM:
I'm on my phone trying to find some kind of "like" button to support that request but cannot find any so I say it here: +1.

PureWeen:
Along these lines would it also be helpful to assign the BindingContext in cases like VM Navigation?

kentcb:
@PureWeen I think that's a good suggestion. Also true of `ViewModelViewHost`. It's not a pain point for me since I almost never use XAML bindings, but I can see how it would be for those who do.

FrancoisM:
True. So far I use Xamarin Forms MVVM with Prism. I never care about BindingContext. Having to deal with it with ReactiveUI is like a regression.
And Binding with Xaml would be nice to have.
ReactiveUI should override Xamarin Forms bindings somehow.

escamoteur:
xamvvm does assign the Bindingcontext of Pages to the ViewModel property

PR #1270, opened 2017-02-13
feat: synchronize ViewModel and BindingContext properties

kentcb:

Feature.

#1256

1. Whenever the `BindingContext` changes on a reactive view, the `ViewModel` is updated accordingly. However, if the `BindingContext` is set to a non-`null` instance of a different type of object, the VM is not updated (allowing for `ViewModel` and `BindingContext` to deviate where desirable).
2. Whenever the `ViewModel` property is assigned by `RoutedViewHost` or `ViewModelViewHost`, the `BindingContext` is also set to the same.

No.


Qonstrukt:
Any reasons you didn't do this the other way around?
IMHO, I'd expect the `BindingContext` to be set whenever I assign an ViewModel,  because I'm working in a project with RxUI I'm more inclined to set a ViewModel than a `BindingContext`.

When setting up a controller in a non-Forms project I'd set up bindings, assign a ViewModel and be done with it. With this implementation I'd set a ViewModel, and nothing happens because I didn't explicitly set a `BindingContext`. So this should at least be very well documented.

kentcb:
@Qonstrukt TBH, I hoped someone would raise it so we could have the discussion :)

I was approaching this from the other side. Normally I end up with this crap in my XAML:

```XML
<ListView>
    <ListView.ItemTemplate>
        <me:SomeView ViewModel="{Binding .}"/>
    </ListView.ItemTemplate>
</ListView>
```
That is, the XAML runtime has assigned the `BindingContext` to `SomeView`, and I need to manually propagate it to `ViewModel`. I would prefer that setting either `BindingContext` or `ViewModel` propagates to the other:

```C#
public TViewModel ViewModel 
{
    get { return (TViewModel)GetValue(ViewModelProperty); }
    set
    {
        SetValue(ViewModelProperty, value);
        this.BindingContext = value;
    }
}

protected override void OnBindingContextChanged()
{
    base.OnBindingContextChanged();
    this.ViewModel = this.BindingContext as TViewModel;
}
```
However, the drawback is that it's potentially breaking. If there is existing code out there that relies on having distinct values for `ViewModel` and `BindingContext`, that code will break. Probably a good thing in my opinion, since it seems an awful practice, but there you go.

To avoid the breaking change, we _could_ use the code I've already committed as well as the change to the `ViewModel` property above, but have it check that the `BindingContext` is not already set to some other object of a type other than `TViewModel`. Frankly, it gets a bit messy and somewhat confusing for users I think. I'd rather just say `BindingContext` propagates to `ViewModel` and vice-versa.

If we can all agree this is the most desirable approach, I'm all for it.

Qonstrukt:
Fair points 😊

I agree that having the `BindingContext` and `ViewModel` properties be equal to each other for most of cases should be preferred, whichever way you set them.

The only case I can imagine is your ViewModel having a property which defines the `BindingContext` for your View. That's why I'd prefer to have the `ViewModel` set the `BindingContext` and not the other way around. 

It'd be interesting to hear what others like @escamoteur expect who use Forms a lot more than I do.

kentcb:
Agreed - further input appreciated.

I was also just thinking that we should probably look to the WPF/UWP RxUI implementations for guidance here, as they have exactly the same problem. I'll look into the implementation tomorrow.

PureWeen:
On all my Xamarin Forms Views I set it both places  
VM => BindingContext     when VM is set
BindingContext => VM    when OnBindingContextChanged occurs

I found if I try to be aware of the scenarios of when to do each I just mess it all up or forget to wire the BindingContext back to the VM in cases like a ListView.  It's easier going in just knowing those are going to be set.

Plus I think when you start implementing RxUI one just assumes that's how it's going to work and the BindingContext will be set. Especially during cases of VM navigation.  The detachment of the BindingContext from the VM is a little confusing first and will probably be a little more confusing if it works one way and not the other

The ViewModel is really just there to be a typed BindingContext so you get a FluentApi over the bindings right?  It doesn't really offer any additional behavior/purpose does it?  So it seems like the two should mirror each other.

escamoteur:
I agree with Shane 100% I always see ViewModel as the same as the Forms BindingContext and so far never had the need to use the two inf different ways. That's one reason that we added an XamvvmRxUI to xamvvm that autoassigns the ViewModel to the ViewModel instead of the BindingContext property duringpage creation and VM navigation.

If RxUI fpr Forms woudl do this automatically (maybe by use Fody to automatically make ViewModel asn alias of BindingContext would be amazing.

Another important pint here is that in Forms the BindingContext automatically get's forwarded to the childviews. That's something ViewModel should also mirror.

GiusepeCasagrande:
I see no reason to make ViewModel differ from BindContext and I believe that we should discourage this.
As I see this, if you need a ViewModel other than what is setting in the BindingContext (and the other way around), something is pretty wrong with you design/code.

So I'm all in for the "breaking" solution.

escamoteur:
Agreed actually this would prevent people from doing crazy stuff

kentcb:
OK, I've just pushed an implementation as discussed. If people could try it out on their projects, that would be much appreciated.

escamoteur:
Just had a look at the code. Important for this to work is that if someone overrides the OnBindindContextChanged  he has to call the base implementation. Is it possible to issue a Warning if an overriding Method does not call the base?

escamoteur:
Another thought. The way Kent solved this problem now only works when you derive from one of this base classes if you just use IViewFor<> you get no support.
Would there be another solution for this? Bringing Fody into play again? Or require to implement somethink like The ViewModelActivator to do the assigning. As BindingContext itself is a BindableProperty it should be possible to register for changes or not?

I have no real solution right now but just changing the base classes seems not the right solution for me.

kentcb:
> Is it possible to issue a Warning if an overriding Method does not call the base?

Not to my knowledge, but it's generally considered poor practice to override a method and neglect to call the base implementation. And the IDE helps out here by filling in the base call for you when you override.

> As BindingContext itself is a BindableProperty it should be possible to register for changes or not?

WPF includes the [DependencyPropertyDescriptor](https://msdn.microsoft.com/en-us/library/system.componentmodel.dependencypropertydescriptor(v=vs.110).aspx) class, which allows you to hook into a dependency property's change notifications _externally_ from the implementation class itself. As best I can tell, XF has no equivalent. Thus, the only option is for the change handler to be triggered from within the implementation class itself.

> Bringing Fody into play again?

I don't know that Fody is an option as a dependency for RxUI itself. Certainly I wouldn't be too keen on having the IL rewritten.

PureWeen:
AFAIK all Bindable Property changes flow through the OnPropertyChanged events 

https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/BindableObject.cs#L588

So in theory I think one could subscribe to PropertyChanged in the constructor and then look for "ViewModel" or "BindingContext"

Though I think I would also lean on this perspective

> Not to my knowledge, but it's generally considered poor practice to override a method and neglect to call the base implementation. And the IDE helps out here by filling in the base call for you when you override.

If you override something and don't call the base implementation and things start acting weird then that's on you :-)

escamoteur:
OK that's the one point,  the more important one for me that this will require to use one of the base classes. So if I have already a forms base class I'm f***** 
I don't think Fody would be so bad.  Realm for instance realizes it's whole c# binding this way

Qonstrukt:
To be honest I think that's a never ending fight. If your base classes extend from the RxUI base classes you're ok. But if you're using base classes of some other framework, it's the choice you made, and you should implement your own combination.

I'm currently looking into combining functionality of MvvmCross and RxUI Activities and ViewControllers, and no-doubt I'll have to make some compromises there as well. Luckily @aritchie made a start with this already here: https://github.com/aritchie/mvxgoodies. But at some point you'll still have to make a choice for which implementation is more important to you.

Stuff like Fody is a nice bonus, there's no reason why there couldn't be a base class and a Fody attribute which produces the same end-result, but that belongs in a separate package IMHO.

escamoteur:
I don not completely agree and I thing Forms is special in that it already provides a Binding Infrastructur only that it uses another propterty for Bindingcontextes as RxUI so IMHO to have an automatic mechanism included in the Forms Verison of RxUI would be great.

Qonstrukt:
But isn't that what this PR exactly accomplishes? 😉 
But only if you use the `Reactive*Page/View` classes.

escamoteur:
yes and in that I see a problem because if you get a new Page or View base class from some other Library like PopupPages or SlideOverkit you have to care about it yourself. In making adoption of RxUIForms easy it would be great to have other solutions as well. Still thinking if Foody wouldn't be the best solution. You won't even need an Attribute because it could look for the IViewFor attribute

kentcb:
Fody has its own ecosystem of add-ins. Such an add-oin is outside the scope of this PR, and doesn't even belong in the RxUI code base. If you wanted to PR something, I suspect the best place to do so would be [here](https://github.com/kswoll/ReactiveUI.Fody).

As @Qonstrukt mentioned, if you step away from what the framework provides, then it's up to you to fill in any gaps.

kentcb:
OK, I've tried this locally and it all looks good. If there are no objections, I'll merge this tomorrow.

PR #1274, opened 2017-02-15
fix: ensure synchronous command execution is lazy

kentcb:

Bug fix.

Synchronous commands execute their workload immediately upon a call to `Execute` instead of requiring a subscription.

Synchronous commands now require a subscription in order for their workload to execute.

The documentation clearly states that subscription is required before commands execute, and that is indeed how asynchronous commands work. So this change brings the code into line with the documentation.

Technically, I suppose I'd have to say yes. However, the circumstances under which this would break someone are:

1. They are using a synchronous command (which are less common than asynchronous commands, in my experience).
2. They are executing the command and not subscribing (which is less common than binding the command or using `InvokeCommand`).

So anyone who does break isn't using the framework correctly and would be bitten if, for example, they tried the same thing with an asynchronous command. Also, if they're subscribed to `IsExecuted` or to the command itself, those subscriptions will never fire if they are not subscribing.


Qonstrukt:
It's quite inconsistent alright, but to be fair, I always thought it was a design decision. Since the code in the command isn't asynchronous, there's not much reason to "subscribe" onto it. But that's an assumption on my part!

The reason I'm telling is because we currently do have (some) code that functions without the Subscribe part because we thought it was by design. And hastily upgrading RxUI would render those functions un-executed, maybe even without noticing. Since unit-testing platform-code is still a hassle it might very well go undetected. (Since the caller is often the View in the platform-code.)

I'm not saying this is a bad change, I'm all for consistency, but we need to make sure to inform users well of this. 👍

PR #1285, opened 2017-02-22
fix: assign VM from correct thread in RoutedViewHost

PureWeen:

Removed a buggy checkin with attempting to replace TPL with ToObservable.  I tried changing the SelectMany to correctly return the Observables and the Do went back to being on the Thread Pool.  So for now I just moved the assignment of the VM up to the finally block of the SelectMany

The VM gets assigned in the Do block. 

It shouldn't

**Other information**:

### DIFF
--- a/src/ReactiveUI.XamForms/XamForms/RoutedViewHost.cs
+++ b/src/ReactiveUI.XamForms/XamForms/RoutedViewHost.cs
@@ -94,7 +94,7 @@ namespace ReactiveUI.XamForms
                         }
 
                         popToRootPending = false;
-                        return Unit.Default;
+                        return x;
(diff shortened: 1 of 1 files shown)


PureWeen:
So here's a really simple repro of the issue I'm referring to

https://github.com/PureWeen/TaskToObservables

On Android it doesn't throw an exception but it still sets the label from the ThreadPool 

on iOS you'll see it throws an exception and then you just can't really do anything with the label

Basically all my repro does is push 5 pages on a stack. Click the button. It pops to root and then tries to set the label afterwards.

kentcb:
Thanks @PureWeen

Issue #1156, opened 2016-09-05
Use pre-canned, cold observables to reduce memory pressure


kentcb:
**Do you want to request a _feature_ or report a _bug_?**

Feature.

**What is the current behavior?**

Commonly required cold observables such as empty, never, and default are new'd up each and every time they are required.

**What is the expected behavior?**

To reduce memory pressure, cold observables can be created once and shared. See [Genesis.Observables](https://github.com/kentcb/Genesis.Observables) for an example of what I mean. ReactiveUI can just define similar constants internally and use them wherever relevant.

**What is the motivation / use case for changing the behavior?**

Reducing memory pressure, particularly for mobile scenarios.

PR #1289, opened 2017-02-24
perf: use shared, static observables to reduce allocations

kentcb:

Performance improvement.

#1156

No.

**Other information**:

Let's allocate less!

### DIFF
--- a/src/ReactiveUI/Observables.cs
+++ b/src/ReactiveUI/Observables.cs
@@ -27,7 +27,7 @@
     /// <summary>
     /// Provides commonly required, statically-allocated, pre-canned observables.
     /// </summary>
-    public static class Observables
+    internal static class Observables
     {
         /// <summary>
         /// An observable that ticks a single, Boolean value of <c>true</c>.
@@ -42,7 +42,7 @@
         /// This observable is equivalent to <c>Observable&lt;bool&gt;.Default</c>, but is provided for convenience.
         /// </para>
         /// </remarks>
-        public static readonly IObservable<bool> False = Observable<bool>.Default;
+        public static readonly IObservable<bool> False = Observable.Return(false);
 
         /// <summary>
         /// An observable that ticks <c>Unit.Default</c> as a single value.</summary>
--- a/src/ReactiveUI/Observables.cs
+++ b/src/ReactiveUI/Observables.cs
@@ -1,0 +1,56 @@
+﻿namespace System.Reactive.Linq
+{
+    /// <summary>
+    /// Provides commonly required, statically-allocated, pre-canned observables.
+    /// </summary>
+    /// <typeparam name="T">
+    /// The observable type.
+    /// </typeparam>
+    internal static class Observable<T>
+    {
+        /// <summary>
+        /// An empty observable of type <typeparamref name="T"/>.
+        /// </summary>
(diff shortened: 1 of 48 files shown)


src/ReactiveUI/ReactiveUI_UWP.csproj
```diff
@@ -72,6 +72,7 @@
     <Compile Include="MessageBus.cs" />
     <Compile Include="SuspensionHost.cs" />
     <Compile Include="ObservableAsPropertyHelper.cs" />
+    <Compile Include="Observables.cs" />
```
olevett: This saddens me (the fact that we need to do it everywhere :()

src/ReactiveUI/Observables.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <summary>
+        /// An observable of type <typeparamref name="T"/> that ticks a single, default value.
+        /// </summary>
+        public static readonly IObservable<T> Default = Observable.Return(default(T));
+    }
+
+    /// <summary>
+    /// Provides commonly required, statically-allocated, pre-canned observables.
+    /// </summary>
+    public static class Observables
```
olevett: Should this be internal (like the `Observable<T>` class)?

src/ReactiveUI/Observables.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// An observable that ticks a single, Boolean value of <c>true</c>.
+        /// </summary>
+        public static readonly IObservable<bool> True = Observable.Return(true);
+
+        /// <summary>
+        /// An observable that ticks a single, Boolean value of <c>false</c>.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This observable is equivalent to <c>Observable&lt;bool&gt;.Default</c>, but is provided for convenience.
```
olevett: Is there a reason to `Default` rather than `Observable.Return(false)` don't think it really matters, but it seems a little odd that it's different for the true/false cases.

src/ReactiveUI/ReactiveUI_UWP.csproj
kentcb: Totally! Bring on VS17.

src/ReactiveUI/Observables.cs
kentcb: It should! Thanks - I copied this from Genesis.Observables and forgot to change both.

src/ReactiveUI/Observables.cs
kentcb: Probably a completely pointless optimization. That is, if it's `Observable.Return(false)` then someone access `Observable<bool>.Default`, then we've cached two instead of one.

src/ReactiveUI/ReactiveUI_UWP.csproj
olevett: It's doable with shared projects now...probably not worth the faff of doing it twice though!

src/ReactiveUI/Observables.cs
olevett: That does make sense, hadn't quite spotted that!

kentcb:
Thanks for the review @olevett!

PR #1291, opened 2017-02-25
Ship ReactiveUI v7.2

kentcb:


ghuntley:
Ready to ship Kent. All issues on the milestone have a single label, mashing that merge button will generate you the release notes and after you press.publish the release will.be public. I needed.to remove.This particular issue from the milestone as the shipping PR shouldn't be assigned to a milestone. Otherwise documentation will fail to generate because the ship PR isn't assigned a label :)

kentcb:
@ghuntley interesting note about removing this issue from a milestone. What prompted me to do that was this line in the docs:

> Verify, once more that the GitHub issue milestone is correct as that will be the version of the software release

Is that talking about something else then?

ghuntley:
K. That needs some clarification. Whats needed is the $version of the milestone needs to be the correct version I.e. 7.2.0 as it is.

Issue #1275, opened 2017-02-15
Duplicate key issue in `controlIds` dictionary in ControlFetcherMixin.cs in ReactiveUI_Android


ginev:
I think this is an issue with the way the `controlIds` dictionary is populated in ControlFetcherMixin.cs: https://github.com/reactiveui/ReactiveUI/blob/develop/src/ReactiveUI/Android/ControlFetcherMixin.cs#L34-L36

In a Xamarin.Android project that uses ReactiveUI, referencing Xamarin.Android assemblies that have resource IDs which are the same word but with different casing the following exception is thrown as soon as the static constructor of `ControlFetcherMixin.cs` is executed:

```
02-15 11:45:16.138 E/mono-rt ( 4726): [ERROR] FATAL UNHANDLED EXCEPTION: System.TypeInitializationException: The type initializer for 'ReactiveUI.ControlFetcherMixin' threw an exception. ---> System.ArgumentException: An item with the same key has already been added.
02-15 11:45:16.138 E/mono-rt ( 4726):   at System.ThrowHelper.ThrowArgumentException (System.ExceptionResource resource) [0x00000] in /Users/builder/data/lanes/3511/501e63ce/source/mono/mcs/class/referencesource/mscorlib/system/throwhelper.cs:72 
02-15 11:45:16.138 E/mono-rt ( 4726):   at System.Collections.Generic.Dictionary`2[TKey,TValue].Insert (TKey key, TValue value, System.Boolean add) [0x0008e] in /Users/builder/data/lanes/3511/501e63ce/source/mono/mcs/class/referencesource/mscorlib/system/collections/generic/dictionary.cs:336 
02-15 11:45:16.138 E/mono-rt ( 4726):   at System.Collections.Generic.Dictionary`2[TKey,TValue].Add (TKey key, TValue value) [0x00000] in /Users/builder/data/lanes/3511/501e63ce/source/mono/mcs/class/referencesource/mscorlib/system/collections/generic/dictionary.cs:192 
[...]

PR #1302, opened 2017-03-03
Clarified exception when WireUpControls fails

Qonstrukt:
This clarifies the exception you get when `ControlFetcherMixin` collects the names of resource ID's in your layout files. It shows you for which ID's it's exactly failing so it's easier to find which names are the culprit.

When you have duplicate names with different casings in use as resource ID's, you get a fairly generic `ArgumentException` which is hard to understand without diving into the code of `ControlFetcherMixin`.

We throw a more explanatory `InvalidOperationException` which shows which names should be fixed to get things going.

Nope, unless someone is already catching the `ArgumentException` themselves, but I can't imagine anyone relying on such functionality.

**Other information**:
As reported in #1275, and we ran into this ourselves as well. Cost multiple of our engineers a good couple of minutes to figure out what's wrong.


ghuntley:
Thanks @Qonstrukt!

Issue #1299, opened 2017-02-28
ViewLocator can't resolve views specified by UWP RoutedViewHost


olevett:
***Note*: for support questions, please ask on StackOverflow: https://stackoverflow.com/questions/tagged/reactiveui . This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
RoutedViewHost on UWP uses the IRoutableViewModel type to specify what view model type to return for ViewLocator calls, which ends up failing

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
```csharp
public interface IRoutableFooViewModel : IRoutableViewModel { }

public class RoutableFooViewModel : ReactiveObject, IRoutableFooViewModel
{
    public IScreen HostScreen { get; set; }
    public string UrlPathSegment { get; set; }
}

public class RoutableFooView : IViewFor<IRoutableFooViewModel>
{
    object IViewFor.ViewModel
    {
        get { return ViewModel; }
        set { ViewModel = (IRoutableFooViewModel)value; }
    }
    public IRoutableFooViewModel ViewModel { get; set; }
}

public void CanResolveViewFromViewModelWithIRoutableViewModelType()
{
	var resolver = new ModernDependencyResolver();

	resolver.InitializeSplat();
	resolver.InitializeReactiveUI();
	resolver.Register(() => new RoutableFooView(), typeof(IViewFor<IRoutableFooViewModel>));

	using (resolver.WithResolver()) {
		var fixture = new DefaultViewLocator();
		var vm = new RoutableFooViewModel();

		var result = fixture.ResolveView<IRoutableViewModel>(vm);
		Assert.IsType<RoutableFooView>(result);
	}
}
```

**What is the expected behavior?**
That this type can be resolved...

**What is the motivation / use case for changing the behavior?**

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the latest stable and snapshot (http://docs.reactiveui.net/en/contributing/snapshot/index.html) versions.**
Any version other than 7.2.

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
I think the only change required would be to change ViewLocator to use `GetType()` rather than `typeof(T)` on line 170.

olevett:
It's possible this has been resolved by #1288 but I'm not entirely certain...

Issue #1298, opened 2017-02-28
ViewLocator can throw null reference exceptions


olevett:
***Note*: for support questions, please ask on StackOverflow: https://stackoverflow.com/questions/tagged/reactiveui . This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
DefaultViewLocator.AttemptViewResolutionFor can throw a null reference exception if ToggleViewModelType returns null.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

**What is the expected behavior?**
Doesn't throw an exception (either, ToggleViewModelType doesn't return null, or there's a null check around it in ResolveView).

**What is the motivation / use case for changing the behavior?**
Not throwing exceptions when they shouldn't be thrown.

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the latest stable and snapshot (http://docs.reactiveui.net/en/contributing/snapshot/index.html) versions.**
7.2

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

SProst:
I am currently able to reproduce this. Going to have to roll back to 7.1 due to a significant change in behavior (it was not throwing the null reference before 7.2).

PR #1311, opened 2017-03-07
View locator changes

olevett:
Fix

#1299 and #1298 the view locator doesn't correctly resolve a couple of cases, and can't handle oddly named interfaces well.

Handle IRoutableViewModel style VMs, and handle interfaces that don't meet conventions.

It sort of undoes a breaking change introduced in 7.2 with the refactorings around the view locator.

**Other information**:
This could conceivably be two separate PRs, but I think it's actually clearer to do it in one

### DIFF
--- a/src/ReactiveUI/ViewLocator.cs
+++ b/src/ReactiveUI/ViewLocator.cs
@@ -106,7 +106,10 @@ namespace ReactiveUI
             }
 
             var toggledType = ToggleViewModelType(viewModel);
-            view = this.AttemptViewResolutionFor(toggledType, contract);
+
+            if (toggledType != null) {
+                view = this.AttemptViewResolutionFor(toggledType, contract);
+            }
 
             if (view != null) {
                 return view;
--- a/src/ReactiveUI/ViewLocator.cs
+++ b/src/ReactiveUI/ViewLocator.cs
@@ -105,11 +105,13 @@ namespace ReactiveUI
                 return view;
             }
 
-            var toggledType = ToggleViewModelType(viewModel);
+            view = this.AttemptViewResolutionFor(ToggleViewModelType(viewModel.GetType()), contract);
 
-            if (toggledType != null) {
-                view = this.AttemptViewResolutionFor(toggledType, contract);
+            if (view != null) {
+                return view;
             }
+
+            view = this.AttemptViewResolutionFor(ToggleViewModelType(typeof(T)), contract);
 
             if (view != null) {
                 return view;
@@ -121,6 +123,7 @@ namespace ReactiveUI
 
         private IViewFor AttemptViewResolutionFor(Type viewModelType, string contract)
         {
+            if (viewModelType == null) return null;
             var viewModelTypeName = viewModelType.AssemblyQualifiedName;
             var proposedViewTypeName = this.ViewModelToViewFunc(viewModelTypeName);
             var view = this.AttemptViewResolution(proposedViewTypeName, contract);
@@ -170,9 +173,8 @@ namespace ReactiveUI
             }
         }
 
-        private static Type ToggleViewModelType<T>(T viewModel)
+        private static Type ToggleViewModelType(Type viewModelType)
         {
-            var viewModelType = typeof(T);
             var viewModelTypeName = viewModelType.AssemblyQualifiedName;
 
             if (viewModelType.GetTypeInfo().IsInterface) {
--- a/src/ReactiveUI/ViewLocator.cs
+++ b/src/ReactiveUI/ViewLocator.cs
@@ -78,7 +78,7 @@ namespace ReactiveUI
         /// </item>
         /// <item>
         /// <description>
-        /// Repeat steps 1 and 2 with the type resolved from the modified name.
+        /// Repeat steps 1-4 with the type resolved from the modified name.
         /// </description>
         /// </item>
         /// </list>
--- a/src/ReactiveUI.Tests/ViewLocatorTests.cs
+++ b/src/ReactiveUI.Tests/ViewLocatorTests.cs
@@ -1,5 +1,4 @@
 using System;
-using ReactiveUI;
 using Splat;
 using Xunit;
 
@@ -71,6 +70,10 @@ namespace ReactiveUI.Tests
 
         public IFooViewModel ViewModel { get; set; }
     }
+
+    public interface StrangeInterfaceNotFollowingConvention { }
+
+    public class StrangeClassNotFollowingConvention : StrangeInterfaceNotFollowingConvention { }
 
     public class DefaultViewLocatorTests
     {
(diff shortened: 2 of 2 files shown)


src/ReactiveUI/ViewLocator.cs
```diff
@@ -93,31 +95,27 @@ public DefaultViewLocator(Func<string, string> viewModelToViewFunc = null)
         public IViewFor ResolveView<T>(T viewModel, string contract = null)
             where T : class
         {
-            var view = this.AttemptViewResolutionFor(viewModel.GetType(), contract);
+            var view = GetCandidateViewModelTypes(viewModel)
```
olevett: The intermediate state is possibly clearer from a review point of view, the big change is adding handling around `vm.GetType()` for the case that we fall through to ToggleViewModel type (i.e. respect the runtime type)

src/ReactiveUI/ViewLocator.cs
```diff
@@ -93,31 +95,27 @@ public DefaultViewLocator(Func<string, string> viewModelToViewFunc = null)
...
-            if (view != null) {
-                return view;
-            }
+            if (view != null) return view;
 
             this.Log().Warn("Failed to resolve view for view model type '{0}'.", typeof(T).FullName);
             return null;
         }
 
+        private IEnumerable<Type> GetCandidateViewModelTypes<T>(T viewModel)
```
olevett: I think this refactoring makes it a little clearer what we're actually doing, happy to revert it though.

src/ReactiveUI/ViewLocator.cs
```diff
@@ -93,31 +95,27 @@ public DefaultViewLocator(Func<string, string> viewModelToViewFunc = null)
...
+        {
+            yield return viewModel.GetType();
+            yield return typeof(T);
+            yield return ToggleViewModelType(viewModel.GetType());
+            yield return ToggleViewModelType(typeof(T));
+        }
+
         private IViewFor AttemptViewResolutionFor(Type viewModelType, string contract)
         {
+            if (viewModelType == null) return null;
```
olevett: This short circuit handles #1298 we could instead handle the null better out of ToggleViewModelType, but I think it's a reasonable thing to have.

src/ReactiveUI/ViewLocator.cs
kentcb: I agree it reads better, but I will say that I started with an enumerable in my own implementation, but then removed it to avoid allocs. I haven't double-checked, but I'm pretty sure the previous implementation doesn't alloc (except `Type` objects, which I believe will be cached by the runtime) whereas this one will alloc with every view resolution.

Do we care? I'm not sure.

olevett: That's an interesting point, and quite probably right - I'm happy to not use enumerables, and just do the set of null checks if we're concerned? (The last commit is the refactoring, but shouldn't functionally change things)

I can try and do some benchmarks, not sure how much this will tell us though?

kentcb: Yeah, benchmarkdotnet could tell us precisely what's going on in terms of memory allocs. Only problem is, we don't really have the infrastructure in place to run it in the solution as it stands. You could easily add a quick check, but it might be throwaway code until we formally decide how to incorporate perf tests into the solution.

Silv3rcircl3:
Maybe it would be a good idea to split it into two parts, ViewTypeLocator and ViewLocator. 
One idea I have is to use a ordinary ContentControl + TemplaceSelector to create a template for every unique view model and that is currently not possible with the ViewLocator since it always returns a instance of the view while I only need the type.

src/ReactiveUI/ViewLocator.cs
olevett: Benchmark results

```
Host Process Environment Information:
BenchmarkDotNet.Core=v0.9.9.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Xeon(R) CPU E3-1231 v3 3.40GHz, ProcessorCount=8
Frequency=3320309 ticks, Resolution=301.1768 ns, Timer=TSC
CLR=MS.NET 4.0.30319.42000, Arch=32-bit RELEASE
GC=Concurrent Workstation
JitModules=clrjit-v4.7.2046.0

Type=ViewResolutionBenchmark  Mode=Throughput

                Method |    Median |    StdDev | Scaled | Scaled-SD |  Gen 0 | Gen 1 | Gen 2 | Bytes Allocated/Op |
---------------------- |---------- |---------- |------- |---------- |------- |------ |------ |------------------- |
               Default | 5.0061 us | 0.0359 us |   1.00 |      0.00 | 610.00 |     - |     - |             890.47 |
                Pr1311 | 4.9428 us | 0.0277 us |   0.99 |      0.01 | 610.20 |     - |     - |             888.82 |
 EnumerableViewLocator | 5.3197 us | 0.1377 us |   1.06 |      0.03 | 687.45 |     - |     - |           1,000.36 |
```

Where Pr1311 is (solid naming, I know) the functional changes, without IEnumerable, and the Enumerable test is with wrapping into an IEnumerable.

```
 public interface IFooViewModel { }

        public class FooViewModel : ReactiveObject, IFooViewModel { }

        public interface IFooView : IViewFor<IFooViewModel> { }

        public class FooView : IFooView
        {
            object IViewFor.ViewModel
            {
                get { return ViewModel; }
                set { ViewModel = (IFooViewModel)value; }
            }
            public IFooViewModel ViewModel { get; set; }
        }

        private readonly ModernDependencyResolver resolver = new ModernDependencyResolver();
        private readonly IDisposable resolverDisposable;
        private readonly DefaultViewLocator defaultViewLocator = new DefaultViewLocator();
        private readonly PR1311ViewLocator pr1311ViewLocator = new PR1311ViewLocator();
        private readonly EnumerableViewLocator enumerableViewLocator = new EnumerableViewLocator();
        private readonly FooViewModel viewModel = new FooViewModel();

        public ViewResolutionBenchmark()
        {
            resolver.InitializeSplat();
            resolver.Register(() => new FooView(), typeof(IViewFor<FooViewModel>));
            resolverDisposable = resolver.WithResolver();
        }

        [Benchmark(Baseline = true)]
        public object Default() => defaultViewLocator.ResolveView(viewModel);

        [Benchmark]
        public object Pr1311() => pr1311ViewLocator.ResolveView(viewModel);

        [Benchmark]
        public object EnumerableViewLocator() => enumerableViewLocator.ResolveView(viewModel);

        public void Dispose()
        {
            resolverDisposable.Dispose();
        }
```

We're making more allocations (10%) but I think it'll depend on route through, I can do more benchmarks if wanted, but on the basis of this I'm happy to not do the Enumerable refactoring/

olevett:
@Silv3rcircl3 I quite like that as an idea, might be worth raising as a separate ticket, in the spirit of smaller changes.

Silv3rcircl3:
@olevett cool, I'll send a PR once this one was merged

src/ReactiveUI/ViewLocator.cs
kentcb: Thanks for this @olevett. Yeah, agreed, we should avoid alloc'ing where possible, especially at such a low level of the framework. Every little bit helps and all. I know we haven't been particularly vigilant about this kind of thing elsewhere, but it might be time to tighten up. This will be easier to manage once we formalise the inclusion of Benchmarkdotnet into the tests.

src/ReactiveUI/ViewLocator.cs
```diff
@@ -93,31 +95,27 @@ public DefaultViewLocator(Func<string, string> viewModelToViewFunc = null)
         public IViewFor ResolveView<T>(T viewModel, string contract = null)
```
kentcb: Do the API docs need to be updated inline with the functional changes?

olevett: Had a quick double check, and have made the only needed change (I think the docs included more functionality than the code did!)

src/ReactiveUI/ViewLocator.cs
olevett: Pushed without the IEnumerable now

kentcb:
Thanks @olevett - much appreciated!

PR #1327, opened 2017-03-25
First pass at updating to VS2017 csproj file

olevett:
So I've had a first pass at moving over to the new csproj formats...apologies for how sweeping the changes are, I've tried to minimise the actual change.

There are a couple of things I'm unsure about, and a couple of things that need working on, but wanted to get thoughts before I spend too long.

# Things to do
- [x] Make sure Cake/CI builds correctly - part of the reason for this PR
- [x] Update nuspec files
- [x] Use common version info - this is a casualty of the new msbuild process, not sure whether it's hugely urgent
- [x] Move test project into new csproj
- [x] Reduce duplicated files
- [x] Check that the API surface hasn't changed
- [x] Build EventBuilder and ReactiveUI.Events
- [x] Look at whether using msbuild directly is bad (Cake seems to do something strange when calling msbuild but I haven't worked out exactly why that is)
- [x] Tidy up, I've done some of this, e.g. Directory.build.props, but the cake build script can probably be made nicer
- [x] Make sure nuget packages are depending on the correct version of each other

# Notes
- Following from the examples from @onovotny, I've moved some files around into "Platform" subfolders, but because of the heavy use of #IFDEFs, there are some files that are now duplicated and I can't quite work out where the best place is. There's also a potentially slightly confusing set of folders (e.g. "windows-common", "apple-common") but again, I'm not sure what the best way to line these up is
- WP8/Win8 has been 🔥 
- I don't have ncrunch, so I've probably broken that
- Playground projects have been 🔥 
- I've removed ReactiveAssert as nothing was referencing it
- I've replaced Console.Log in some tests with Debug.Log as it was causing compiler errors


ghuntley:
Can you burn/delete projects+code related Windows Store 8.1 and Windows Phone whilst you are in there. You'll find that they won't even load in 2017.

ghuntley:
Thanks for taking lead on this, happy to jump on Skype and pair with you. 👍

onovotny:
I can help too....but why is https://github.com/olevett/ReactiveUI/tree/vs2017-upgrading/src/Microsoft.Reactive.Testing that there? Please use the NuGet.

onovotny:
I would make the update include the Rx 3.0 stuff too as it removes tons of cruft.

onovotny:
https://github.com/olevett/ReactiveUI/blob/vs2017-upgrading/src/ReactiveUI/ReactiveUI.csproj#L85 <-- that should be 5.2.2 or 5.2.3 to work well with netstandard stuff.

onovotny:
Most of this can be removed: https://github.com/olevett/ReactiveUI/blob/vs2017-upgrading/src/ReactiveUI/ReactiveUI.csproj#L19-L53

They're all in by default with the MSBuild.Sdk.Extras.

 You only might need some of your custom DefineConstants. You also may want your TargetPlatformVersion higher for UAP.

ghuntley:
Thoughts: 
- Migrate to VS2017 in this PR, 
- Minimise deletion of code for old platforms, that can be as seperate PR (less change to codebase)
- My gut says don't upgrade to Rx3/4 or netstandard in this PR. Let's aim for a stable release using existing dependencies?
- Migrate nuget packaging as need be; drop packaging for windows phone and windows 8.1 aka winrt.
- update build script.

onovotny:
@ghuntley Rx 3 is a must IMO. I would not release without it and it gets rid of tons of old nuget cruft.

ghuntley:
Can you burn/delete projects+code related Windows Store 8.1 and Windows Phone whilst you are in there. You'll find that they won't even load in 2017.

ghuntley:
@onovotny RX3 definitely next public release on nuget.org, my gut says it wpuls.Be safer/thinner vertical slice to do it as a seperate PR.

onovotny:
@ghuntley check your slack.

ghuntley:
Note: we will do one final release of 7.x before merging this PR because there are some scheduler bugs which consumers need the fix to. 

This PR becomes when merged progresses the project towards a 8.0 release once we have done an appropriate level of manual testing. As soon as this PR is merged that's where the line in the sand is drawn - no more 7.x releases and the contents of this PR will be automatically published to the MyGet feed.

olevett:
So I've got the build passing now, things still left though (that I've broken/not got quite behaving)
- Build EventBuilder and ReactiveUI.Events
- Look at whether using msbuild directly is bad (Cake seems to do something strange when calling msbuild but I haven't worked out exactly why that is)
- Tidy up, I've done some of this, e.g. Directory.build.props, but the cake build script can probably be made nicer
- Check API surface

olevett:
I've run through approval tests without doing merging, and it looks like there are a couple of differences
- internals visible to are different - expected
- XamlGeneratedNamespace is missing, I don't actually know what this is for...thoughts?
- everything else is the same

```
namespace XamlGeneratedNamespace
{
    public sealed class GeneratedInternalTypeHelper : System.Windows.Markup.InternalTypeHelper
    {
        public GeneratedInternalTypeHelper() { }
        protected override void AddEventHandler(System.Reflection.EventInfo eventInfo, object target, System.Delegate handler) { }
        protected override System.Delegate CreateDelegate(System.Type delegateType, object target, string handler) { }
        protected override object CreateInstance(System.Type type, System.Globalization.CultureInfo culture) { }
        protected override object GetPropertyValue(System.Reflection.PropertyInfo propertyInfo, object target, System.Globalization.CultureInfo culture) { }
        protected override void SetPropertyValue(System.Reflection.PropertyInfo propertyInfo, object target, object value, System.Globalization.CultureInfo culture) { }
    }
}
```

olevett:
NuGet/Home#4337 and NuGet/Home#4790 are relevant to this

kentcb:
This is freaking awesome @olevett - thanks for all the hard work.

Apologies - I just noticed a requested review. I've been snowed under preparing for a product launch end of next week. Once that's done with, I'll find some time to review this.

Can't freaking wait to move to nice project files+netstandard+Rx3 etc. This is a massive step in that direction.

olevett:
This latest commit is a bit of a "I give up" on getting GitVersioning to do what would be nice. Because of some...quirks...in the packaging pipeline as documented on the nuget tickets mentioned before, the version gets baked in at the wrong point. By passing it in explicitly it's at least consistent and correct (and this more closely matches how it was working before all these changes, though we're passing it through msbuild rather than an AssemblyInfo file).

I think it's worth getting to work with the GitVersionTask in MSBuild and that should be a bit simpler and make locally builds from VS2017 closer to cake builds, but not sure if it should block merging this PR.

olevett:
To make things a bit easier to review, I've done some commit squashing and moving things around, and raised as a separate PR (#1332), so I'm closing this in favour of that one.

Issue #1295, opened 2017-02-26
Issue Building ReactiveUI with a pure VS2017 machine (NO VS2015 ever)


ChrisPulman:
***Note*: for support questions, please ask on StackOverflow: https://stackoverflow.com/questions/tagged/reactiveui . This repository's issues are reserved for feature requests and bug reports.**

**Do you want to request a *feature* or report a *bug*?**
Contributers Issue

**What is the current behavior?**
Will open and build using VS2015 with correct Android SDK's installed and Win Mobile 8.1 SDK.
However when opened on a Windows 10 machine with only VS2017 Enterprise + all relevant add-on's fails to build all Andriod projects, ReactiveUI.Blend_WinRT, and ReactiveUI.Blend_Net45 project
execute Build.bat and get a failed build on a machine without VS2015 installed

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

**What is the expected behavior?**
Open with VS2017 restore packages and Build resulting in a successful build
Open with VS2015 restore packages and Build resulting in a successful build
execute Build.bat and get a successful build

**What is the motivation / use case for changing the behavior?**
VS2017 is going to be released on March 7th, it has many beneficial features and support for C#7, there is also a Mac version of this (I don't personally own a Mac but... I am starting to see a need to have one even if I only use it for testing)

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the latest stable and snapshot (http://docs.reactiveui.net/en/contributing/snapshot/index.html) versions.**
Source code

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
![reactiveuibuilderrorwithvs2017](https://cloud.githubusercontent.com/assets/4910015/23340027/a8f94922-fc26-11e6-93ef-2c43ca7cae3c.PNG)
![reactiveuiideerrorwithvs2017](https://cloud.githubusercontent.com/assets/4910015/23340062/8cd28b04-fc27-11e6-847c-5799afa386dd.PNG)

Possible fix for some of the issues would be to add a nuget package Microsoft.Xaml.Interactivity.Unofficial for the ReactiveUI.Blend_WinRT project and System.Windows.Interactivity.WPF package for the ReactiveUI.Blend_Net45 project.
[...]

Issue #1123, opened 2016-06-18
use new rx bits in rxui v8


ghuntley:
This will be the last breaking change allowed in the release, once merged we will enter into a 30 day alpha and then release v7 on NuGet.org

https://www.dotnetfoundation.org/blog/rx-net-welcome

tiksn:
How can we access alpha/beta version? will it be preview package on nugget or in myget?

onovotny:
Just one update...the latest bits are Rx 3.1 ;)

oliverw:
This was not intended. Please undo. My change. Sorry about that.

PR #1332, opened 2017-04-10
Updating to VS2017 csproj etc

olevett:
This is essentially a reordering/recombining of PR #1327 to make it a bit easier to review/follow and probably a bit easier to merge.

I've done this because I had a lot of back and forth on the initial PR and this should just be a bit nicer.

## Notes
- each commit probably doesn't build on its own, I don't want to spend time making that happen...
- It's probably not worth spending much time looking at 38da0d8, as it's just deleting a load of things that we won't need going forwards - this could probably be a completely seperate PR but I think there might be some odditites during the build...
- There are a few things on the original PR that are of note
  - WP8/Win8 has been 🔥
  - Following from the examples from @onovotny, I've moved some files around into "Platform" subfolders, but because of the heavy use of #IFDEFs, there are some files that are now duplicated and I can't quite work out where the best place is. There's also a potentially slightly confusing set of folders (e.g. "windows-common", "apple-common") but again, I'm not sure what the best way to line these up is
  - Playground projects have been 🔥 
  - I've removed ReactiveAssert as nothing was referencing it
  - I've replaced Console.Log in some tests with Debug.Log as it was causing compiler errors
- This upgrades to Rx3 and Splat 2


build.cake
```diff
@@ -177,7 +166,7 @@ Task("GenerateEvents")
             int success = 0;    // exit code aka %ERRORLEVEL% or $?
             if (process.GetExitCode() != success)
             {
-                Error("Failed to generate events for '{0}'", platform);
+                Error("Failed to generate events for '{0}'", platform);
```
shiftkey: :fire: whitespace

build.cake
```diff
@@ -487,12 +380,12 @@ Task("PublishRelease")
 //////////////////////////////////////////////////////////////////////
 
 Task("Default")
+    .IsDependentOn("UpdateAppVeyorBuildNumber")
     .IsDependentOn("CreateRelease")
     .IsDependentOn("PublishPackages")
     .IsDependentOn("PublishRelease")
     .Does (() =>
```
shiftkey: ❔ does this `Does` function need to exist at all?

build.cake
```diff
@@ -66,38 +62,6 @@ var packageWhitelist = new[] { "ReactiveUI-Testing", "ReactiveUI-Events", "React
...
-        Symbols                  = false,
-        Verbosity                = NuGetVerbosity.Detailed,
-        OutputDirectory          = artifactDirectory,
-        BasePath                 = basePath,
-    });
-};
-
 Action<string> SourceLink = (solutionFileName) =>
 {
     GitLink("./", new GitLinkSettings() {
```
shiftkey: There's a number of GitLink errors [here](https://ci.appveyor.com/project/ghuntley/reactiveui/build/538.Branch.pull/1332/merge.Sha.e0bf30c75068aadea285a7604e40d1d405dee16c#L1951) due to what I suspect is the change in project format:


> Failed to load project 'C:\projects\reactiveui\.\src\ReactiveUI\ReactiveUI.csproj': The default XML namespace of the project must be the MSBuild XML namespace. If the project is authored in the MSBuild 2003 format, please add xmlns="http://schemas.microsoft.com/developer/msbuild/2003" to the <Project> element. If the project has been authored in the old 1.0 or 1.2 format, please convert it to MSBuild 2003 format.  C:\projects\reactiveui\src\ReactiveUI\ReactiveUI.csproj

build.cake
olevett: Not sure, wouldn't be surprised if it's something Cake needs, might try without.

build.cake
olevett: Oops, missed that one then amongst the other errors. Will take a look at that, at least some of the .NetStandard projects have it behaving.

olevett:
The XMLDOC issues is something I've been gradually clearing up, but it's more change than needed here - they're just little commits here and there though.

olevett:
I think the package names may have changed, need to double check that...

kentcb:
WOW - fantastic job @olevett! I am so pumped to see how nice things look now, and am really excited by how much easier this will make simple things like renaming some files. This is a much-needed transition for RxUI and your work here is _really_ appreciated.

Here are some specific feedback points. I hope you don't mind me listing here instead of against specific line numbers. The code review experience with a large PR seems a little clunky, so I gave up trying.

It might be that we table some of these for post-merge. I'm fine with that if that's what people prefer.

1. Has the time come to split up _shared/Registrations.cs_ such that it is specific to each platform? Right now it's a confusing mess of ifdefs. I'd even go so far as to say we should probably work towards removing as many ifdefs as we can throughout the entire project. By moving platform-specific code to _Platforms/`$PLATFORM`_ we can obviate the need for ifdefs. Technically, I think we _could_ use partial classes to remove all ifdefs, but that might be overkill and prove less readable.
2. Could we name any file containing a `PlatformRegistrations` class to _PlatformRegistrations.cs_? There are currently two different file names in use, neither of which matches the class name. I realize file names generally don't match class names throughout the code base, but that's because it embodies a feature rather than a type. In this case, it's just a type. However, I'd also be happy if they're both called _Registrations.cs_.
3. The comment at the top of _shared/Registrations.cs_ isn't accurate anymore either. Separate PR though - just thought I'd mention it because I saw it.
4. Why is _shared/Android_ capitalized when every other platform is not?
5. Should RxUI reference Rx 3.1.1 or just 3.0.0? The latter would allow consumers to choose a later version _if_ they wish, but the former requires that they reference at least 3.1.1
6. I vote to kill off the legacy stuff now. I mean, it's all marked as obsolete and if people haven't transitioned by now...
7. I wonder why we have _Resources/Resource.Designer.cs_? It's empty. I'm guessing it got generated in the old solution setup and has been copied across unnecessarily? There are multiple projects with this.
8. Can we kill off _ReactiveUI_UWP.nuget.targets_ and _ReactiveUI_UWP.project.lock.json_? Or at least move them to the appropriate platform folder?
9. Microsoft.Reactive.Testing now supports netstandard. This seems like we can finally have reactiveui-testing target netstandard, and only netstandard? This would be a big win.
10. What is the `82a7f48d-3b50-4b1e-b82e-3ada8210c358` service included by the tests project? 🤔 
11. It's not clear to me what is generating the NuGet packages. I was expecting to see the new csproj format configured to generate the package, but that appears not to be the case. And the `Package` target in the build script does not seem to do it either...

olevett:
Thanks @kentcb!

Replies by number, generally, I think I agree but wanted to not change too much code in this PR (as you say, it's really painful to review!) and instead have some follow up, more focused PRs - happy to raise as issues to keep track of?

1. I think so - this is one of the biggest benefits to the new CSPROJ I think, we should hopefully be able to lose most of the comments
2. Yeah, thing that's pretty reasonable - think there's a follow up exercise to rename files (and there are some files that contain two sort of related classes that would be nice to split and rename etc)
3. Agreed.
4. Because that's how I typed it first...will have a go at renaming, but Windows hates file names...
5. Possibly, I just picked the latest in nuget, don't think there's any disadvantage in dropping to 3.0.0 though
6. Cool, any other legacy stuff that needs to go?
7. I think it got generated later, by VS2017 I think, I think it covers the Android builds, but we can probably remove it as it's empty.
8. Yes, my bad!
9. If netstandard is a reasonable thing for the testing platform to target, that's fine - I think it used to be quite broad in targets, so didn't want to break that
10. I'm not sure! I built a new test project and it got included, it might be the live testing (which needs a newer version of xunit anyway)
11. Line 203 of the build.cake script does this (the "pack" target). The 2017 csproj just generates packages based on names etc. One thing I've only spotted recently is that the naming is different (e.g. ReactiveUI.Xaml rather than ReactiveUI-Xaml) so I need to tweak that I think

RyanMelenaNoesis:
> It might be that we table some of these for post-merge.

My vote would be to get this merged ASAP.  Once everything is updated to VS 2017 and a modern RX I think others (myself included) would be happy to assist in the clean-up / refactor effort.

olevett:
Looking at other github issues, I think this will also resolve #1295 and #1123

build.cake
onovotny: SourceLink seems more current here -- github.com/ctaggart/SourceLink

olevett: That's what Rx.Net's using isn't it?

ghuntley:
@olevett your work here is greatly appreciated. `develop` is now netstandard only and v7.x.x  retired. Time to start validation/registration testing and when 👍 this will be shipped as v8.0.0.

ghuntley:
note; I used a merge commit here as each commit was fine grained and it'll make debugging this easier when testing.

Issue #1314, opened 2017-03-08
Memory leak when nesting ViewModelViewHosts in WPF


kentcb:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

Memory leaks when nesting `ViewModelViewHost` instances. It seems the `DependencyProperty` infrastructure is holding onto an instance of the `ViewModelViewHost`, which in turn has a reference to a VM:

![image](https://cloud.githubusercontent.com/assets/1901832/23700032/7306b034-0441-11e7-9b82-e201bd754b7a.png)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

The best repro I have is the source code for my book samples, which is not currently publicly available. If someone wants to take a look at this before I do then ping me.

**What is the expected behavior?**

Memory should not leak, regardless of how nested `ViewModelViewHost` instances are.

**What is the motivation / use case for changing the behavior?**

Leaking memory is bad, mkay?

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the latest stable and snapshot (http://docs.reactiveui.net/en/contributing/snapshot/index.html) versions.**

7.2

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

offan:
1) Why reverse navigation creates a new copy of the view that is in the ViewModelViewHost?
2) If the ViewModelViewHost has set the visibility property Collapsed and reverse navigation (HostScreen.Router.NavigateBack.Execute(new System.Reactive.Unit());) have a memory leak. because of (fe.WhenAnyValue(x => x.IsHitTestVisible).SkipWhile(x => !x)) not dispose.

ChrisPulman:
Should be fixed via #3105

PR #1336, opened 2017-04-24
Fix ViewModelViewHost memory leak

Silv3rcircl3:

Bug fix

The ViewModelViewHost leaks memory on XAML platforms.

The subscription to the SizeChangedEvent will be disposed via WhenActivated and it no longer leaks memory.

No

**Other information**:

### DIFF
--- a/src/ReactiveUI/Xaml/ViewModelViewHost.cs
+++ b/src/ReactiveUI/Xaml/ViewModelViewHost.cs
@@ -54,7 +54,7 @@ namespace ReactiveUI
 
         public string ViewContract
         {
-            get { return this.viewContract; }
+            get { return viewContract; }
             set { ViewContractObservable = Observable.Return(value); }
         }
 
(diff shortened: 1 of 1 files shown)


kentcb:
NOTE: there's a chance this fixes #1314 - will need to check.

src/ReactiveUI/Xaml/ViewModelViewHost.cs
```diff
@@ -101,21 +101,20 @@ public ViewModelViewHost()
                     }
 
                     var viewLocator = ViewLocator ?? ReactiveUI.ViewLocator.Current;
-                    var view = viewLocator.ResolveView(x.ViewModel, x.Contract) ?? viewLocator.ResolveView(x.ViewModel, null);
+                    var view = viewLocator.ResolveView(x.ViewModel, x.Contract) ?? viewLocator.ResolveView(x.ViewModel);
```
ghuntley: Hey Marc (@Silv3rcircl3) good catch on the subscription not being disposed. Can you share with me why this line was changed though?

src/ReactiveUI/Xaml/ViewModelViewHost.cs
```diff
@@ -101,21 +101,20 @@ public ViewModelViewHost()
...
 
                     if (view == null) {
-                        throw new Exception(String.Format("Couldn't find view for '{0}'.", x.ViewModel));
+                        throw new Exception($"Couldn't find view for '{x.ViewModel}'.");
```
ghuntley: I'm okay with refactoring away from `String.Format` into the C#6 shorthand. 👍

src/ReactiveUI/Xaml/ViewModelViewHost.cs
```diff
@@ -101,21 +101,20 @@ public ViewModelViewHost()
...
                     view.ViewModel = x.ViewModel;
                     Content = view;
                 }));
-            });
 
-            this
-                .WhenAnyObservable(x => x.ViewContractObservable)
-                .ObserveOn(RxApp.MainThreadScheduler)
-                .Subscribe(x => this.viewContract = x);
+                d(this.WhenAnyObservable(x => x.ViewContractObservable)
```
ghuntley: Hey Marc (@Silv3rcircl3) good catch on the subscription not being disposed. 👍

src/ReactiveUI/Xaml/ViewModelViewHost.cs
Silv3rcircl3: @ghuntley the second parameter is optional and therefore always null, so it could be safely removed. It was my R# alt + return muscle memory that forces me to "fix" it ;-) same as with "this" and the string.Format, but if you prefer to keep it I'm fine with it too.

njannink:
When is this change planned to be integrated? I was doing memory profiling of my application and I'm having memory leaks because of this. Is there otherwise a workaround how to get this fixed?

Silv3rcircl3:
@njannink I simply copied the source code in my project with the fix and use my local VMVH. There is another issue with this one, it's calling WhenActivated a second time when you replace the ViewMode with a new one. I still need to create another PR for this, you can fix it by simply using ContenControl instead of TransitionContentControl as base class.

Issue #1345, opened 2017-05-16
Removing a Range of items from a ReactiveList causes a crash with ReactiveCollectionViewSource  on iOS. Might be expected behavior


PureWeen:
**Do you want to request a *feature* or report a *bug*?** 
bug

**What is the current behavior?**
Removing a range from a ReactiveList using RemoveRange causes the ReactiveCollectionViewSource to remove items in such a way that it will cause an NSInternalInconsistencyException 
Reason: attempt to delete item 5 from section 0 which only contains 4 items before the update 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

I found this nifty scenario thing @kentcb  did
http://cl.ly/270m0d1q2N2U

And plugged a scenario into it

```
using System;
using System.Reactive.Concurrency;
using ReactiveUI;

namespace RxUIReactiveTableViewSourceRepro
{
    public class Scenario18ViewModel : ScenarioViewModelBase
    {
        public override void Start()
        {
            RxApp.MainThreadScheduler.Schedule<string>(
                null,
                TimeSpan.FromSeconds(1),
                (s, st) =>
                {
                    this.allItems.Add(new ItemViewModel("Item 1") { IsActive = true });
                    this.allItems.Add(new ItemViewModel("Item 2") { IsActive = true });
                    this.allItems.Add(new ItemViewModel("Item 3") { IsActive = true });
                    this.allItems.Add(new ItemViewModel("Item 4") { IsActive = true });
                });

            RxApp.MainThreadScheduler.Schedule<string>(
             null,
             TimeSpan.FromSeconds(5),
             (s, st) =>
             {

                 allItems.RemoveRange(1, 3);
             });
        }

        public override string Name => "A Batch Delete of things";

        public override string Description
            => "Items deleted off tail end of code";
    }
}

```

This will cause a crash that says
```
Foundation.MonoTouchException: Objective-C exception thrown.  Name: NSInternalInconsistencyException Reason: attempt to delete item 5 from section 0 which only contains 4 items before the update
```

From what I can tell

The issue happens here
```
var normalizedUpdates = IndexNormalizer.Normalize(updates);
```

The updates go in as indexes 1,2,3 but they come out as 1,3,5

I understand what the IndexNormalizer is doing and why. Especially for cases where you do something like
```
RemoveAt(4);
RemoveAt(0);
RemoveAt(9);
```

[...]

PR #1346, opened 2017-05-17
Changed it to add pending changes all at the same starting index when…

PureWeen:
… using RemoveRange

Bug Fix

https://github.com/reactiveui/ReactiveUI/issues/1345

 

It shouldn't 

I tested a few variations of removing items and the only one that reaches the code I added is RemoveRange which appears to be fixed by the code I added

I also tested the scenarios in @kentcb 's application he wrote for testing list scenarios and it all seemed happy.

- [X ] The commit follows our guidelines: https://github.com/reactiveui/reactiveui#contribute

**Other information**:


src/ReactiveUI/Cocoa/CommonReactiveSource.cs
```diff
@@ -330,7 +330,19 @@ private void SubscribeToSectionInfoChanges(int sectionInfoId, IReadOnlyList<TSec
...
+
+                                    if (y.Change.OldItems?.Count > 1 && y.Change.Action == NotifyCollectionChangedAction.Remove) 
+                                    { 
+                                        foreach(var item in y.Change.OldItems) {
+
+                                            var args = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Remove, item, y.Change.OldStartingIndex);
+                                            this.pendingChanges.Add(Tuple.Create(y.Section, new PendingChange(args)));
+                                        }
+                                    } 
+                                    else {
```
kentcb: Nitpick: brace on same line here when others aren't.

src/ReactiveUI/Cocoa/CommonReactiveSource.cs
```diff
@@ -330,7 +330,19 @@ private void SubscribeToSectionInfoChanges(int sectionInfoId, IReadOnlyList<TSec
                                             });
                                     }
 
-                                    this.pendingChanges.Add(Tuple.Create(y.Section, new PendingChange(y.Change)));
```
kentcb: The related issue describes this problem as being to do with index normalization. Did that not turn out to be the case then? If not, should the responsibility be moved there? Is this a platform-specific thing, or iOS only?

kentcb: Whoops, I meant platform-agnostic, or iOS only?

PureWeen: hehe yea I wrote and re-wrote that ticket probably 5 times as my understanding of what was happening kept deepening so the language might not be quite as accurate.  

The same RemoveRange code for me works fine on Android. On Android I'm using a "ReactiveListAdapter" which does really nothing that exciting compared to  CommonReactiveSource.  I find that iOS always requires the more "exciting" solutions :-/ 

The issue is very specific to how iOS does the batch updates and the magic involved with making sure list changes propagate correctly to an iOS batch update

This comment summarizes it all very well 
https://github.com/reactiveui/ReactiveUI/blob/b418df64918da0c78a378af1b846afbab0f4e5ae/src/ReactiveUI/Cocoa/IndexNormalizer.cs#L68

As far as where the logic should live I think you're right. I followed the code further down the rabbit hole and now have the fix summed up with a very small change.

The way I did it at first mainly spawned from  wanting to affect as little as possible. Making changes further down into the core made me nervous that there were other permutations that I'm not quite aware of that I would be breaking but being cautious probably had the opposite result.  For example the if else I did I'm pretty sure was ultimately pointless and could have been summed up with a single line but I was just being cautious :-)

Ultimately the new solution I've checked in seems to be address the problem more at the core.

src/ReactiveUI/Cocoa/CommonReactiveSource.cs
PureWeen: Not applicable anymore but  I didn't notice I did that... I'm doing more right now on VS for Mac since the latest VS 17  / 15.2 is pretty messed up and VS for Mac really likes fighting me it seems when I'm trying to format my code.

PR #1348, opened 2017-05-18
Allow ToProperty to generate Lazy properties or other types by making…

bradphelan:

Feature

   `IObservable<T>` when used with `ToProperty` must target a property of type T

`IObservable<T>` when used with `ToProperty` can be used with a property of any type U. The inner type of the observable is decoupled from the type of the targeted property.

The primary motivation is allow the use of

IObservable<Lazy<T>> to target properties of type T

the added test case demonstrates

        class LZPClass : ReactiveObject
        {
            public LZPClass()
            {
                Lazy<int> init = new Lazy<int>(()=>25);
                _Value = Observable.Return( new Lazy<int>( () => 20 ) ).ToProperty( this, p=>p.Value, initialValue: init);
            }
            readonly ObservableAsPropertyHelper<Lazy<int>> _Value;
            public int Value => _Value.Value.Value;
        }
        [Fact]
        public void ToPropertyShouldWorkWithLazyProperties()
        {

            var lzp = new LZPClass();
            Assert.Equal( lzp.Value, 20 );
        }

Note that the with _ToProperty_ the user has always been responsible for writing 

     _Helper.Value

Now the user can write

    _Helper.Value.blah.blah.foo.bar

if they wish as long as the type of the value returned by bar matches the property type.

No. I've added an extra type parameter but this should be automatically inferred given current usage. Only if users have explicitly defined the type params will there be a compile failure. This should be quite rare. 

- [ x] The commit follows our guidelines: https://github.com/reactiveui/reactiveui#contribute
- [ x] Tests for the changes have been added (for bug fixes / features)
- [ x] Docs have been added / updated (for bug fixes / features)

**Other information**:


src/ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -229,11 +229,11 @@ public static class OAPHCreationHelperMixin
         /// An initialized ObservableAsPropertyHelper; use this as the backing field 
         /// for your property.
         /// </returns>
-        public static ObservableAsPropertyHelper<TRet> ToProperty<TObj, TRet>(
-            this IObservable<TRet> This,
+        public static ObservableAsPropertyHelper<TObs> ToProperty<TObj, TProp, TObs>(
```
olevett: I think it would be worth adding XMLDOC comments for the type params now they do more.

src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
```diff
@@ -229,6 +229,26 @@ public void ToPropertyShouldFireBothChangingAndChanged()
...
+
+            private readonly ObservableAsPropertyHelper<Lazy<int>> value;
+
+            public int Value => value.Value.Value;
+        }
+
+        [Fact]
+        public void ToPropertyShouldWorkWithLazyProperties()
+        {
+            Assert.Equal( new LazyPropertyFixture().Value, 20 );
```
olevett: I think this should be `Assert.Equal(expected, actual);` not that it's likely to make much difference...

src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
```diff
@@ -229,6 +229,26 @@ public void ToPropertyShouldFireBothChangingAndChanged()
             Assert.Equal("Baz", resultChanged[1].Value);
         }
 
+        private class LazyPropertyFixture : ReactiveObject
+        {
+            public LazyPropertyFixture()
+            {
+                value = Observable
+                    .Return(new Lazy<int>(() => 20 ))
+                    .ToProperty(this, p=>p.Value, new Lazy<int>(()=>25));
```
olevett: Given the `Observable.Return` will pipe straight through in this case, should the `initialValue` be used at all? Would it be worth adding a test case around `initialValue` being used first or do the other test cases cover it off (in which case is it worth dropping it here to reduce noise).

bradphelan:
Can somebody either merge this or reject it please.

bradphelan:
Can somebody either merge this or reject it please.

bradphelan:
Can somebody either merge this or reject it please.

kentcb:
@bradphelan  I guess I'm not understanding the motivation here...perhaps you can elaborate.

OAPH can be inherently lazy by defering the subscription, so what value is there in adding a `Lazy<T>` wrapper on top? To take your cited example, one could write:

```C#
    class XYZClass : ReactiveObject
    {
        public XYZClass()
        {
            _Value = Observable
                .Return(20)
                .ToProperty(this, p=>p.Value, initialValue: 25, deferSubscription: true);
        }
        readonly ObservableAsPropertyHelper<int> _Value;
        public int Value => _Value.Value;
    }
```

bradphelan:
I believe that Lazy was just an example. I think, from memory iwas
transporting.

IObservble<LazyTask<T>>

LazyTask is an application specific class we have that is Task like but
doesn't start till Value is called where I believe

LazyTask<T>.Value -> Task<T>

In this case it is was important not to defer the subscription but only
start the task running when asked for it.

My memory is a bit hazy on the issue but it was something along those lines.

kentcb:
> In this case it is was important not to defer the subscription but only
start the task running when asked for it.

But that's exactly what deferral achieves...

Apologies but my vote is to reject this change because I see no tangible benefit (you might convince me otherwise with a compelling use case or two), so the added complexity does not justify the change. It seems to me that the stated use cases are better served with Rx constructs.

bradphelan:
No you are missing the point. It is an IObservable<Lazy<T>> not IObservable<T>. I am not caring about deferring subscription to the observable but deferring evaluation of the lazy unless the property is accessed. For example over a period of 100 seconds the IObservable<Lazy<T>> might deliver 100 Lazy<T> objects. However if the property  is only accessed10 times then only 10 of the 100 Lazy<T> objects will get evaluated thus saving 90 evaluations.

If I take your solution and flatten IObservable<Lazy<T>> to IObservable<T> with deferred subscription then after the first time subscription all further Lazy<T> objects will be evaluated.

kentcb:
If the property changes 100 times (be it lazy or not), why would only 10 of those changes be required/observed? Typically reactive properties are bound, so all changes would be observed...

bradphelan:
As you said yourself. ``They are typically bound``. Which means not always. There is nothing in the ReactiveUI framework that says you **must** bind a property to XAML. 

Off the top of my head. 

- A command that when activated accesses a property on the viewmodel. That property is the result of subscribing to ``IObservable<Lazy<T>>`` and only when the button is clicked.
- A property that is bound and unbound. ie it is only bound when a certain sub part of the UI is shown, say like a popup. No need to evaluate the expensive function when the popup is not shown and no need to create a completely new viewmodel to abstract over this problem. Just hide the ``Lazy<T>`` behind the property access.

kentcb:
Thanks Brad.

> As you said yourself. They are typically bound

Poor choice of wording on my behalf. What I meant was typically they're bound, but if not then they're part of some other pipeline (like a `WhenAnyValue`). After all, if the property isn't being used reactively, why is it being defined reactively?

> A command that when activated accesses a property on the viewmodel. That property is the result of subscribing to IObservable<Lazy<T>> and only when the button is clicked.

In that case, the property would be better designed as a method that is invoked by the command (or is just moved directly into the command execution logic). What is the value of the property in such a scenario if it is updated reactively (lazy or not) but its value is only used at specific points in time (imperatively)?

> A property that is bound and unbound.

In this case, one can simply ref-count the source observable. When there are no subscribers (i.e. the popup disappears), it detaches from the source. Then it reattaches when a subscriber is re-established (the popup appears again).

bradphelan:
``Lazy<T>`` a  It is not the same as ``Func<T>`` or a method call.  If a lazy is asked once for a value it calculates. If it is asked a second time it returns the cached value. ``Func<T>`` will always calculate the value.  If I have a stream of ``Lazy<T>`` and bind it to a property ``t`` of type ``T`` then  everytime ``t`` is accessed it returns the freshest value but the freshest value is only calculated at the last  possible moment. This is perfect for a command. If the command is executed very fast then it grabs cached values. If the command is executed very slow then it skips evaluating ``Lazy<T>``.

> In this case, one can simply ref-count the source observable. When there are no subscribers (i.e. the popup disappears), it detaches from the source. Then it reattaches when a subscriber is re-established (the popup appears again).

ReactiveUI is about making hard things easy. You are describing workarounds that make easy things hard. I have a simple property that implements caching and cache clearing and you want to replace this with UI lifetime tracking and ref counted observables. Neither of these are impossible but they are fiddly.

Of course I have my own implementation of ToProperty  that does what I need. If you don't want to merge this so other users can take advantage of a more abstract interface then it's ok with me. I'll keep on using my implementation.

PR #1362, opened 2017-05-25
docs: fix formatting and improve API docs for XF's VMVH

kentcb:

Chore

Just improving formatting and API docs for the XF `ViewModelViewHost`, in preparation for some actual code tweaks.


src/ReactiveUI.XamForms/ViewModelViewHost.cs
```diff
@@ -70,7 +89,9 @@ public ViewModelViewHost()
                 (vm, contract) => new { ViewModel = vm, Contract = contract, });
 
             var platform = Locator.Current.GetService<IPlatformOperations>();
-            if (platform == null) {
+
+            if (platform == null)
+            {
```
olevett: Is there an extra newline here? (I think my rebracer at least is pro `{` on the same line)

kentcb: Good catch. My Rebracer was playing up. I've just restarted VS and re-applied, then pushed another commit.

PR #1363, opened 2017-05-25
chore: don't publish ReactiveUI-Core as package no longer exists

olevett:
Fix ish. Remove the "Core" publish target, which was removed during the merge.

It's possible that the target should actually be kept, in which case we need to reinstate it in nuget/csproj generation.

**Other information**:

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -56,7 +56,7 @@ var buildVersion = gitVersion.FullBuildMetaData;
 // Artifacts
 var artifactDirectory = "./artifacts/";
 var testCoverageOutputFile = artifactDirectory + "OpenCover.xml";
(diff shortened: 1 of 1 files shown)


ghuntley:
nah; wrong approach. I want to keep the core package. Working on packaging now. Check Slack for some more insights.

olevett:
More slack discussion on this around https://reactivex.slack.com/archives/C02AJB872/p1495795159523437

ghuntley:
Had a change of heart on this one.

a) using `reactiveui` means the NuGet count on the main package is higher. (i.e. marketing)
b) couldn't find a compelling reason to continue using a metapackage.
c) undoing this won't be so hard for users in the future if this turns out to be bad idea (tm)

Issue #1354, opened 2017-05-23
ReactiveList.AddRange throws "Range actions are not supported" given a list of length 2<=n<=10, under certain circumstances


domyd:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
`ReactiveList.AddRange` and `ReactiveList.InsertRange` throw a `System.NotSupportedException` "Range actions are not supported." when passing them a list with at least 2, but at most 10 number of elements, under certain circumstances that are still not entirely clear to me (see below examples).

`RxApp.SupportsRangeNotifications` is `true`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
The "cleanest" yet oddest example I came up with was this specimen:

### ViewModel
```csharp
public class MainViewModel : ReactiveObject
{
    public MainViewModel()
    {
        // This works for any number of elements
        SearchResults.AddRange(Create(5));

        Observable.Start(() => { })
            .Do(_ =>
            {
                // This also works for any number of elements
                SearchResults.AddRange(Create(5));
            })
            .ObserveOn(RxApp.MainThreadScheduler)
            .Subscribe(_ =>
            {
                // This fails for [2..10] elements
                SearchResults.AddRange(Create(5));
            });

        IEnumerable<string> Create(int numElements)
            => Enumerable.Range(1, numElements).Select(i => $"item_{i}");
    }

    public ReactiveList<string> SearchResults { get; } = new ReactiveList<string>();
}
```
### View
```xaml
<Grid>
    <ListView ItemsSource="{Binding SearchResults}"></ListView>
</Grid>
```

---

... which manifests itself in a more realistic scenario, like the following:

### ViewModel
```csharp
public class MainViewModel : ReactiveObject
{
    public MainViewModel()
    {
        // This works for any number of elements
        SearchResults.AddRange(Create(5));

        Search = ReactiveCommand.Create(() => { });
        Search.Subscribe(_ =>
        {
            // This fails for [2..10] elements
            SearchResults.AddRange(Create(5));
        });
    }

    private static IEnumerable<string> Create(int numElements)
        => Enumerable.Range(1, numElements).Select(i => $"item_{i}");

    public ReactiveCommand<Unit, Unit> Search { get; }

    public ReactiveList<string> SearchResults { get; } = new ReactiveList<string>();
}
```
### View
```xaml
<Grid>
    <Grid.RowDefinitions>
[...]

PR #1366, opened 2017-05-25
fix: check for unsupported range actions in ReactiveList on WPF

domyd:
Bug fix

#1354 

It really shouldn't (I'm like 99.9% sure). There shouldn't be any new code paths where `NotifyCollectionChangedEventArgs` events that were previously never sent, would now be sent (on `net45` platforms).

**Other information**:
So I've thought of a couple ways to go about this.

One would be to simply set `RxApp.SupportsRangeNotifications` to `false` on the `net45` platform. In that case it would add elements one-by-one if the diff list didn't meet the reset threshold. It would be a simpler change, but there would be a performance penalty.

Then there's the way I did it ... what I don't like about it is that it's sort of duplicating the logic from the WPF validation method, however since that class apparently hasn't changed in 13 years, and probably never will again, it seems *okay* to me. I've thought a little about catching a `NotSupportedException` in the `ReplaceIfUnsupported` method to get rid of the duplicated logic, but that doesn't seem worth it for the performance loss.

Are there any other simpler/more reliable ways to do this that I may have missed?

Regarding the test cases, I'm really not sure whether these should go inside `ReactiveCollectionTest.cs` or `PropertyBindingTest.cs` as they kind of belong to both; for now I've put them in the former.


domyd:
Any feedback on this?

ElijahReva:
@domyd  Hi, could you give me access to your fork? I'll update PR.

domyd:
@ElijahReva Yeah, I've added you as a collaborator on my fork (I hope that gives you push access to all branches?).

glennawatson:
Given now that @ElijahReva has updated the PR to pass tests, can we start the review process on this one?

glennawatson:
Given now that @ElijahReva has updated the PR to pass tests, can we start the review process on this one?

src/ReactiveUI/ReactiveList.cs
```diff
@@ -29,19 +29,31 @@ public class ReactiveList<T> : IReactiveList<T>, IReadOnlyReactiveList<T>, IList
...
+            // certain circumstances, so we catch those here first
+            this.CollectionChanged?.Invoke(this, replaceIfUnsupported(args));
+        }
+        
+        NotifyCollectionChangedEventArgs replaceIfUnsupported(NotifyCollectionChangedEventArgs args)
+        {
+            // see System.Windows.Data.ListCollectionView.ValidateCollectionChangedEventArgs
+            switch (args.Action) {
+                case NotifyCollectionChangedAction.Add when args.NewItems.Count != 1:
+                case NotifyCollectionChangedAction.Remove when args.OldItems.Count != 1:
```
glennawatson: Do we want to do this for all platforms?

domyd: I don't think so. WPF/.NET Framework seems to be the only target platform that is picky about these conditions ([src](https://referencesource.microsoft.com/#PresentationFramework/src/Framework/System/Windows/Data/ListCollectionView.cs,0c35e5d5d6bb7c2f)). But I assume that, since this piece of code is under a `#if NET_461` guard, it will get included only there (at least that was my thinking a year ago).

glennawatson: Makes sense.

src/ReactiveUI.Tests/ReactiveCollectionTest.cs
```diff
@@ -984,6 +984,82 @@ public void DerviedCollectionShouldHandleItemsRemoved()
...
+        [WpfFact]
+        public void DataboundReactiveListDoesNotThrowForInsertRange()
+        {
+            var vm = new PropertyBindViewModel();
+            var view = new PropertyBindView { ViewModel = vm };
+            var fixture = new PropertyBinderImplementation();
+            fixture.OneWayBind(vm, view, m => m.SomeCollectionOfStrings, v => v.FakeItemsControl.ItemsSource);
+            vm.SomeCollectionOfStrings.ResetChangeThreshold = int.MinValue;
+
+            foreach (var item in Create(5))
```
olevett: I'm not entirely convinced that Create is clearer than an inline Range here

src/ReactiveUI.Tests/ReactiveCollectionTest.cs
```diff
@@ -984,6 +984,82 @@ public void DerviedCollectionShouldHandleItemsRemoved()
...
+            {
+                vm.SomeCollectionOfStrings.Add(item);
+            }
+
+            var exception = Record.Exception(() => {
```
olevett: Given our expectation is no exception, do we gain anything by trapping and asserting later (wouldn't the test just throw anyway?)

glennawatson:
@olevett Hey Olly if you don't mind can you check to make sure the unit tests match what you're after?

olevett:
@domyd / @glennawatson have changed the base branch to master, hopefully tests all pass and then will merge in.

PR #1370, opened 2017-05-27
Target netstandard1.1 for ReactiveUI.Testing

### LINKED ISSUE #1340: Add netstandard target for ReactiveUI.Testing
Following on from #1332, we should change the target for ReactiveUI.Testing to include NetStandard as Microsoft.Reactive.Testing now supports this target.

kentcb:

Fixes #1340.

ReactiveUI.Testing targets a whole bunch of platforms.

There is no platform-specific code, so just target netstandard 1.1.

No.

**Other information**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -237,5 +237,7 @@ tools/
 
 # ReactiveUI
 artifacts/
-src/CommonAssemblyInfo.cs
 src/ReactiveUI.Events/Events_*.cs
+
+# macOS
+.DS_Store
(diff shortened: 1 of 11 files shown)


kentcb:
@olevett good point - I hadn't noticed that `Microsoft.Reactive.Testing` is now netstandard 1.3. I've updated that.

Then I found that the **ReactiveUI.Tests** project was failing to build because it can't reference **ReactiveUI.Testing** anymore due to it being netstandard. I believe this means it needs to target net46 minimum, but when I try that, I run into a whole different problem. All tests are failing with:

```
Message: System.IO.FileLoadException : Could not load file or assembly 'System.Reactive.PlatformServices, Version=3.0.1000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```

_Ugh_, this tooling will be the death of me...

olevett:
I think there's likely to be a point on that where we need to upgrade xunit as well, and that breaks some of the threading stuff that is depended on... Gotta love change!

PR #1371, opened 2017-05-27
housekeeping: only include events source if present

### LINKED ISSUE #1369: Events don't build out of the box
***Note*: for support questions, please ask on StackOverflow: https://stackoverflow.com/questions/tagged/reactiveui . This repository's issues are reserved for feature requests and bug reports.**

Bug...ish

Grabbing the source from github and building fails with:

> 1>CSC : error CS2001: Source file 'C:\Users\Kent\Repository\ReactiveUI\src\ReactiveUI.Events\Events_NET45.cs' could not be found.

1. Clone the repo
2. Open in VS2017 and build

Should be able to build a fresh clone.

Lower the barrier to entry, reduce friction.

Latest source code.

The problem is that the source code is missing:

image

That makes sense, given that it's generated as part of the build. However, the project file is set up to assume that the code is there:

```XML
  <ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
    <Compile Include="Events_NET45.cs" />
     ....
  </ItemGroup>
```

Perhaps a better approach would be to include a condition that the file exists? Will need to play around to know for sure...

kentcb:

chore. Fixes #1369.

Events project doesn't build out of the box due to missing generated source files.

Source files are only included in the project if they're present, meaning it won't fail due to a missing file.

No.

**Other information**:


ghuntley:
Not a fan of this immediately. Need to think about it some more.

olevett:
I think this has reared it's head now because Events wasn't part of the main RxUI solution before, so we could revert to that state and have a separate solution?

Alternatively, would it be better to have a pre-build step that calls through to the EventBuilder somehow?

PR #1380, opened 2017-06-03
chore: update to xunit 2

### LINKED ISSUE #1301: Update xunit nuget packages
***Note*: for support questions, please ask on StackOverflow: https://stackoverflow.com/questions/tagged/reactiveui . This repository's issues are reserved for feature requests and bug reports.**

Neither really

Old xunit version

Newer xunit version

XUnit 2 has a couple of features that would be helpful for some scenarios, in particular data driven tests via Theories (this would be useful in covering the wide variety of tests in ViewLocatorTests for example)

olevett:
Fixes #1301

Uses xunit 1.9.x

Use xunit 2.x

Unlikely, this shouldn't be being consumed by end users anyway.

**Other information**:
The biggest change that impacts this is the threading model for xunit 2 has changed a bit, so some tests are marked as STA or WPF which ensures they're running on the right thread.

I'm a little concerned the tests could be a bit flaky because of the changes, so one thing I'm keen for this PR to do is build it on a different machine.

If tests are flaky, we might need to either rework them a bit, possibly doing one of the following
- explicitly set a thread/Rx dispatcher
- change the way RxApp handles unit test runners
- use async/await (:( I know...) because that makes xunit play nice

### DIFF
--- a/src/ReactiveUI.Tests/MessageBusTest.cs
+++ b/src/ReactiveUI.Tests/MessageBusTest.cs
@@ -2,21 +2,14 @@
 // The .NET Foundation licenses this file to you under the MS-PL license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Reactive.Testing;
+using ReactiveUI.Testing;
+using System;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
-using Xunit;
-using System;
-using System.Linq;
-using System.Collections.Generic;
-using ReactiveUI;
-using System.IO;
-using System.Text;
-using ReactiveUI.Testing;
-using ReactiveUI.Tests;
 using System.Threading;
-
-using Microsoft.Reactive.Testing;
 using System.Threading.Tasks;
+using Xunit;
 
 namespace ReactiveUI.Tests
 {
@@ -121,18 +114,18 @@ namespace ReactiveUI.Tests
         }
 
         [Fact]
-        public void MessageBusThreadingTest()
+        public async Task MessageBusThreadingTest()
         {
             var mb = new MessageBus();
             int? listenedThread = null;
             int? otherThread = null;
             int thisThread = Thread.CurrentThread.ManagedThreadId;
 
-            Task.Run(() => {
+            await Task.Run(() => {
                 otherThread = Thread.CurrentThread.ManagedThreadId;
                 mb.Listen<int>().Subscribe(_ => listenedThread = Thread.CurrentThread.ManagedThreadId);
                 mb.SendMessage<int>(42);
-            }).Wait();
+            });
 
             Assert.NotEqual(listenedThread.Value, thisThread);
             Assert.Equal(listenedThread.Value, otherThread.Value);
--- a/src/ReactiveUI.Tests/MessageBusTest.cs
+++ b/src/ReactiveUI.Tests/MessageBusTest.cs
@@ -5,10 +5,13 @@
 using Microsoft.Reactive.Testing;
 using ReactiveUI.Testing;
 using System;
+using System.Reactive.Concurrency;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Reactive.Testing;
+using ReactiveUI.Testing;
 using Xunit;
 
 namespace ReactiveUI.Tests
@@ -18,17 +21,17 @@ namespace ReactiveUI.Tests
         [Fact]
         public void MessageBusSmokeTest()
         {
-            var input = new[] {1, 2, 3, 4};
+            var input = new[] { 1, 2, 3, 4 };
 
             var result = (new TestScheduler()).With(sched => {
                 var source = new Subject<int>();
                 var fixture = new MessageBus();
 
                 fixture.RegisterMessageSource(source, "Test");
-                Assert.False(fixture.IsRegistered(typeof (int)));
-                Assert.False(fixture.IsRegistered(typeof (int), "Foo"));
+                Assert.False(fixture.IsRegistered(typeof(int)));
+                Assert.False(fixture.IsRegistered(typeof(int), "Foo"));
 
-                var output = fixture.Listen<int>("Test").CreateCollection();
+                var output = fixture.Listen<int>("Test").CreateCollection(scheduler: ImmediateScheduler.Instance);
 
                 input.Run(source.OnNext);
 
@@ -39,20 +42,19 @@ namespace ReactiveUI.Tests
             input.AssertAreEqual(result);
         }
 
-
         [Fact]
-        public void ExplicitSendMessageShouldWorkEvenAfterRegisteringSource() 
+        public void ExplicitSendMessageShouldWorkEvenAfterRegisteringSource()
         {
             var fixture = new MessageBus();
             fixture.RegisterMessageSource(Observable<int>.Never);
-         
+
             bool messageReceived = false;
             fixture.Listen<int>().Subscribe(_ => messageReceived = true);
-         
+
             fixture.SendMessage(42);
             Assert.True(messageReceived);
         }
-     
+
         [Fact]
         public void ListeningBeforeRegisteringASourceShouldWork()
         {
--- a/src/ReactiveUI.Tests/Platforms/xaml/ActivationForViewFetcherTest.cs
+++ b/src/ReactiveUI.Tests/Platforms/xaml/ActivationForViewFetcherTest.cs
@@ -33,7 +33,7 @@ namespace ReactiveUI.Tests
             var activation = new ActivationForViewFetcher();
 
             var obs = activation.GetActivationForView(uc);
-            var activated = obs.CreateCollection();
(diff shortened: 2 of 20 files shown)


src/ReactiveUI.Tests/MessageBusTest.cs
```diff
@@ -117,18 +110,18 @@ public void RegisteringSecondMessageSourceShouldMergeBothSources()
...
-        public void MessageBusThreadingTest()
+        public async Task MessageBusThreadingTest()
         {
             var mb = new MessageBus();
             int? listenedThread = null;
             int? otherThread = null;
             int thisThread = Thread.CurrentThread.ManagedThreadId;
 
-            Task.Run(() => {
+            await Task.Run(() => {
```
olevett: This wasn't that stable locally, but doing async/await rather than `Wait()` seemed to make it happier

src/ReactiveUI.Tests/Platforms/xaml/ActivationForViewFetcherTest.cs
```diff
@@ -95,14 +98,14 @@ public void IsHitTestVisibleDeactivatesFrameworkElement()
...
-        [Fact]
+        [WpfFact]
         public void FrameworkElementIsActivatedAndDeactivatedWithHitTest()
         {
             var uc = new TestUserControl();
             var activation = new ActivationForViewFetcher();
 
             var obs = activation.GetActivationForView(uc);
-            var activated = obs.CreateCollection();
+            var activated = obs.CreateCollection(scheduler: ImmediateScheduler.Instance);
```
olevett: I get the feeling this shouldn't be necessary, but it looks like the way RxApp works, sometimes MainThreadScheduler might not be set as ImmediateScheduler in unit tests (we set _MainThreadScheduler but _UnitTestScheduler wins in general).

src/ReactiveUI.Tests/TestConfiguration.cs
```diff
@@ -0,0 +1,3 @@
+﻿using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true, MaxParallelThreads = 1)]
```
olevett: This is to make xunit2 behave more like xunit 1.

kentcb:
This LGTM @olevett!

Using Test Explorer, I can build and successfully run all tests (assuming I unload the Events* and AndroidSupport projects, since they fail to build on my machine).

What's interesting is that NCrunch seems to be working much better, too. I had to update the config (which I can PR separately once this is merged), but now only one test fails. And the test in question is `SchedulerShouldBeCurrentThreadInTestRunner`, which seems to be failing for the usual reasons of the scheduler ambient context being incredibly confusing and flaky code. It will be great to be able to use NCrunch instead of Test Explorer!

src/ReactiveUI.Tests/MessageBusTest.cs
kentcb: Yeah, calling `Wait` is a big no-no. See [here](https://twitter.com/bradwilson/status/828780630540382208).

src/ReactiveUI.Tests/Platforms/xaml/ActivationForViewFetcherTest.cs
kentcb: Yep, this is the exact problem I ran into with the `SchedulerShouldBeCurrentThreadInTestRunner` test under NCrunch. I dream of a day we can completely remove ambient contexts from ReactiveUI.

src/ReactiveUI.Tests/MessageBusTest.cs
olevett: Hmm, hadn't seen that, but does explain things! I think this was the only consistent deadlock, so hopefully we're not doing it elsewhere!

kentcb:
Ugh, I wanted to merge this, but build is failing with OpenCover error now. Am guessing it's an intermittent problem. I don't think there's an easy way for me to just force another build 🤔 Maybe @ghuntley can chime in?

ghuntley:
image

olevett:
That test seems to be the most flakey of the lot, but also, if it's not working possibly a lot of other tests won't behave quite right.

Will spend some more time digging at it, unless anyone has better ideas!

kentcb:
Wow, the `RxApp` code frigging sux. I can see what's happening here, but no idea how it ever worked, nor how to solve this problem for good. Here's the sequence of events:

1. `RxApp` static ctor is executing, which calls `Locator.CurrentMutable.InitializeReactiveUI()`
2. This results in platform registrations being registered, which in turn sets `RxApp.MainThreadScheduler` to `new WaitForDispatcherScheduler(() => DispatcherScheduler.Current)`
3. The `RxApp` static ctor continues executing and assigns `_MainThreadScheduler = CurrentThreadScheduler.Instance`
4. `RxApp.MainThreadScheduler` then proceeds to ignore that value because `__UnitTestMainThreadScheduler` is set to a different value

I'm 99% sure the bug is in the static ctor. Instead of assigning to `_MainThreadScheduler`, I suspect it should be assigning to the property itself. When I do that, the test in question passes, but now a bunch of other tests fail or, worse, stall completely. _sigh_

Will also dig a bit further...

olevett:
I'm wondering if part of this is due to the ThreadStatic `_UnitTestMainThreadScheduler`/`_UnitTestTaskpoolScheduler`. With these, order and threading matters. (e.g. if RxApp is initialized on one thread, but gets used on another, the value of MainThreadScheduler does matter).

I'm not entirely sure if the ThreadStatic stuff actually makes that much sense with schedulers as `TestUtils.WithScheduler` is gated to prevent multiple overrides of the scheduler (you could conceivably do this some other way which I guess makes any change here a breaking one anyway). The other cases it gets thrown around are quite difficult to trace and understand, but may be more helpful. Just removing the ThreadStatic stuff does however cause a whole load of tests, so maybe it is important...

olevett:
As a quick test of building etc, I've disabled the scheduler test that's failing - I don't think this is great (either we should fix it or 🔥 it) but I want to make sure CI can consistently build and run the other tests...

olevett:
As per suggestion from @kentcb I've been skipping most tests and then incrementally adding them back in. I've been using a quick extra build process to run tests 10 times locally which should hopefully catch out any intermittent failures, I've included this as cake task called `MinimalBuildAndTest`. I'd like to run this a much higher number of times before I'm completely happy, but tests so far don't seem to be crazily broken (I've made a couple of changes like explicit schedulers, but nothing enormous).

kentcb:
@olevett I had a crazy idea that I was gonna try, but haven't had the time...it may not work, but let me run it by you in case it helps.

I was thinking I'd add a console app to the solution, it would reference unit tests. When running the app it would:

* reflectively find all `Fact`s (think we can ignore theories because they complicate matters and not sure there even are any)
* randomize the order of the found tests
* run all found tests in the randomized order _n_ * _d_ times, where _n_ is some constant (let's say 100) and _d_ is the depth of recursion to be discussed shortly...
* if failure is detected, we have found a set of tests that fail when executed together (save this off in a variable for later reference)
* split the set of tests in two and recursively repeat the above process, increasing _d_ by 1

Thus we'd whittle down the tests to a very small set that fail when executed together. It's not perfect, I know, but it could be an automated way to whittle the situation down to just a handful of tests quite quickly.

Anyway, just an idea that may amount to nothing, but thought I'd pass it on since it's unlikely I'm going to find the time to do this myself.

olevett:
@kentcb tht sounds like a neat solution, at the moment I've got a cake build that does the iterations part (the easy bit!)

My one concern would be the test that causes issues might be a passing test, so only running failing tests might not pick it up correctly. Not sure if that’s a valid concern.

kentcb:
> My one concern would be the test that causes issues might be a passing test

Yeah, that's why I'm suggesting to find groups of tests where any test within the group fails (didn't word that very clearly initially). And the randomisation gives ample opportunity to find combinations of tests that might all pass individually, but not when executed together.

Ultimately, if this automation could narrow down to a small set of e.g. 10 or so tests (or even 2!) that cannot be executed together without occasional failures, would save a lot of your time.

src/ReactiveUI.Tests/TestUtilsTest.cs
```diff
@@ -7,10 +7,5 @@ namespace ReactiveUI.Tests
 {
     public class TestUtilsTest
     {
-        [Fact]
-        public async Task WithAsyncScheduler()
```
olevett: This test just seemed to break everything... I'm not even really sure what it's testing, so thought losing it seemed safest...

olevett: This test just seemed to break everything... I'm not even really sure what it's testing, so thought losing it seemed safest...

olevett:
I think this is getting pretty near to being reliable, if anyone's interested in the breakdown or a build script that thrashes the tests a few times to pick up threading quirks, there's a branch h[here](ttps://github.com/olevett/ReactiveUI/tree/%231301-xunit-update-breakdown) but I've squashed down to reduce the noise a bit.

olevett:
Rebased because file header changes confused git...

PR #1393, opened 2017-06-30
Splitting activation classes

olevett:
Splitting up a file into multiple files (on class boundaries)

Most activation logic in one class

Roughly class per file, which should make finding things a bit easier.

Shouldn't do. Namespaces etc are all the same.

**Other information**:
This is a stylistic thing that I think has come up before, but was really painful with individual csprojs explicitly including each file. This felt like an easy first place to change, and should hopefully reduce the maintenance burden.


olevett:
This needs a bit of work in terms of changes made for xunit 2, and things like file headers...

Issue #925, opened 2015-08-29
ReactiveUI .net 4.5 without WPF


grokys:
Hi, this is related to #734 but coming at it from a slightly different angle.

I'm currently working on https://github.com/grokys/Perspex/ which aims to be a cross-platform UI toolkit somewhat similar to WPF. I would like to be able to support ReactiveUI but its being tied to WPF poses problems with cross-platform usage, in that when run on mono ReactiveUI tries to load PresentationFramework which doesn't exist on that platform.

I realize that you probably wouldn't want to add support for such a fledgling project directly in ReactiveUI right now, but I was looking to at least find a way to work around the problem.

What would be the best thing to do? Is there a simple hack I can use to make ReactiveUI not try to load WPF on mono? Or should I create a perspex-specific fork of ReactiveUI for now?

I'd be interested to hear you thoughts.

SuperJMN:
Waiting for this to be fixed. :+1:

grokys:
I was wondering if a stub PresentationFramework.dll for mono might fix this at least in the short-term?

ghuntley:
'spoke with @grokys today. Avalonia currently has had to maintain their own fork because of this issue, they would _love_ to not have their own fork.

SuperJMN:
Hi guys! 

@ghuntley and I have been talking and he proposed the following:
"What if you just did a stub assembly. That way resolution works but stuff goes.boom if invoked.
If there's no wpf on the running then I don't see how it can go boom. this is using really silly assumptions right now btw.
It's more ramblings than a proposal. I don't know enough on wpf internals to give proper direction at this stage"

grokys:
Yep, I mentioned this above in a comment:

> I was wondering if a stub PresentationFramework.dll for mono might fix this at least in the short-term?

However it wasn't as easy as I anticipated. I can't remember the details but I seem to remember the stub had to go pretty deep.

petersondrew:
The issue I'm running up against is that referencing RxUI from our PCL libraries works fine, but our unit test libraries have to target NET45/Mono in order for things like Moq to be available. When we run our unit test libraries under mono, RxUI looks for PresentationFramework.dll which isn't available.

Any ideas on how to solve this? Could RxUI-Core be separated from the PresentationFoundation requirement?

petersondrew:
[...]

PR #1406, opened 2017-07-31
feature: move wpf out of main package to fix unit testing story on mono and unblock other ui frameworks

jkoritzinsky:
This change moves the WPF support out of the main package and into a new `reactiveui-wpf` package. This enables multiple avenues, including:
 * No need for a separate Mono build to support Mono
 * `reactiveui-winforms` has no dependency on WPF assemblies
 * UI Frameworks on Mono or .NET Framework can use ReactiveUI without having to maintain their own forks or custom builds that patch out WPF service hook-ins. (i.e. Avalonia).

#925. Currently, ReactiveUI assumes by default that an application referencing it on .net 4.5 wants to use WPF. Since WPF is not available on Mono, current ReactiveUI packages cannot be used on Mono platforms. Also, UI frameworks like Avalonia have to maintain their own forks of ReactiveUI to patch out code that might cause the runtime to load any WPF types.

The core ReactiveUI package can be used on the .NET Framework or Mono without requiring referencing WPF at all.

💥💥💥 

This PR is a breaking change and will require documentation in the RELEASENOTES to instruct users of WPF to install the `reactiveui-wpf` and `reactiveui-events-wpf` packages.  `reactiveui` package no longer contains any WPF references.

💥💥💥

- [ x ] The commit follows our guidelines: https://github.com/reactiveui/reactiveui#contribute

**Other information**:
I don't know where to update the docs. Please tell me where to look and I will update them as needed!


jkoritzinsky:
@ghuntley Can you review this?

src/ReactiveUI.Wpf/ReactiveUI.Wpf.csproj
```diff
@@ -0,0 +1,38 @@
...
+  <PropertyGroup>
+    <TargetFramework>net45</TargetFramework>
+    <DefineConstants>$(DefineConstants);WPF</DefineConstants>
+    <Description>WPF specific extensions to ReactiveUI</Description>
+    <PackageId>reactiveui-wpf</PackageId>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\ReactiveUI\CreatesCommandBinding.cs" Link="CreatesCommandBinding.cs" />
+    <Compile Include="..\ReactiveUI\Platforms\windows-common\ActivationForViewFetcher.cs" Link="windows-common/ActivationForViewFetcher.cs" />
```
olevett: Sad that this needs individual file references :(

olevett: This shows up with a little cross in VS for me (I don't think VS really likes it!) - long term, I'd be pro splitting the implementations (maybe sharing a base class for common functionality so we don't have the mess of #ifdefs) but for now, I think these might be neater just being included in the root of the project (the link should be fairly easy to spot, and the windows-common doesn't make a huge amount of difference within the context of the WPF project)

src/ReactiveUI/CreatesCommandBinding.cs
```diff
@@ -4,15 +4,20 @@
...
 
 #if NETFX_CORE
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Input;
+#elif WPF
+using System.Windows;
 #endif
 
+#if WPF
+namespace ReactiveUI.Wpf
```
olevett: I'd be tempted at this point to just pull the WPF copy into this project, and then remove the various `if`s from the shared copy. I think it's a bit neater than redefining the namespace

jkoritzinsky: The `RoutedEventArgs` type is in the WPF assemblies, so including the `Click` event as a default event won't work unless we use the `EventArgs` type.

ghuntley:
Couple changes were required:
* Each package needs to be white-listed in `build.cake` otherwise the version won't be pinned or published to MyGet (or NuGet)
* Also needed to split out of `reactiveui-events`

This PR is a breaking change and will require documentation in the RELEASENOTES to instruct users of WPF to install the `reactiveui-wpf` and `reactiveui-events-wpf` packages. The `reactiveui` package no longer contains any net45 assemblies.

ghuntley:
@jkoritzinsky thank-you very much for doing this. Fixes #925

PR #1418, opened 2017-08-08
Add netcoreapp build

### LINKED ISSUE #1417: .NET Core Compatible Build
Feature

Referencing ReactiveUI (v8 alpha) in .NET Core crashes in PlatformRegistrations since it is pulling in the .NET Standard build that throws the exception.

ReactiveUI references common services, or registers nothing at all without throwing an exception when referencing it in .NET Core.

No .NET Core apps or UI frameworks that run on .NET Core can use ReactiveUI (such as Avalonia).

All versions have this bug in their PCL versions, but it is only applicable to v8 since v8 is the only one that is referencable by a .NET Core app.

Either make the .NET Standard build not throw or add a .NET Core build that doesn't throw and keep the .NET Standard build throwing.

jkoritzinsky:
Support for a .NET Core non-crashing build.

Any .NET Core usage of ReactiveUI will crash.

A .NET Core build will not crash upon load. The user must supply a scheduler for the `RxApp.MainThreadScheduler` depending on their platform.

**What might this PR break?**
Nothing.

**Other information**:

Fixes #1417

### DIFF
--- a/src/ReactiveUI/Platforms/netcoreapp1.0/PlatformRegistrations.cs
+++ b/src/ReactiveUI/Platforms/netcoreapp1.0/PlatformRegistrations.cs
@@ -12,6 +12,7 @@ namespace ReactiveUI
         public void Register(Action<Func<object>, Type> registerFunction)
         {
             RxApp.TaskpoolScheduler = TaskPoolScheduler.Default;
+            RxApp.MainThreadScheduler = DefaultScheduler.Instance;
         }
     }
 }
--- a/src/ReactiveUI/Platforms/netcoreapp1.0/PlatformRegistrations.cs
+++ b/src/ReactiveUI/Platforms/netcoreapp1.0/PlatformRegistrations.cs
@@ -1,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Reactive.Concurrency;
+
+namespace ReactiveUI
+{
+    public class PlatformRegistrations : IWantsToRegisterStuff
+    {
+        public void Register(Action<Func<object>, Type> registerFunction)
+        {
+            RxApp.TaskpoolScheduler = TaskPoolScheduler.Default;
+        }
+    }
+}
--- a/src/ReactiveUI/ReactiveUI.csproj
+++ b/src/ReactiveUI/ReactiveUI.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.1;net45;uap10.0;Xamarin.iOS10;Xamarin.Mac20;MonoAndroid70</TargetFrameworks>
+    <TargetFrameworks>netstandard1.1;net45;uap10.0;Xamarin.iOS10;Xamarin.Mac20;MonoAndroid70;netcoreapp1.0</TargetFrameworks>
     <AssemblyName>ReactiveUI</AssemblyName>
     <RootNamespace>ReactiveUI</RootNamespace>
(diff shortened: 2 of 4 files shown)


jkoritzinsky:
@ghuntley Can you take a look when you have a chance?

src/ReactiveUI/Platforms/netcoreapp1.0/PlatformRegistrations.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+using System.Reactive.Concurrency;
+
+namespace ReactiveUI
+{
+    public class PlatformRegistrations : IWantsToRegisterStuff
+    {
+        public void Register(Action<Func<object>, Type> registerFunction)
+        {
+            RxApp.TaskpoolScheduler = TaskPoolScheduler.Default;
```
olevett: I'm a little wary of this for a couple of reasons
- I think TaskpoolScheduler should already be configured on non-portable platforms (which covers netcoreapp1.0) e.g. https://github.com/reactiveui/ReactiveUI/blob/develop/src/ReactiveUI/RxApp.cs#L38
- I think the MainThreadScheduler should also get set a default value (I think DefaultScheduler.Instance) if nothing else sets it and I think setting TaskpoolScheduler but not MainThreadScheduler is potentially a bit confusing.

I'd be interested to know if there are problems if this line is omitted, or if it's fine without. I think I'd keep the PlatformRegistrations class however as it adds some value on it's own.

I also think it might be worth wiring up an `IPlatformOperations` even if it just returns null, as I think there are cases it would cause a warning to be logged otherwise (@ghuntley may disagree though...)

src/ReactiveUI/ReactiveUI.csproj
```diff
@@ -51,7 +51,11 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'MonoAndroid70' ">
     <Compile Include="Platforms\android\**\*.cs" />
     <Compile Include="Platforms\xamarin-common\**\*.cs" />
-  </ItemGroup>  
+  </ItemGroup>
+  
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.0' ">
+    <Compile Include="Platforms\netcoreapp1.0\**\*.cs" />
```
olevett: Related to this, I'd add an additional `PropertyGroup` to `Directory.build.targets` for `netcoreapp`. I think it might make sense to define the `NETFX_CORE` constant there too, but can't remember what exactly that includes off the top of my head. Will double check when I get the chance.

jkoritzinsky: I'll add the `PropertyGroup`, but I can't add the `NETFX_CORE` define since that define currently backs all usings of the `Windows.UI.*` namespaces.

src/ReactiveUI/Platforms/netcoreapp1.0/PlatformRegistrations.cs
jkoritzinsky: * It looks like it is. Should I remove that setting from all of the platforms?
* I'll set it to `DefaultScheduler.Instance`

The only place that logs a warning about a missing `IPlatformOperations` looks to be in `RoutedViewHost`, which is only available in builds using the `Platforms/windows-common` files.

src/ReactiveUI/ReactiveUI.csproj
olevett: That seems fine then, might be we want to make NETFX_CORE two separate defines in future, but don't think it should be in scope for this change.

src/ReactiveUI/Platforms/netcoreapp1.0/PlatformRegistrations.cs
olevett: I think don't worry about removing it everywhere for now, I think separate issue (smaller scope = nicer to review :))

Don't worry about the PlatformOperations then, does look like we shouldn't be hitting it at all :)

PR #1426, opened 2017-08-14
housekeeping: added identity signing to NuGet packages

onovotny:
Adds code signing of the artifacts with the .NET Foundation certificate


ghuntley:
NOTE: This is _identity_ signing the assemblies so if you right -> click properties you can verify that the binaries were produced by us. This is **not strong naming** and is completely unrelated.

ghuntley:
Needed to change a couple things:
- Retargeted to master
- Failing to sign (due to environment variable missing) is now `throw`able and will fail a build
- Signing is not run on pull-request builds as the variable isn't set (for security reasons)
- Renamed some variables/locations for 💄 reasons

Apart from that thank-you once again Oren.

PR #1437, opened 2017-08-25
Changed to add pending changes all at the same starting index when using RemoveRange

PureWeen:
Changed to add pending changes all at the same starting index when using RemoveRange

Bug Fix

https://github.com/reactiveui/ReactiveUI/issues/1345

 

It shouldn't 

I tested a few variations of removing items and the only one that reaches the code I added is RemoveRange which appears to be fixed by the code I added

I also tested the scenarios in @kentcb 's application he wrote for testing list scenarios and it all seemed happy.

- [X ] The commit follows our guidelines: https://github.com/reactiveui/reactiveui#contribute

**Other information**:


src/ReactiveUI/Platforms/ios/CommonReactiveSource.cs
```diff
@@ -410,9 +410,13 @@ private static IEnumerable<Update> GetUpdatesForEvent(PendingChange pendingChang
                         .Range(pendingChange.NewStartingIndex, pendingChange.NewItems == null ? 1 : pendingChange.NewItems.Count)
                         .Select(x => Update.CreateAdd(x));
                 case NotifyCollectionChangedAction.Remove:
+                    //Use OldStartingIndex for each "Update.Index" because the batch update processes and removes items sequentially
```
olevett: Might be worth a space between "//" and the comment here

src/ReactiveUI/Platforms/ios/CommonReactiveSource.cs
```diff
@@ -410,9 +410,13 @@ private static IEnumerable<Update> GetUpdatesForEvent(PendingChange pendingChang
...
+                    //opposed to as one Range operation. 
+                    //For example if we are removing the items from indexes 1 to 5. 
+                    //When item at index 1 is removed item at index 2 is now at index 1 and so on down the line.
                     return Enumerable
                         .Range(pendingChange.OldStartingIndex, pendingChange.OldItems == null ? 1 : pendingChange.OldItems.Count)
-                        .Select(x => Update.CreateDelete(x));
+                        .Select(x => Update.CreateDelete(pendingChange.OldStartingIndex));
```
olevett: Do we have any test covering this? Would be worth having one if possible -> if we do, that might also be a better place for the above comment to live.

PureWeen: Possibly.. When I first wrote this there weren't any iOS specific tests but I see now there are. I built a repro project  https://github.com/PureWeen/RxUIReactiveTableViewSourceRepro.git  so would just need to integrate that concept into the tests.  I'll poke around next week see what I can do.

olevett: If tests are a pain (I don't think we've got a concrete runner) don't worry too much. In future, I'm keen to get proper tests running on platforms, but that's a fairly long term aim.

src/ReactiveUI/Platforms/ios/CommonReactiveSource.cs
PureWeen: I added these but it doesn't seem to be propagating here...

There's an "Update branch" button that I click and it says

"Update branch attempt failed
Head branch was modified. Review and try the merge again."

PR #1441, opened 2017-08-29
core: make ViewContractAttribute.Contract public

### LINKED ISSUE #1440: Make ViewContractAttribute.Contract public
The `ViewContractAttribute.Contract` property is currently internal. This makes perfect sense because it is only used in the `IMutableDependencyResolver` extension method.

However, we would also use this attribute to support another type of container (Autofac). We use this code: https://github.com/alexeyzimarev/ReactiveUI.Autofac/pull/2/files#diff-79d85e72a88e19d0308087e4c1091914R29

When the `Contract` property is internal, we cannot access it. So the adapter is not working for RxUI 7 onwards, only for 6.5. I think there is no issue making `Contract` public again.

alexeyzimarev:
The ViewContractAttribute.Contract property is currently internal but to build a custom service locator, which supports view contract attribute, we need access to it.

### DIFF
--- a/src/ReactiveUI/ViewAttributes.cs
+++ b/src/ReactiveUI/ViewAttributes.cs
@@ -24,7 +24,7 @@ namespace ReactiveUI
             Contract = contract;
         }
 
-        internal string Contract { get; }
+        public string Contract { get; }
     }
 
     /// <summary>


ghuntley:
Resolves #1440.

Thanks for the PR @alexeyzimarev 👍

PR #1463, opened 2017-09-05
test: testing approval tests should fail

olevett:
This should fail as the API surface has changed


olevett:
Boom.

@ghuntley thanks for doing the hard work bits 👍

PR #1470, opened 2017-09-06
Remove build.sh

eatdrinksleepcode:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

Build fix

**What is the current behavior? (You can also link to an open issue here)**

```build.sh``` is not currently supported (see [#1465](https://github.com/reactiveui/ReactiveUI/pull/1465#issuecomment-327377549)):
> Currently, the only end-to-end way for all platforms to build ReactiveUI is on
Windows via build.cmd.

**Please check if the PR fulfills these requirements**
- [x] The commit follows our guidelines: https://github.com/reactiveui/reactiveui#contribute
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

PR #1472, opened 2017-09-06
test: use new threads explicitly in message bus test

olevett:
Changes a test that sometimes goes 💥 

MessageBus test sometimes doesn't do what's expected, because using tasks doesn't necessarily make any guarantees about threading,

Explicitly use a new thread, which should make sure things happen as we'd expected.

**What might this PR break?**
Nothing.

**Other information**:
Spotted in #1470 but I don't think it's a unique scenario.

### DIFF
--- a/src/ReactiveUI.Tests/MessageBusTest.cs
+++ b/src/ReactiveUI.Tests/MessageBusTest.cs
@@ -1,15 +1,12 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MS-PL license.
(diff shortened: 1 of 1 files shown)


eatdrinksleepcode:
Nit: looks like the byte order mark got removed from the first line? Not sure if this repo cares about such things.

Other than that, 👍

olevett:
Am going to assume BOM isn't too important right now, as it's not consistent across the code base anyway,

ghuntley:
@eatdrinksleepcode come join us in Slack. We would love to have you. Https://ReactiveUI.net/slack

ghuntley:
@eatdrinksleepcode removing BOM is okay and the way, it's remenants before we had .editorconfig/started efforts to standardise.

PR #1477, opened 2017-09-07
Made command parameter expression binding no longer require ViewModel to exist at time of Binding

glennawatson:
Made command parameter expression binding no longer require ViewModel to exist at time of Binding

It changes it so that the view model is no longer required when you pass in a expression for the Parameter on the View Model for BindCommand()

This relates to bug #1164 

If you have a null ViewModel at time of registration (eg your control is created but the view model is set later on) it will crash due to the viewModel.WhenAnyValue() causing a Null Exception

It uses the ViewModel property expression binding instead which binds the the View's VM rather than the directly to the VM. This uses the existing Reflection.Rewrite and Reflection.ViewModelWhenAnyValue to generate the observable to the parameter binding.  

**What might this PR break?**
Potentially the parameter passing to the command. 

**Other information**:
Tests added to check for delayed VM setting.


src/ReactiveUI.Tests/CommandBindingTests.cs
```diff
@@ -170,15 +170,20 @@ public class CustomClickButton : Button
             this.CustomClick?.Invoke(this, EventArgs.Empty);
     }
 
-    public class CommandBindView : IViewFor<CommandBindViewModel>
+    public class CommandBindView : ReactiveObject, IViewFor<CommandBindViewModel>
```
olevett: I think this should be an additional test VM (`ReactiveObjectCommandBindView`?). 

This is assuming it's a backward compatible change, e.g. ViewModels that do exist ahead of time on POCO views should still work.

glennawatson:
Sure will make the changes tonight for you.. That's odd I'm not getting notifications about the change requests.

src/ReactiveUI.Tests/CommandBindingTests.cs
glennawatson: Done.

PR #1484, opened 2017-09-16
Add warning about disposing command-to-event bindings when deactiving…

jonstodle:
XML docs update

The XML docs for the `toEvent` parameter of `BindCommand()` states

> If specified, bind to the specific event instead of the default.

The XML docs now states

> If specified, bind to the specific event instead of the default.
> NOTE: If this parameter is used inside WhenActivated, it's important to dispose the binding when the view is deactivated.

The additional notice is to make it clear that if the `toEvent` parameter is used it's important to dispose the subscription. If this is not done, it will lead to multiple subscriptions to the event, which will (most likely) lead to unwanted behavior.

Feedback on the added text is very welcome.

I'm also making a PR to the docs to mention this behavior.

**What might this PR break?**
Shouldn't break anything

**Other information**:

### DIFF
--- a/src/ReactiveUI/CommandBinding.cs
+++ b/src/ReactiveUI/CommandBinding.cs
@@ -41,7 +41,9 @@ namespace ReactiveUI
         /// <param name="withParameter">The ViewModel property to pass as the
         /// param of the ICommand</param>
         /// <param name="toEvent">If specified, bind to the specific event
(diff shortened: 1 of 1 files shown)


olevett:
Seems like pretty reasonable extra text to me, we can always adjust in future 👍

PR #1506, opened 2017-10-06
docs: upgraded version of xam forms and changed ondestroy to use flush instead of shutdown

PureWeen:
Bug fix and upgraded xamarin forms version

Currently in OnDestroy on the MainActivity it was calling BlobCache.Shutdown which completely shuts down the cache from ever being used again. In cases where the MainActivity is destroyed but static things stay in memory this caused the app to stop working.

you could recreate by
- app started
- click back button to minimize app (this destroys the main activity)
- start app back up and nothing renders

```
 protected override void OnDestroy()
        {
            BlobCache.LocalMachine.Flush().Wait();
            base.OnDestroy();
        }
```

Here's a longer discussion about this
https://github.com/akavache/Akavache/issues/342

As an example it'd probably be good to just not use LocalMAchine all together but this works for now

**What might this PR break?**

**Other information**:


ghuntley:
Thank-you @PureWeen love your work!

Issue #1504, opened 2017-10-06
reactiveui-wpf.csproj triggers defect/crash in visual studio for windows when opening reactiveui.sln


ghuntley:
**Do you want to request a *feature* or report a *bug*?**

critical bug that completely blocks software development activities

**What is the current behavior?**

Visual Studio 2017 project system crashes when ReactiveUI.sln is opened

![error-message-after-visual-studio-crashes](https://user-images.githubusercontent.com/127353/31282003-f4650cea-aafc-11e7-9503-c96193cf60d8.png)

**Reproduction Steps*

- Clone https://github.com/reactiveui/reactiveui/tree/fix-vswin-crash
- File -> Open `ReactiveUI.sln`
- 💥 

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
# Diagnostic Logs
```
=====================
9/10/2017 11:33:47 PM
Recoverable
System.ArgumentException: 'dimensionValuesMap' must contain at least one element.
Parameter name: dimensionValuesMap
   at Microsoft.Requires.NotNullOrEmpty(IEnumerable values, String parameterName)
   at Microsoft.VisualStudio.ProjectSystem.ProjectConfigurationsServiceBase.CreateConfigurations(ImmutableArray`1 orderedDimensionNames, ImmutableDictionary`2 dimensionValuesMap)
   at Microsoft.VisualStudio.ProjectSystem.DeclaredDimensionsProjectConfigurationsService.<CalculateKnownProjectConfigurationsAsync>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.ProjectSystem.UnconfiguredProjectCache`1.<GetValueAsync>d__6.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.VisualStudio.ProjectSystem.ProjectConfigurationsServiceBase.<GetKnownProjectConfigurationsAsync>d__51.MoveNext()
[...]

PR #1507, opened 2017-10-09
[WIP] fix: reactiveui-wpf.csproj triggers defect/crash in visual studio for windows when opening reactiveui.sln

ghuntley:

bug

* File -> Open -> ReactiveUI.sln 💥 upon loading

error-message-after-visual-studio-crashes

See this entire thread https://twitter.com/onovotny/status/915908695015477248 and #1504 

# WARNING # 
This isn't fixed yet; CI will pass green and produce valid binaries but VSWIN if opened will crash.


ghuntley:
When the following workaround (remove + import target) is used

```
<PropertyGroup>
    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
</ProperyGrouip>
```

Visual Studio no longer crashes on startup but instead opens up a treasure trove of other problems


image

onovotny:
That looks like it's trying to reference the netstandard1.1 target in RxUI rather than the net45 one in there....weird.

@emgarten @rohit21agrawal @rrelyea @unniravindranathan any idea what's going on here?

ghuntley:
Additional information:

* Remove `ReactiveUI.WPF`
* Remove `ReactiveUI.Events.WPF`
* Remove `ReactiveUI.Testing` ('cause it deps on reactiveui-wpf)

Then everything is fine until you add a `File -> New Net461 Class Library` and then more 💥happens 

image

onovotny:
What if you update to the latest SourceLink. I think earlier versions brought in dependencies

onovotny:
that last bit is because of this https://github.com/reactiveui/ReactiveUI/blob/develop/src/Directory.build.targets#L25-L27

That needs to be in a Choose/When block since the legacy project system always evaluates items

onovotny:
Needs to be done like this: https://github.com/reactiveui/ReactiveUI/blob/develop/src/Directory.build.props#L25-L33

onovotny:
@ghuntley that one is failing because it's not marked as a legacy project in the props. You may not need the restoreprojectstyle.

Issue #1491, opened 2017-09-28
ObservableAsPropertyHelper needs to provide a way for users to determine if it's subscribed


glennawatson:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
If you deferSubscribe, there is no way to know at current if there is a subscription later on.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

**What is the expected behavior?**
To have a access method that allows access to the variable. 

**What is the motivation / use case for changing the behavior?**
It is costly for me with a large number of items (currently 200000+) to not use deferSubscription in my model/view model land to not to defer subscribe. I have a disposable in the OAPH and want to dispose if there has been a subscription only when doing a set operation. Having this accessor available will allow avoiding doing this work.

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the [latest stable](https://www.nuget.org/profiles/reactiveui) and [development snapshot](https://reactiveui.net/docs/getting-started/installation/nuget-packages/development-package-feed)**
All. I am using the latest 8.0 beta 89 for this bug.

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
```csharp
    /// <summary>
    /// Returns if the ObservableAsPropertyHelper is subscribing to the 
    /// base Observable stream.
    /// </summary>
    public bool HasBeenSubscribed()
    {
        return (Interlocked.Read(ref _activated) == 1);
    }
```

I am going todo a pull request tonight.

kentcb:
@glennawatson Just to be sure I understand the problem here: you have an `OAPH<IDisposable>` on your VMs and it is deferred. Upon cleaning up your VMs, you want to clean up that disposable resource _if it has been created_.

Is this correct?

If so, I agree in principal with your proposal. I _do_ think there may be more elegant ways to solve your problem, like a `DisposeMany` operator (as provided by DD I think):

```C#
this.someProperty = someObservable
    .Select(...)
    .DisposeMany()
    .ToProperty(..., defer: true);
```

[...]

PR #1512, opened 2017-10-11
feature: add IsSubscribed property to ObservableAsPropertyHelper<T>

glennawatson:
Introduces a IsSubscribed property to the OAPH. 

No property and therefore no easy way to know if a subscription (and therefore the value) has changed. This addresses #1491

A property in the OAPH

**What might this PR break?**
Shouldn't break any existing code.

**Other information**:

### DIFF
--- a/src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
+++ b/src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
@@ -152,7 +152,7 @@ namespace ReactiveUI.Tests
         }
 
         [Fact]
-        public void OAPHDeferSubscriptionParameterIsSubscribeIsNotTrueInitially()
+        public void OAPHDeferSubscriptionParameterIsSubscribedIsNotTrueInitially()
         {
             var observable = Observable.Create<int>(o => {
                 o.OnNext(42);
--- a/src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
+++ b/src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MS-PL license.
 // See the LICENSE file in the project root for more information.
 
@@ -149,6 +149,23 @@ namespace ReactiveUI.Tests
             Assert.False(isSubscribed);
             Assert.Equal(42, fixture.Value);
             Assert.True(isSubscribed);
+        }
+
+        [Fact]
+        public void OAPHDeferSubscriptionParameterIsSubscribeIsNotTrueInitially()
+        {
+            var observable = Observable.Create<int>(o => {
+                o.OnNext(42);
+                o.OnCompleted();
+
+                return Disposable.Empty;
+            });
+
+            var fixture = new ObservableAsPropertyHelper<int>(observable, _ => { }, 0, true);
+
+            Assert.False(fixture.IsSubscribed);
+            Assert.Equal(42, fixture.Value);
+            Assert.True(fixture.IsSubscribed);
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
```diff
@@ -152,6 +152,23 @@ public void OAPHDeferSubscriptionParameterDefersSubscriptionToSource()
         }
 
         [Fact]
+        public void OAPHDeferSubscriptionParameterIsSubscribeIsNotTrueInitially()
```
kentcb: minor typo here (missing 'd'), but I had to read this several times for it to click

src/ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -134,6 +134,20 @@ public T Value
         }
 
         /// <summary>
+        /// Gets a boolean value indicating whether the ObservableAsPropertyHelper
+        /// has subscribed to the source Observable.
+        /// Useful for scenarios where you use deferred subscription and want to know if 
+        /// the ObservableAsPropertyHelper Value has been accessed yet.
+        /// </summary>
+        public bool IsSubscribed
```
kentcb: I know we have no consistency in this regard, but should we just use C# 7 expression-bodied property here? 7 lines will reduce to 1. Per above, I would expect to see:
```C#
public bool IsSubscribed => _activated > 0;

File: src/ReactiveUI/ObservableAsPropertyHelper.cs [hunk 3]
```diff
@@ -26,7 +26,7 @@ public sealed class ObservableAsPropertyHelper<T> : IHandleObservableErrors, IDi
         T _lastValue;
         readonly IConnectableObservable<T> _source;
         IDisposable _inner;
-        private int _activated;
+        private long _activated;
```
kentcb: Is this necessary? I get why you've changed the type to `long` because `Interlocked.Read` requires it, but the reason there is no `int` overload is because a 32-bit read is guarateed to be atomic, no? And besides, does it really matter if `IsSubscribed` isn't 💯 bang up to date? Even with an `Interlocked` call, you'll still have a race condition if you do any meaningful work in response to `IsSubscribed` and multiple threads are in play...

At worst I would think `_activated` could just be marked as `volatile`...

src/ReactiveUI/ObservableAsPropertyHelper.cs
glennawatson: I think that's actually C# 6 :)

glennawatson: C# 7 would be 

public bool IsSubscribed 
{
  get => _activated > 0;
}

kentcb: Yeah, I meant 6 ;) They're all blending into one at this point 🙃

src/ReactiveUI/ObservableAsPropertyHelper.cs
glennawatson: Yeah I was debating that during the submission. Left the volatile out for the moment unless you can think of a good reason to add? As you said the C# standard does say int's are atomic read operations.

src/ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -26,7 +26,7 @@ public sealed class ObservableAsPropertyHelper<T> : IHandleObservableErrors, IDi
         T _lastValue;
         readonly IConnectableObservable<T> _source;
         IDisposable _inner;
-        private int _activated;
+        private  int _activated;
```
kentcb: Sorry, minor quibble but we've ended up with a superfluous space here.

src/ReactiveUI/ObservableAsPropertyHelper.cs
kentcb: I think _technically_ the use of `volatile` is required on some architectures to ensure that `_activated` is read from main memory rather than from a cache, and therefore being possibly stale. However, in practice I don't know what architectures these are and am not qualified to know the scenarios in which this would actually bite.

However, I do think the `IsSubscribed` property is merely indicative and not something people should be encouraged to use anyway, so I don't see `volatile` as necessary (but again, I could be easily swayed by an argument from those who know better).

kentcb:
Thanks again @glennawatson!

Issue #1508, opened 2017-10-09
If a method is obsolete then add [Obsolete] to the generated result


ghuntley:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

EventBuilder scans assemblies in a platform and generates out `observable.fromevent` for each eventhandler for that platform.

https://github.com/reactiveui/ReactiveUI/tree/develop/src/EventBuilder

![image](https://user-images.githubusercontent.com/127353/31337360-aeebc586-ad46-11e7-9695-01fbcbc08588.png)

Sometimes these events are obsolete (or soon will be). 

**What is the motivation / use case for changing the behavior?**

Attach `[Obsolete]` on the generated events so that consumers of ReactiveUI that use `Obsolete` events inherit the designs made by the downstream framework designers.

GKotfis:
I'll try deal with it.

GKotfis:
I have some solution but wondering if Message and Error flag from OsboleteAttribute should be also propagated?
```
public class Class1
{
    [Obsolete()]
    public event EventHandler SampleEvent;

    [Obsolete("With message")]
    public event EventHandler SampleEventWithObsoleteMessage;
}
```
**Result:**
```
        [Obsolete]
        public IObservable<System.EventArgs> SampleEvent {
            get { return Observable.FromEventPattern<System.EventHandler, System.EventArgs>(x => This.SampleEvent += x, x => This.SampleEvent -= x).Select(x => x.EventArgs); }
        }

        [Obsolete("With message")]
        public IObservable<System.EventArgs> SampleEventWithObsoleteMessage {
            get { return Observable.FromEventPattern<System.EventHandler, System.EventArgs>(x => This.SampleEventWithObsoleteMessage += x, x => This.SampleEventWithObsoleteMessage -= x).Select(x => x.EventArgs); }
        }
```

ghuntley:
Great idea! Any flags or messages should be passed back up.

ghuntley:
Fixed in #1514

PR #1514, opened 2017-10-11
feat: propagate obsolete attributes to generated observables

GKotfis:

feature

#1508 

If source event has [Obsolete] attribute generated event also will has it. Additional properties like Message, IsError are also passed.

**What might this PR break?**

nothing

**Other information**:
btw Visual Studio automatically format some code in `EventTemplateInformation.cs`. Some discussion already started about that at slack.

### DIFF
--- a/src/EventBuilder/Cecil/EventTemplateInformation.cs
+++ b/src/EventBuilder/Cecil/EventTemplateInformation.cs
@@ -88,8 +88,8 @@ namespace EventBuilder.Cecil
 
             return new ObsoleteEventInfo
             {
-                Message = obsoleteAttribute?.ConstructorArguments?.ElementAtOrDefault(0).Value?.ToString() ?? string.Empty,
-                IsError = bool.Parse(obsoleteAttribute?.ConstructorArguments?.ElementAtOrDefault(1).Value?.ToString() ?? bool.FalseString)
+                Message = obsoleteAttribute.ConstructorArguments?.ElementAtOrDefault(0).Value?.ToString() ?? string.Empty,
+                IsError = bool.Parse(obsoleteAttribute.ConstructorArguments?.ElementAtOrDefault(1).Value?.ToString() ?? bool.FalseString)
             };
         }
 
--- a/src/EventBuilder/Cecil/EventTemplateInformation.cs
+++ b/src/EventBuilder/Cecil/EventTemplateInformation.cs
@@ -40,12 +40,10 @@ namespace EventBuilder.Cecil
             var ret = RenameBogusWinRTTypes(param.ParameterType.FullName);
 
             var generic = ei.EventType as GenericInstanceType;
-            if (generic != null)
-            {
+            if (generic != null) {
                 foreach (
                     var kvp in
-                        type.GenericParameters.Zip(generic.GenericArguments, (name, actual) => new {name, actual}))
-                {
+                        type.GenericParameters.Zip(generic.GenericArguments, (name, actual) => new { name, actual })) {
                     var realType = GetRealTypeName(kvp.actual);
 
                     ret = ret.Replace(kvp.name.FullName, realType);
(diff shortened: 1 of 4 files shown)


src/EventBuilder/Cecil/EventTemplateInformation.cs
```diff
@@ -81,6 +79,20 @@ private static string GetRealTypeName(TypeReference t)
...
+        private static ObsoleteEventInfo GetObsoleteInfo(EventDefinition ei)
+        {
+            var obsoleteAttribute = ei.CustomAttributes.FirstOrDefault(attr => attr.AttributeType.FullName.Equals("System.ObsoleteAttribute"));
+
+            if (obsoleteAttribute == null) 
+                return null;
+
+            return new ObsoleteEventInfo
+            {
+                Message = obsoleteAttribute?.ConstructorArguments?.ElementAtOrDefault(0).Value?.ToString() ?? string.Empty,
```
kentcb: If `ElementAtOrDefault(0)` is indeed the default (`null`), won't this crash? Shouldn't it just be using `ElementAt`, since we'd expect that ctor argument to _always_ be present? Same goes for line below.

GKotfis: I tested it manually with prepared example library:
```csharp
 public class Class1
    {
        public event EventHandler SampleEvent;

        [Obsolete()]
        public event EventHandler SampleObsoleteEvent;

        [Obsolete("Some message")]
        public event EventHandler SampleObsoleteEventWithMessage;

        [Obsolete("Some message", true)]
        public event EventHandler SampleObsoleteEventWithMessageAndError;
    }
```
The results is:
```csharp
public IObservable<System.EventArgs> SampleEvent {
            get { return Observable.FromEventPattern<System.EventHandler, System.EventArgs>(x => This.SampleEvent += x, x => This.SampleEvent -= x).Select(x => x.EventArgs); }
        }


        [Obsolete]
        public IObservable<System.EventArgs> SampleObsoleteEvent {
            get { return Observable.FromEventPattern<System.EventHandler, System.EventArgs>(x => This.SampleObsoleteEvent += x, x => This.SampleObsoleteEvent -= x).Select(x => x.EventArgs); }
        }


        [Obsolete("Some message")]
        public IObservable<System.EventArgs> SampleObsoleteEventWithMessage {
            get { return Observable.FromEventPattern<System.EventHandler, System.EventArgs>(x => This.SampleObsoleteEventWithMessage += x, x => This.SampleObsoleteEventWithMessage -= x).Select(x => x.EventArgs); }
        }


        [Obsolete("Some message", true)]
        public IObservable<System.EventArgs> SampleObsoleteEventWithMessageAndError {
            get { return Observable.FromEventPattern<System.EventHandler, System.EventArgs>(x => This.SampleObsoleteEventWithMessageAndError += x, x => This.SampleObsoleteEventWithMessageAndError -= x).Select(x => x.EventArgs); }
        }
```

I intentionally used null-conditional operator to avoid such problems.

kentcb: It's not using the elvis operator though for the case I'm talking about:
```
ElementAtOrDefault(0).Value
```

GKotfis: `ElementAt()` throws exception in case element does not exist. You can use `[Obsolete]` attribute without any arguments. As I see we can remove one null check for `ConstructorArguments` - it's always initialized. So it'll be: `obsoleteAttribute.ConstructorArguments.ElementAtOrDefault()`

kentcb: I think you might be missing my point :)

I realize `ElementAt` throws when the element doesn't exist. But so does `ElementAtOrDefault(0).Value`. The difference is that `ElementAt` will throw with an `IndexOutOfRangeException` whereas the latter will throw a less useful `NullReferenceException`.

So my question is around how this code actually works when there is no element at index `0`, for example. I'm suggesting it will crash with NRE because `Value` is being dereferenced.

GKotfis: It works because `CustomAttributeArgument` is a struct (value type).

kentcb: Aha! Perfect - thanks, I obviously did not realise that. Sorry for wasting your time :)

GKotfis: No problem :)

kentcb:
Many thanks @GKotfis! A very useful feature.

ghuntley:
You legend @GKotfis  thanks for doing this.

Issue #1505, opened 2017-10-06
nameof() overload for ObservableAsPropertyHelper mixin. 


glennawatson:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

In the ObservableAsPropertyHelper MixIn's it currently has a option of only using Expression to get the property name we are interested in observing. In my own code I got a dramatic boost in speed by using the nameof() operator instead since you never really want to use a "sub-property". I want to add to the code base a overload that allows you to specify a string instead of the Expression. This was a massive speed boost in my Observation objects on a project I'm working on. Unlike in WhenAnyValue() where you can have sub-properties the motivation for using Expression (beyond being consistent with like the WhenAnyValue syntax) is not as strong. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

**What is the expected behavior?**
Allow a overload to a string which is the propertyName. This allows the use of nameof() to avoid the runtime hit of Expression.

**What is the motivation / use case for changing the behavior?**
Performance. 

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the [latest stable](https://www.nuget.org/profiles/reactiveui) and [development snapshot](https://reactiveui.net/docs/getting-started/installation/nuget-packages/development-package-feed)**
ReactiveUI 8

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
If people are happy to add a overload, I will provide a PR.

kentcb:
This sounds good @glennawatson. Send in the PR!

As you say, parity across all RxUI operators isn't currently possible. However, I think the main reason is the internal reliance on the expression, not because of property chains. After all, we could always have overloads that take a `string[]` instead of an expression, where each `string` represents one property in the chain. Sound crazy? Not if you've got a Fody addon that rewrites all expression-based invocations as `string`-based invocations, which is something I could imagine being useful for OAPH once you provide the impl.

PR #1522, opened 2017-10-15
feature: nameof() overload for ObservableAsPropertyHelper mixin

glennawatson:
…yHelper, which allows use of the nameof() operator for performance based applications

Address #1505 - It adds a string overload of the ToProperty operator that can be used with the nameof() operator or later a fody. This helps when you have a large number of objects referencing a OAPH object.

To use the Expression<Func<>> to interpolate the name of the OAPH.

Adds a additional mixin that allows the use of a string. 

**What might this PR break?**
Since it's just adding a new override it shouldn't break existing use of the Expression. If other people have provided similar overrides in their own code could potentially cause issues. Also if the Fody uses these overloads it's probably best to use the string property version for it. 

**Other information**:
See also https://github.com/reactiveui/website/pull/54

### DIFF
--- a/src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
+++ b/src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
@@ -230,6 +230,27 @@ namespace ReactiveUI.Tests
         {
             var fixture = new OaphNameOfTestFixture();
 
+            var changing = false;
+            var changed = false;
+
+            fixture.PropertyChanging += (sender, e) => changing = true;
+            fixture.PropertyChanged += (sender, e) => changed = true;
+
+            Assert.False(changing);
+            Assert.False(changed);
+
+            fixture.IsOnlyOneWord = "baz";
+
+            Assert.True(changing);
+            Assert.True(changed);
+        }
+
+        [Theory]
+        [InlineData(new string[] { "FooBar", "Bazz" }, new string[] { "Foo", "Baz" }, new string[] { "Bar", "azz" })]
+        public void ToProperty_NameOf_ValidValuesProduced(string[] testWords, string[] first3Letters, string[] last3Letters)
+        {
+            var fixture = new OaphNameOfTestFixture();
+
             var firstThreeChanging = fixture.ObservableForProperty(x => x.FirstThreeLettersOfOneWord, beforeChange: true).CreateCollection(scheduler: ImmediateScheduler.Instance);
             var lastThreeChanging = fixture.ObservableForProperty(x => x.LastThreeLettersOfOneWord, beforeChange: true).CreateCollection(scheduler: ImmediateScheduler.Instance);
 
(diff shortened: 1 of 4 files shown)


src/ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -210,6 +210,27 @@ public static class OAPHCreationHelperMixin
             return ret;
         }
 
+        static ObservableAsPropertyHelper<TRet> observableToProperty<TObj, TRet>(
```
kentcb: Formatting here seems quite a ways off when reviewing in GitHub - mixture of tabs and spaces?

src/ReactiveUI/ObservableAsPropertyHelper.cs
```diff
@@ -210,6 +210,27 @@ public static class OAPHCreationHelperMixin
...
+        TRet initialValue = default(TRet),
+        bool deferSubscription = false,
+        IScheduler scheduler = null)
+    where TObj : IReactiveObject
+        {
+            Contract.Requires(This != null);
+            Contract.Requires(observable != null);
+            Contract.Requires(property != null);
+
+            var ret = new ObservableAsPropertyHelper<TRet>(observable,
```
kentcb: Can we, like, have each arg on its own line? And return directly instead of via `ret`?

src/ReactiveUI.Tests/ObservableAsPropertyHelperTest.cs
```diff
@@ -248,6 +226,47 @@ public void ToPropertyShouldFireBothChangingAndChanged()
         }
 
         [Fact]
+        public void ToProperty_NameOf_ShouldFireBothChangingAndChanged()
+        {
+            var fixture = new OaphNameOfTestFixture();
+
+            // NB: This is a hack to connect up the OAPH
```
kentcb: Confused by this. OAPH should be subscribing and connected up by default, unless the subscription has been defered.

glennawatson: This is a copy paste from the other test nearby. If not needed, likely not needed in previously written test.

src/ReactiveUI/ObservableAsPropertyHelper.cs
glennawatson: This is consistent with the non-nameof versions, do you want those changed too?

kentcb: No, as mentioned in Slack - let's just keep this focused on your changes. Otherwise we'll end up down a rabbit hole.

kentcb:
👍 Approved. Please remind me @ghuntley - are we merging this into v8 or holding off?

ghuntley:
@kentcb this is 🚢 into v8 and new packages will be published in the next couple of minutes @glennawatson. Reason for this is that it is adding public API surface, glenn did the meaningful work for us and we know he'll stick around to help out as need be.  Rule of thumb for whether to hold off implementing something into v8 is "does this steal time from maintainers?" or "is the surface area of change too much when we there's already been so much change in this release?" (aka let's wait before optimizing existing surface area)

PR #1534, opened 2017-11-10
fix: several ReactiveCommand bugs

### LINKED ISSUE #1525: Disposing of an observable involving a ReactiveCommand will leave the ReactiveCommand with CanExecute set to false
Bug (possibly expected behavior)

If you dispose of the observable returned from
```
ReactiveCommand.Execute()
```

while it's executing, the command won't transition back to IsExecuting False.  Which means future setups with things like InvokeCommand are basically dead in the water.

```
   [Fact]
        public void IsExecutingIsFalseAfterCommandDisposal()
        {
            (new TestScheduler()).With(sched => {
                
                int executeCount = 0;
                var execute = Observables.Unit.Delay(TimeSpan.FromSeconds(1), sched);
                var fixture = ReactiveCommand.CreateFromObservable(() => {
                    executeCount++;
                    return execute;
                }, outputScheduler: sched);
                var isExecuting = fixture
                    .IsExecuting
                    .CreateCollection(scheduler: ImmediateScheduler.Instance);
                
                var source = new Subject<Unit>();

                var disp = 
                    source
                        .InvokeCommand(fixture);


                source.OnNext(Unit.Default);
                
                sched.AdvanceByMs(100);
                Assert.Equal(2, isExecuting.Count);
                Assert.False(isExecuting[0]);
                Assert.True(isExecuting[1]);
                Assert.Equal(1, executeCount);
                
                //Dispose of execute
                disp.Dispose();
                sched.AdvanceByMs(901);

                //isexecuting is now stuck set to true until next executed
                //Assert.Equal(3, isExecuting.Count);
               // Assert.False(isExecuting[2]);


                source.OnNext(Unit.Default);
                sched.AdvanceByMs(100);
                Assert.Equal(2, executeCount); //FAILS
                
            });
        }
```

For a command that's disposed of to transition IsExecuting back to false

A lot of setups for RxUI take the form

```
this.WhenActivated(disp
...
this.WhenAnyValue(x=> x)
.InvokeCommand(cmd)
.DisposeWith(disp)
```

If the cmd is in the middle of executing and you navigate away from the page then come back to the page then that setup is basically now just dead in the water as the cmd will just sit there with "IsExecuting == true" and "CanExecute == false"

I was able to surface this issue in the Cinephile sample application by adding a 10 second Delay to the ReactiveCommand used in that app

```
  (movies, genres) =>
                    {
                        return movies
                                .Results
                                .Select(movieDto => MapDtoToModel(genres, movieDto));
                    })
                    .Delay(TimeSpan.FromSeconds(10));
```

- I set the pagesize to like 5 so the command would trigger multiple times
- Scrolled down to trigger the command
- clicked on a movie before the command could finish
- Navigated back to main screen
- now when you scroll and it tries to Invoke "InvokeCommand(LoadMovies)" it is no longer able to invoke it

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the [latest stable](https://www.nuget.org/profiles/reactiveui) and [development snapshot](https://reactiveui.net/docs/getting-started/installation/nuget-packages/development-package-feed)**

I wrote the included unit test against the develop branch. I saw this behavior in 7.4.0 as well

Not sure how to fix off hand ... Not sure if it's just expected behavior

kentcb:

fixes

1. `InvokeCommand` relies on Rx's `WithLatestFrom` operator, which is...weird. It subscribes to first observable first, and then the second, which can result in missing values.
2. Multiple in-flight command executions are not supported. More specifically, a count of in-flight executions is not kept, so it behaves a little weird (saying it can execute when another execution may be in flight).
3. Disposing of an in-flight execution subscription is now handled correctly. Fixes #1525.

1. `InvokeCommand` relies on our own custom `WithLatestFromFixed` operator, which publishes the source observable so values aren't missed.
2. A count of executions is tracked so that `IsExecuting` and `CanExecute` report intuitive values.
3. Termination of execution observables is correctly handled, such that disposing an in-flight execution does not screw things up.

**What might this PR break?**

Nothing - it should fix several scenarios.

**Other information**:

### DIFF
--- a/src/ReactiveUI.Tests/ReactiveCommandTest.cs
+++ b/src/ReactiveUI.Tests/ReactiveCommandTest.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MS-PL license.
 // See the LICENSE file in the project root for more information.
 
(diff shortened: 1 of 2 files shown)


Qonstrukt:
I've merged this locally and can confirm this fixes the problem. 👍

kentcb:
I meant to link to the [related Rx issue](https://github.com/Reactive-Extensions/Rx.NET/issues/444).

kentcb:
Thanks for confirming @Qonstrukt! I've rebased the branch on top of your Xamarin.Android fixes and will hopefully get it merged a bit later today. Reviews welcome!

kentcb:
Thanks for the review @olevett

PR #1540, opened 2017-11-22
fix: activate iOS views synchronously

tgjones:

This is either a bug fix or enhancement, depending on how you look at it. It applies the equivalent activation fix for `UIView` subclasses, that #1182 already did for `UIViewController` subclasses.

#1182 changed `UIViewController` subclasses to use `ViewWillAppear` instead of `ViewDidAppear` for activation, such that activation occurs before the view becomes visible.

This PR does the equivalent thing for `UIView` subclasses.

`UIView` subclasses (correctly) override `WillMoveToSuperview`, which is called before the view becomes visible. However, within those overrides, they schedule an async callback on the UI thread like this:

``` csharp
public override void WillMoveToSuperview(UIView newsuper)
{
    base.WillMoveToSuperview(newsuper);
    RxApp.MainThreadScheduler.Schedule(() => (newsuper != null ? activated : deactivated).OnNext(Unit.Default));
}
```

Because `Schedule` schedules an async callback on the UI thread, the callback may (and in my testing, will) be called after the view becomes visible. This means that there's a visible flicker where you first see uninitialized views, and then they are populated with data.

The activation callbacks are now called synchronously from `UIView` subclasses, such that activation occurs before the view becomes visible. Instead of the code above, the `WillMoveToSuperview` overrides become simpler:

``` csharp
public override void WillMoveToSuperview(UIView newsuper)
{
    base.WillMoveToSuperview(newsuper);
    (newsuper != null ? activated : deactivated).OnNext(Unit.Default);
}
```

This matches the existing behaviour for [ReactiveCollectionReusableView](https://github.com/reactiveui/ReactiveUI/blob/develop/src/ReactiveUI/Platforms/ios/ReactiveCollectionReusableView.cs#L86) and [ReactiveCollectionView](https://github.com/reactiveui/ReactiveUI/blob/develop/src/ReactiveUI/Platforms/ios/ReactiveCollectionView.cs#L86), which already called the activation callback synchronously.

**What might this PR break?**

If anybody relies on the current behaviour of activation occurring after views become visible, this would be a breaking change. But that was almost certainly undesirable behaviour in the first place.

**Other information**:

### DIFF
--- a/src/ReactiveUI/Platforms/apple-common/ReactiveControl.cs
+++ b/src/ReactiveUI/Platforms/apple-common/ReactiveControl.cs
@@ -93,7 +93,7 @@ namespace ReactiveUI
 #else
             base.ViewWillMoveToSuperview(newsuper);
 #endif
-            RxApp.MainThreadScheduler.Schedule(() => (newsuper != null ? activated : deactivated).OnNext(Unit.Default));
+            (newsuper != null ? activated : deactivated).OnNext(Unit.Default);
         }
 
         void ICanForceManualActivation.Activate(bool activate)
--- a/src/ReactiveUI/Platforms/apple-common/ReactiveImageView.cs
+++ b/src/ReactiveUI/Platforms/apple-common/ReactiveImageView.cs
@@ -95,7 +95,7 @@ namespace ReactiveUI
 #else
             base.ViewWillMoveToSuperview(newsuper);
 #endif
-            RxApp.MainThreadScheduler.Schedule(() => (newsuper != null ? activated : deactivated).OnNext(Unit.Default));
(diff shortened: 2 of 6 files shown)


tgjones:
The AppVeyor build failure looks unrelated to my changes, but please let me know if I need to fix something.

Qonstrukt:
I actually ran into this (again) today, so I'm all for it!

kentcb:
Great - thanks @tgjones! And yeah, the build error is definitely unrelated :/

kentcb:
Build fixed in #1542, just waiting for merge before I can merge this.

tgjones:
Great, thanks for merging!

PR #1544, opened 2017-12-03
Fix Modal WPF dialog from Interaction handler

### LINKED ISSUE #1396: Modal WPF dialog from Interaction handler breaks ReactiveUI viewmodel
A bug.

Opening a dialog (using OpenDialog()) from an interaction handler breaks ReactiveUI in the opened dialog. In my demo, ToProperty stops applying updates to the ObservableAsProperty value after the first change until the dialog is closed. When Open() is used instead, everything works as expected.

If the Interaction is not used, and the dialog is opened from a WPF click event handler, everything works as expected. Maybe a threading issue?

1. Download, open and run the attached VS2017 solution: [SSCCE.zip](https://github.com/reactiveui/ReactiveUI/files/1130651/SSCCE.zip)
2. A dialog with 2 buttons appears. Click "Open as window" to see the correct, intended behavior. Note that the "username" textbox has a red border when the textbox is empty. When text is entered, the red border disappears. Also note that this window is not modal (you can still interact with the first window). Close the window.
3. Click "Open as dialog". Note that the dialog no longer works correctly. Entering text in the username field does not make the red border disappear. Note that this window is modal.

Opening a modal dialog using OpenDialog() should not break the viewmodel behavior.

It is broken. My modal dialog doesn't work.

I am running Windows 10 x64, and using Visual Studio 2017 for a WPF application.
I tested this with ReactiveUI v7.4.0 and v7.0.0, both showing identical behavior.
I tested this with .NET 4.7 and .NET 4.5.2, both showing identical behavior.

scemino:
Closes #1396

Bug

Opening a dialog (using OpenDialog()) from an interaction handler breaks ReactiveUI in the opened dialog.

Now the demo SSCCE.zip works.

**What might this PR break?**
I hope nothing...

**Other information**:
For this kind of bugs, I don't know how to create a unit test.

### DIFF
--- a/src/ReactiveUI/Interactions.cs
+++ b/src/ReactiveUI/Interactions.cs
@@ -169,8 +169,7 @@ namespace ReactiveUI
             }
 
             return RegisterHandler(interaction => {
-                handler(interaction);
-                return Observables.Unit;
+                return Observable.Start(() => handler(interaction), RxApp.MainThreadScheduler);
             });
         }
 


kentcb:
Thanks for the report @scemino.

Any chance you could attach a simple repro WPF project? I'm not convinced using `Observable.Start` is the right fix here. The docs for that method specifically state it is for *synchronous* handlers, which generally you only want to use in unit tests. In a WPF app I'd rather expect you to register an asynchronous handler.

scemino:
Yes you can test it with the sample provided by the issue #1396.

glennawatson:
Closing issue due to solution shown in #1396 using async.

PR #1569, opened 2018-02-17
fix: allow ANDROID_HOME to contain spaces when building

domyd:
Bug fix in build process.

MSBuild aborts if `/p:AndroidSdkDirectory` contains spaces (e.g. with `/p:AndroidSdkDirectory=C:\Program Files (x86)\Android\android-sdk`, which is where Xamarin installs it by default).

Added quotes to the Android SDK path passed to MSBuild.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -116,7 +116,7 @@ Task("BuildEventBuilder")
             ArgumentCustomization = args => args.Append("/bl:eventbuilder.binlog /m")
         }
         .SetConfiguration("Release")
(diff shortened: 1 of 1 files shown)


domyd:
```
Events_UWP.cs(10889,140): error CS1513: } expected [C:\projects\reactiveui\src\ReactiveUI.Events\ReactiveUI.Events.csproj]
```

That was rather weird ... I wonder what could have caused that? It worked the second time.

kentcb:
Thanks @domyd!

PR #1571, opened 2018-02-19
[DoNotMerge] ISubject --> INPC support

bradphelan:

This add ISubject combinators

* Switch
* Select
* CombineLatest

for making transformations on subjects composable. For example to make a system of configuration scaled units this unit test show how it could be done.

        [Fact]
        public async void SubjectObservableConstraintCombinatorsShouldWork()
        {
            const double mtoinches = 39.3701;
            const double mtoyard = 1.09361;

            IObserver<Option<Exception>> errors = Observer.Create<Option<Exception>>(e => { });

            // Imagine this to be a user preference somewhere. We
            // could grab this with an IOC discovery or whatever.
            var scale = new BehaviorSubject<double>(mtoinches);

            // This is our model. We always work in units of meters 
            // of course.
            var model = new BehaviorSubject<double>(1); // in meters

            // This is our view model. It should map to the model via
            // the user preference ``scale``.
            ISubject<double> view = model.CombineLatest(scale, Converter.Multiply, errors);

            // Check that the view is correct initially
            Assert.Equal(mtoinches, (await view.FirstAsync().ToTask()));
            view.OnNext(mtoinches * 3.0);
            Assert.Equal(3.0, (await model.FirstAsync().ToTask()));

            // Change the user preference to yards
            model.OnNext(1.0);
            scale.OnNext(mtoyard);

            // Check the view is now in yards
            Assert.Equal(mtoyard, (await view.FirstAsync().ToTask()));
            view.OnNext(mtoyard * 5.0);
            Assert.Equal(5.0, (await model.FirstAsync().ToTask()));
        }

As well the following helpers are available

**SubjectToPropertyHelper**

Extends the idea of ObservableToPropertyHelper except this allows RW properties to be
generated from subjects.

        public static SubjectAsPropertyHelper<TRet> ToReadWriteProperty<TObj, TRet>
            ( this ISubject<TRet>          subject
            , TObj                         This
            , Expression<Func<TObj, TRet>> property
            , TRet                         initialValue      = default(TRet)
            , bool                         deferSubscription = false
            , IScheduler                   scheduler         = null ) where TObj : IReactiveObject

and

**TwoWayBindTo**
Allows binding of subjects to INPC properties

        public static IDisposable TwoWayBindTo<T>
            ( this ISubject<T> modelSubject
            , ISubject<T> viewSubject
            , Func<T, bool> validateView = null
            , IScheduler scheduler = null
            , IObservable<Option<IObservable<Unit>>> updateViewOn = null
            , IObservable<Option<IObservable<Unit>>> updateModelOn = null
            )

and

**PropertySubject**
Allows generates of subjects from INPC properties

        /// <summary>
        /// Generate a subject from an INPC property
        /// </summary>
        public static ISubject<TValue> PropertySubject<TObject, TValue>
            ( this TObject This
            , Expression<Func<TObject, TValue>> property
            , bool debounce = true
            )
            where TObject : class

**What might this PR break?**

It only adds new features

- [x ] Tests for the changes have been added (for bug fixes / features)
Maybe coverage is not enough
No docs updated. This is really just a request for comment at the moment.


bradphelan:
It would be nice if there would be a comment, a rejection or something here. Makes me feel the project is dying.

kentcb:
Don't think I'm getting it. I'd appreciate a simple example of what is impossible/hard to achieve today without this infrastructure, and how this helps achieve it.

bradphelan:
It's composable two way binding binding much in the same way that RX is
compos able 1 way binding. The example is in the test case I think. You
want to have configurable two way binding between a viewmodel in display
units and a model in meters. The display units can change, the user input
may change or the model might change. Keeping it all consistent is fiddly.
The combinators make this easier.

It's certainly possible to do without much in the same way it's possible to
do RX with just callbacks.  Note that the two way binding methods in RXUI
are not really composable so doing clever stuff with converters that change
behaviour due to other inputs becomes tricky.

PR #1579, opened 2018-03-12
fix: use netstandard 2.0 compatible version of splat

ghuntley:

Fix

Uses Splat v2.0 which isn't compatible with netstandard2.0

Uses Splat v3.0 which is compatible with netstandard2.0

**What might this PR break?**

Folks who are using pre-release builds will be automatically upgraded to splat3.0 when they upgrade their binary dependancy if they are using Visual Studio for Windows and packageref. 

**Other information**:


src/ReactiveUI/ReactiveUI.csproj
```diff
@@ -57,7 +57,7 @@
   
   <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.0' ">
     <Compile Include="Platforms\netcoreapp1.0\**\*.cs" />
-    <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.1.1" />
+    <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.3.0" />
```
onovotny: This causes package downgrade warnings. Please don't change this for netcoreapp1.x.

olevett: It being 4.1.1 also causes issues with splat3 I think (https://ci.appveyor.com/project/dotnetfoundation/reactiveui/build/1.0.442#L37419)

onovotny: That's because splat is broken: 

https://github.com/reactiveui/splat/blob/3.0.0/src/Splat/Splat.csproj#L19

Should not be referencing 4.3 there either.

Best solution overall is to remove netcoreapp1.0 and netstandard1.x. min version should be net461 and netstandard2.0.

ghuntley:
The ReactiveUI targets need to also be bumped to netstandard20.


> C:\projects\reactiveui\src\ReactiveUI.Testing\ReactiveUI.Testing.csproj : error NU1202: Package Splat 4.0.0 is not compatible with netstandard1.1 (.NETStandard,Version=v1.1). Package Splat 4.0.0 supports:

onovotny:
Is there still a need for a netxoreapp version? What's there that's not in the netstandard version?

ghuntley:
netcore is used for unit testing, xamarin workbooks and some folks actually use us serverside. We will need it for anything related to blazor as well.

ghuntley:
Red build after adjusting targets for UAP. Not sure what's going on here yet.

> C:\Program Files\dotnet\sdk\2.1.200-preview-007517\Sdks\Microsoft.NET.Sdk\build\Microsoft.PackageDependencyResolution.targets(167,5): error : Assets file 'C:\code\reactive
ui\ReactiveUI\src\ReactiveUI.Testing\obj\project.assets.json' doesn't have a target for 'UAP,Version=v10.0.14393'. Ensure that restore has run and that you have included '
uap10.0.14393' in the TargetFrameworks for your project. [C:\code\reactiveui\ReactiveUI\src\ReactiveUI.Testing\ReactiveUI.Testing.csproj]

src/Directory.build.targets
```diff
@@ -1,11 +1,11 @@
...
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == 'net45'">
-    <DefineConstants>$(DefineConstants);NET_45;XAML</DefineConstants>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net461'">
+    <DefineConstants>$(DefineConstants);NET_461;XAML</DefineConstants>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == 'uap10.0'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'uap10.0.16299'">
     <DefineConstants>$(DefineConstants);NETFX_CORE;XAML;WINDOWS_UWP</DefineConstants>
     <TargetPlatformVersion>10.0.14393.0</TargetPlatformVersion>
```
olevett: We can lose these two I think

src/ReactiveUI.Winforms/Properties/AssemblyInfo.cs
```diff
@@ -7,4 +7,4 @@
 using System.Runtime.CompilerServices;
 using System.Windows;
 
-[assembly: InternalsVisibleTo("ReactiveUI.Tests_Net45")]
+[assembly: InternalsVisibleTo("ReactiveUI.Tests_Net461")]
```
olevett: I don't think this is actually used anywhere (other than in the API surface tests, which this *should* break)

src/ReactiveUI.Events/ReactiveUI.Events.csproj
```diff
@@ -14,12 +14,12 @@
...
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'uap10.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'uap10.0.16299' ">
     <Compile Include="Events_UWP.cs" Condition="Exists('Events_UWP.cs')" />
-    <SDKReference Include="WindowsDesktop, Version=10.0.14393.0">
+    <SDKReference Include="WindowsDesktop, Version=10.0.16299.0">
       <Name>Windows Desktop Extensions for the UWP</Name>
     </SDKReference>
-    <SDKReference Include="WindowsMobile, Version=10.0.14393.0">
+    <SDKReference Include="WindowsMobile, Version=10.0.16299.0">
```
onovotny: This can go away. There really is no more windows mobile anymore, sadly.

src/ReactiveUI.Wpf/CpsWorkaround.targets
```diff
@@ -0,0 +1,5 @@
+<Project>
```
onovotny: This file isn't needed, can be deleted

onovotny:
@ghuntley I get that, but you don't need a netcoreapp version for that. Put the logic from `platforms\netcoreapp2.0` into `platforms\netstandard2.0` and you'll be all set.

You should only need a netcoreapp version if there's types you need there that aren't in netstandard. Workbooks, Blazor, et al, will happily use the netstandard version.

src/ReactiveUI/Platforms/netstandard2.0/PlatformRegistrations.cs
```diff
@@ -3,14 +3,16 @@
...
 using System;
+using System.Reactive.Concurrency;
 
 namespace ReactiveUI
 {
     public class PlatformRegistrations : IWantsToRegisterStuff
     {
         public void Register(Action<Func<object>, Type> registerFunction)
         {
-            throw new Exception("You are referencing the Portable version of ReactiveUI in an App. Reference the platform-specific version.");
```
ghuntley: This is a big change that I'm not comfortable with. We used to throw if the end user didn't reference the right packages as part of bait and switch now we don't. Now it'll be harder for the end user and us to troubleshoot registrations as previously we had certainty that things would explode. Thoughts?

src/ReactiveUI/Platforms/netstandard1.1/PlatformRegistrations.cs
```diff
@@ -1,16 +0,0 @@
...
-
-using System;
-
-namespace ReactiveUI
-{
-    public class PlatformRegistrations : IWantsToRegisterStuff
-    {
-        public void Register(Action<Func<object>, Type> registerFunction)
-        {
-            throw new Exception("You are referencing the Portable version of ReactiveUI in an App. Reference the platform-specific version.");
```
ghuntley: This is a big change that I'm not comfortable with. We used to throw if the end user didn't reference the right packages as part of bait and switch but now we don't. Now it'll be harder for the end user and us to troubleshoot registrations as previously we had certainty that things would explode if the user did something wrong. Now it's silent. Thoughts?

ghuntley: This is a big change that I'm not comfortable with. We used to throw if the end user didn't reference the right packages as part of bait and switch but now we don't. Now it'll be harder for the end user and us to troubleshoot registrations as previously we had certainty that things would explode if the user did something wrong. Now it's silent. Thoughts?

olevett: I think I'm happy either way TBH.

From what @onovotny has said, PackageRefs should prevent the problem we used to have from happening (e.g. if you import RxUI in a shared project, and then use that in an app directly, you should get the platform specific version rather than the netstandard version). 

I think it's fairly minor either way, so if we're worried I think better to ship with the status quo and then we can merge netcoreapp and netstandard when we're happy the tooling will solve it.

onovotny: You do not need to reference the right packages anymore because NuGet is better. PackageRef does this always. The only people this bits are people who are using packages.config and mess up.

The Xamarin Templates, and all the other defaults everywhere, use PackageRef now.

Qonstrukt: The only situation where this might still go wrong is for VS4Mac users. As adding the first package to PackageRef manually is still needed for it to use PackageRefs. By default it still assumes packages.config. I hope this will be fixed in 7.5.

Even if you set up a new solution, it adds required packages in a packages.config file. Annoying...

PR #1602, opened 2018-05-01
feature: what if we were licensed under MIT?

### LINKED ISSUE #1580: Change to GPL compatible license?
I was wondering if there is any particular reason the Ms-PL was picked as the license. Would it be possible to change to a GPL-compatible permissive license like MIT, so unfortunate people who have been forced into the GPL license can use this awesome library?

ghuntley:

Fixes #1580 

I keep getting reports from folks who want to contribute but their employer has a restrictive list of pre-approved licenses and MSPL isn't on that list. 

What if we enabled those to help us? Much easier than making them go through internal corporate bs.

**What might this PR break?**

Feelings (tm). This PR is under review by the .NET Foundation legal team. If no issues, then we merge!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,55 +1,23 @@
-Microsoft Public License (Ms-PL)
+The MIT License (MIT)
 
-This license governs use of the accompanying software. If you use the software,
-you accept this license. If you do not accept the license, do not use the
-software.
+Copyright (c) .NET Foundation and Contributors
 
-1. Definitions
+All rights reserved.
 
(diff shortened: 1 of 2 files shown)


paulcbetts:
+1 from me, since DNF owns the copyright we can actually do this, and I didn't have any Secret VeryGood Reason to MSPL this

ghuntley:
I dedicate this PR to @glennawatson. Thanks for all your help mate.

olevett:
:)

Issue #1596, opened 2018-04-26
Stamp <Product>$(AssemblyName) ($(TargetFramework))</Product> into builds


ghuntley:
That way folks can right click on the assembly and know if they are looking at the bait or the switch. 

See https://github.com/xamarin/Essentials/blob/b85fdb2b1623e458037146e1ab4a4f33ed2a8978/Xamarin.Essentials/Xamarin.Essentials.csproj#L15 for instructions how to do it.

I recommend adding it to our `Directory.props` instead of to each csproj that way it's done once and automatically implemented for current and future csprojs.

glennawatson:
This was fixed by #1604

PR #1604, opened 2018-05-01
stamp ReactiveUI ($(TargetFramework)) into builds

cabauman:
Stamp <Product>$(AssemblyName) ($(TargetFramework))</Product> into builds

No stamp

Stamp <Product>$(AssemblyName) ($(TargetFramework))</Product> into builds

**What might this PR break?**
Nothing

**Other information**:
Issue: #1596


ghuntley:
I'm having a look at this. I think  AssemblyName is not being defined/resolved. Best way to troubleshoot this is to take the binlog and put it in http://msbuildlog.com/

ghuntley:
re: not pre-checking

It's all good!

re: wtf is going on

image

image

AssemblyName is set when ProductName is evaluated but it's coming out as null. Strange.

re: next steps

I'm parking this one and moving on. We'll put it in the next release. It's pretty low priority.

Asesjix:
In most cases the project name is the same as the assembly name (all of us), then $(MSBuildProjectName) will work.. I also do not understand why AssemblyName sometimes does not work.

But why not just write "ReactiveUI ($(TargetFramework))", the product stays the same or what do you think? ;-)

cabauman:
Yep, looks good on this end.
reactiveui stamp

olevett:
Thanks @cabauman

PR #1607, opened 2018-05-01
Added basic Bind support to macOS for NSControls

### LINKED ISSUE #1603: Bind is currently not working two-way on NSTextFields.
We are looking for someone to research what would be involved to improve this situation.

Deliverables:

* Research and come up with recommendations
* PR the main project
* Update the website documentation.
* Update our integration tests.

See https://github.com/reactiveui/ReactiveUI/pull/1601 and https://github.com/reactiveui/ReactiveUI/pull/1598#issuecomment-385487486

Qonstrukt:
This actually makes `WhenAny` work on the `*Value` properties of any `NSControl`. This enables stuff like `Bind` to work two-way.
This fixes #1603. 

Notification events from AppKit NSControl elements aren't observed.

Make macOS behave more like iOS in this regard.

**What might this PR break?**
Any existing use of `Bind` where other methods are being used in conjunction to create a sense of two-way binding.

**Other information**:
Some other platform registrations were added as well which enable some basic functionality that was platform-independent anyway. Like `NSDate` to `DateTime` conversion and a basic `ISuspensionDriver`.


src/ReactiveUI/Platforms/apple-common/ObservableForPropertyBase.cs
```diff
@@ -83,6 +88,7 @@ protected void Register(Type type, string property, int affinity, Func<NSObject,
             typeProperties[property] = info;
         }
 
+#if UIKIT
```
olevett: If this really is UIKit only, I'd be tempted to push it down into the UIKitObservableForProperty, though that may be a breaking change so not too worried.

Qonstrukt: Yeah I figured it would be more logical to have all these together rather than having to look through multiple files with equal functionality. But I get what you're saying.

PR #1613, opened 2018-05-05
Migrate to VSTS

ghuntley:

housekeeping

We use Cake+GitReleaseManager+GitVersion with AppVeyor to do CI/CD

All CD responsibilities have been 🔥 from Cake. We now use GitNerdVersion. We use VSTS for build and CI/CD.


ElijahReva:
Is it possible to make Cake skip build target if they are not installed in the environment? Downloading 25 Gb or build tools to be able to work with the solution may be not the easy way for the newcomers. Also, tests are using only net461 target, I guess due to WPF stuff, so this is not necessary to run the test suite.

ghuntley:
@ElijahReva can you please open a proposal over at http://github.com/reactiveui/rfcs - that's an aspect that needs to get better. It will need a lot of engineering effort tho. It's something I would like to see us do in time.

ghuntley:
@onovotny I'm happy to merge this when you are. Use gitcommits as release notes for now.

ElijahReva:
@ghuntley I could do this but I have not enough expertise in the cross-plat area. For example, my idea is just to use `netstandard2.0` for core library and move all plat-specific stuff to separate projects\nugets. In such case, when VS started, it will disable projects that have no appropriate targets installed and make the first-time contributions much easier. Is it possible? That's what expects? Should I RFC this?

Have no idea how it could be done using `Cake` other than creating conditional targets per `*.csproj` file.

PR #1625, opened 2018-05-09
housekeeping: pin xamarin.forms version used by eventbuilder

ghuntley:

dodgy workaround that has been documented in the maintainer guide

unpinned builds

pinned builds

**What might this PR break?**

If we forget to bump this when bumping reactiveui.xamforms and then the correct version of events won't be code generatoed.

**Other information**:


ElijahReva:
Could it be the result of MSBuild update or something changed in CI environment? I also wasn't able to build locally even after pulling this commit. Still getting same `Failed to resolve!!! netstandard`. What else should be in the environment?

PR #1620, opened 2018-05-08
fix: disable WeakEventManager for net461 build (#1617)

ElijahReva:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
Fix 

**What is the current behavior? (You can also link to an open issue here)**
https://github.com/reactiveui/ReactiveUI/issues/1617

**What is the new behavior (if this is a feature change)?**
Use NET_461 instead of NET_45 pre-processor directive to disable WeakEventManager for WPF platform

**What might this PR break?**
Nothing.

**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

Issue #1372, opened 2017-05-27
Proposal: move ReactiveList to Legacy and recommend DD instead


kentcb:
`ReactiveList<T>` is complicated, has loads of bugs, and no one really understands the code. Meanwhile, DynamicData does way more than `ReactiveList<T>` ever did.

For v8, I propose we move `ReactiveList<T>` to the `Legacy` namespace with a deprecation warning, then officially recommend DD as the way to work with collections in a reactive fashion.

Thoughts?

aalmada:
DynamicData uses Rx.NET 3.0. Is this already supported in RxUI?

modplug:
@aalmada Dynamic Data supports both Rx.NET 3.x and Rx.NET 2.x

kentcb:
Thinking about this a bit more after a rare good night's sleep, the blocker right now will be ReactiveUI's use of `ReactiveList` for the routing infrastructure. There may be one or two other places where it is used internally too.

So perhaps a more realistic goal here is to deprecate it when we revamp the routing story (v9/v10?).

olevett:
On the Rx.Net 3.x front (sort of irrelevant now) RxUI v8 is Rx.Net 3 currently.

ghuntley:
Not sure if adding deprecation attribute though - it will cause any projects which are configured to fail on warnings to break (this is a good configuration). I think moving `ReactiveList<T>` into the legacy namespace and promoting DD is the way to go.

GiusepeCasagrande:
> Not sure on adding the deprecation attribute though - it will cause any projects which are configured to fail on warnings to break
I would say that depends on how "hard" is to change from `ReactveList<>` to DD. If it's a matter of few changes, straight forward, I would be happy to see my CI broken with the update to v.8 (assuming that this would be made on v.8) and with a easy fix that would improve greatly my solution. A disclaimer: I haven't touched DD yet :)

jcmm33:
[...]

PR #1626, opened 2018-05-09
[WIP] housekeeping: resolve broken build due to NET45 compilation symbol

ghuntley:

bugfix

CI was offline and I didn't do a local build before merging https://github.com/reactiveui/ReactiveUI/pull/1620

This PR is to investigate what's needed to get green build and verify the API approval surface contract test that is currently failing. Maybe we need to bless the fixed surface. Maybe we need to rework #1620 

**What might this PR break?**

we are already broken


ghuntley:
image

I'm retiring for the night. @olevett any chance you could 👀 ?

olevett:
I think this is correct in being an API change, we're now using the net46 specific set of ReactiveList properties, rather than the incorrect ones before - probably we approved it too eagerly when introducing/bumping up net platform versions.

I've also renamed the "e" parameter to "args" which is more consistent, but means our approval tests one pick up this compile time issue again, I think that's better, and maybe we need some better tests on ReactiveList to check what's going on (though see #1372)

PR #1645, opened 2018-05-16
EditorConfig from style guide

Asesjix:
Feature

I am a new member of ReactiveUI and have my problems with it through the code. Why? Because I'm used to certain code styles from the .NET universe e.g. docs, samples and so on. In the code I noticed some places where e.g. properties small and members with or underlined and then again without and so on.
I think you all know what I mean ;-)

We're part of the .NET Foundation and so i think we should adopt the code style.
In addition, help the new devs to get into the code ;-)

**What might this PR break?**
N/A

**Other information**:
I would start to adapt the code according to the rules after which the pr was merged

Ref: https://github.com/dotnet/corefx/blob/master/.editorconfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,78 +3,154 @@
 # top-most EditorConfig file
 root = true
 
-# Don't use tabs for indentation.
+# Default settings:
+# A newline ending every file
+# Use 4 spaces as indentation
 [*]
-indent_style = space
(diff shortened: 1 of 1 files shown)


paulcbetts:
Please don't go into projects as your first PR to demand they change the coding style

Issue #1648, opened 2018-05-17
Winforms Designer generates automatically a ViewModel = null


Asesjix:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
If a UserControl placed in a Form or UserControl , the Designer generates automatically a ViewModel = null in .Designer.cs
The result, is the ViewModel is null in WhenActivated.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. CustomUserControl
2. CustomViewModel
3. Some WhenActivated stuff
4. CustomUserControl > Form1

**What is the expected behavior?**
The Designer should not generate a ViewModel = null in .Designer.cs

**What is the motivation / use case for changing the behavior?**
User experience

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the [latest stable](https://www.nuget.org/profiles/reactiveui) and [development snapshot](https://reactiveui.net/docs/getting-started/installation/nuget-packages/development-package-feed)**

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

Qonstrukt:
About which designer are we actually talking? The iOS designer? Wouldn’t that only define a property if you didn’t yourself in your main class file?

Asesjix:
Oh sorry, yes it's the WinForms Designer.
When a custom control is placed on a form, the designer generates the code for it in a separate file, usually the developer does not intervene here because it is not the codebehind for the developer.
https://github.com/reactiveui/ReactiveUI/blob/c4c48cf1ec92935a2cb77c04bc793fef7022c600/integrationtests/IntegrationTests.WinForms/MainForm.Designer.cs#L39
After this line, the designer generates:
this.loginControl1.ViewModel = null;

Qonstrukt:
Did you implement `IViewFor<TViewModel>` yourself on that control? Because then you could use the `DesignerSerializationVisibility` attribute on the ViewModel property perhaps to prevent it from being initialised in the designer file? I'm not familiar with WinForms, but that's what I found about the subject since it's not really RxUI specific.

Asesjix:
I know, thanks for the tip! I will try is later 👍

Asesjix:
That's it. Thank you! So that a user does not get the problem he must use the ReactiveUserControl or set the attribute itself.

Issue #1590, opened 2018-04-23
Winforms ViewModelControlHost Bug


ThrasherGX:
**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
When a viewmodel is hosted in the ViewModelControlHost , the ViewModelControlHost sets the Content property before it sets the ViewModel property, which results in the ViewModel property on the View being null during the WhenActivated block, as well as during the view's Load event.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Host any viewmodel within the ViewModelControlHost control, then attempt to directly access the ViewModel property during the View's WhenActivated block.  A null reference exception should result.

**What is the expected behavior?**
Based on the behavior of the WPF ViewModelViewHost, the ViewModel should be set before the view is added to the visual tree, which means that the ViewModel would be non null during the WhenActivated block.

**What is the motivation / use case for changing the behavior?**
Currently, any use of the ViewModel during the WhenActivated block requires usage of WhenAnyValue then using Subscribe to ensure that the code will not be called until the ViewModel has been set.  This is not consistent with the WPF ViewModelViewHost, and may cause code that works in WPF to fail in Winforms due to this difference.

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the [latest stable](https://www.nuget.org/profiles/reactiveui) and [development snapshot](https://reactiveui.net/docs/getting-started/installation/nuget-packages/development-package-feed)**
I encountered this issue using ReactiveUI-Winforms version 7.4.1.  Based on examination of the source code for the [develop] branch, the Content property of the ViewModelControlHost is still set before the ViewModel property is, even in .

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
This issue should be very simple to fix, simply move line 98 (this.Content = view;) of https://github.com/reactiveui/ReactiveUI/blob/develop/src/ReactiveUI.Winforms/ViewModelViewHost.cs
after line 99 (view.ViewModel = x.ViewModel;).  This would ensure that the ViewModel would be non null during the WhenActivated block.

Asesjix:
I also think it makes more sense to set the viewmodel before adding the control, but I can not understand the described problem.
You are welcome to view my examples at [Asesjix/ReactiveUI.Winforms.Samples](https://github.com/Asesjix/ReactiveUI.Winforms.Samples).
[...]

PR #1650, opened 2018-05-18
Winforms stuff (TargetFramework, #1590, ReactiveUserControl)

Asesjix:
Fix, Feature

- TargetFramework is not set in csproj
- #1590 
- ReactiveUserControl is missing
- Design mode detection dont work for sub controls on a form/control

- TargetFramework is set in csproj
- ViewModelViewHost now set the viewmodel before the view is set to content (like wpf)
- Create ReactiveUserControl #1648 (used wpf as reference)
- Design mode detection work for sub controls on a form/control but the designer still crashed (fix comes in other pr)

**What might this PR break?**
Nothing

**Other information**:


Asesjix:
With PR #1651 and this PR the 💥 winforms designer works.

src/ReactiveUI.Winforms/ActivationForViewFetcher.cs
```diff
@@ -44,12 +46,38 @@ public IObservable<bool> GetActivationForView(IActivatable view)
...
+            } else {
                 // Show a friendly warning in the log that this view will never be activated
                 this.Log().Warn("Expected a view of type System.Windows.Forms.Control but it is {0}.\r\nYou need to implement your own IActivationForViewFetcher for {0}.", view.GetType());
                 return Observable<bool>.Empty;
             }
         }
+
+        private bool GetCachedIsDesignMode(Control control)
+        {
+            if (_isDesignModeCache.HasValue == false) {
```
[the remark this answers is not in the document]
glennawatson: If you use Lazy<> you probably don't need the thread protections since this is likely happening on the main thread. So turn those off potentially.

src/ReactiveUI.Winforms/ActivationForViewFetcher.cs
```diff
@@ -13,23 +13,25 @@ namespace ReactiveUI.Winforms
...
-                return Observable<bool>.Default;
-            }
-
             // Startup: Control.HandleCreated > Control.BindingContextChanged > Form.Load > Control.VisibleChanged > Form.Activated > Form.Shown
             // Shutdown: Form.Closing > Form.FormClosing > Form.Closed > Form.FormClosed > Form.Deactivate
             // https://docs.microsoft.com/en-us/dotnet/framework/winforms/order-of-events-in-windows-forms
 
             if (view is Control control) {
+
+                if (GetCachedIsDesignMode(control)) {
```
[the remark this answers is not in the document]
Asesjix: I'm not sure why I added the Designtime check there. 
I mean in Winforms Designer the events are executed and the control was considered active and that led to errors, because e.g. the ViewModel is null at design time.

Asesjix:
@glennawatson I hope I made you happy ;-)

src/ReactiveUI.Winforms/ReactiveUI.Winforms.csproj
```diff
@@ -1,12 +1,12 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
 
   <PropertyGroup>
-    <TargetFrameworks>net461</TargetFrameworks>
+    <TargetFramework>net461</TargetFramework>
```
glennawatson: Use TargetFrameworks. MsBuild.Sdk.Extras require it. Also there is fundamentally no difference apart from the ability to be recursive with checks.

src/ReactiveUI.Winforms/ActivationForViewFetcher.cs
```diff
@@ -44,12 +46,40 @@ public IObservable<bool> GetActivationForView(IActivatable view)
...
                 // Show a friendly warning in the log that this view will never be activated
                 this.Log().Warn("Expected a view of type System.Windows.Forms.Control but it is {0}.\r\nYou need to implement your own IActivationForViewFetcher for {0}.", view.GetType());
                 return Observable<bool>.Empty;
             }
         }
+
+        private bool GetCachedIsDesignMode(Control control)
+        {
+            if (_isDesignModeCache == null) {
+                _isDesignModeCache = new Lazy<bool>(() => {
```
glennawatson: Just create a constructor and put this in there (without the null check). 

There is a performance penalty for the if() check that is unnecessary.

Also you can simplify the statement and turn off multi threaded protection by passing null to the second parameter

```
_isDesignModeCache = new Lazy>bool>(() => GetIsDesignModel(control), false);
```

src/ReactiveUI.Winforms/ActivationForViewFetcher.cs
```diff
@@ -13,23 +13,25 @@ namespace ReactiveUI.Winforms
 {
     public class ActivationForViewFetcher : IActivationForViewFetcher, IEnableLogger
     {
+        private Lazy<bool> _isDesignModeCache = null;
```
glennawatson: You don't really need the = null; on the end of that. It'll default to that regardless.

Asesjix:
The TargetFramework i copied from Wpf Project.
https://github.com/reactiveui/ReactiveUI/blob/56b52f7c3ba8f74b1efb930a28799425e0086bbf/src/ReactiveUI.Wpf/ReactiveUI.Wpf.csproj#L8

glennawatson:
I will update the wpf project then. Thanks. The target framework is wrong regardless

src/ReactiveUI.Winforms/ActivationForViewFetcher.cs
Asesjix: The control is not available there. How should that work in the constructor?

src/ReactiveUI.Winforms/ActivationForViewFetcher.cs
```diff
@@ -44,12 +46,38 @@ public IObservable<bool> GetActivationForView(IActivatable view)
...
+            } else {
                 // Show a friendly warning in the log that this view will never be activated
                 this.Log().Warn("Expected a view of type System.Windows.Forms.Control but it is {0}.\r\nYou need to implement your own IActivationForViewFetcher for {0}.", view.GetType());
                 return Observable<bool>.Empty;
             }
         }
+
+        private bool GetCachedIsDesignMode(Control control)
+        {
+            if (_isDesignModeCache == null) {
```
glennawatson: I'm personally happy with the cached value regardless, Lazy<T> just added more code in the end over this design and we don't need the multithreaded protection in this case.

Issue #1016, opened 2016-01-07
WinForms Designer breaks when child control uses WhenActivated


shiftkey:
_Note:_ I haven't tested this on anything before VS2015 Update 1. 

Repro:

```
git clone https://github.com/shiftkey/reactiveui-winforms-example.git
git checkout 3d5ac4580b64dea7e65d771119c9b53a28e40fca
```

Then open the `WindowsFormsApplication1.sln` solution in Visual Studio, build the project and open the `Form1` design view. You should see something like this:

<img width="877" alt="screen shot 2016-01-07 at 8 50 17 pm" src="https://cloud.githubusercontent.com/assets/359239/12168124/4abfc996-b580-11e5-9155-34f08ee33a78.png">

**THIS IS IMPORTANT I THINK:** The **Don't know how to detect when WindowsFormsApplication1.CustomControl is activated/deactivated, you may need to implement IActivationForViewFetcher** error can be traced to [this code](https://github.com/reactiveui/ReactiveUI/blob/3f725c808b1d4c8457f0d3204e0a071aa462cd75/ReactiveUI/Activation.cs#L147). I'm not really across everything that's happening when RxUI is loaded in a design time context - I'd love any help if others know more... 

The stack trace is all kinds of boring:

```
at ReactiveUI.ViewForMixins.WhenActivated(IActivatable This, Func`1 block, IViewFor view)
at ReactiveUI.ViewForMixins.WhenActivated(IActivatable This, Action`1 block, IViewFor view)
at ReactiveUI.ViewForMixins.WhenActivated(IActivatable This, Action`1 block)
at WindowsFormsApplication1.CustomControl..ctor() in C:\Users\shiftkey\Documents\Visual Studio 2015\Projects\WindowsFormsApplication1\WindowsFormsApplication1\CustomControl.cs:line 20 
```

As soon as I saw this I felt myself dragged back to the distant past, when I cared about the design time experience of WPF libraries and found myself immensely frustrated by all  the rough edges that resulted when you tried to play the game. So I ran the other way.

So we have a top-level form that uses `WhenActivated` - it doesn't have an issue in the design view - and it hosts a child `UserControl` which also uses `WhenActivated - but it triggers the error.

I'm logging this for a couple of reasons:
- unless the rest of the core team has objections, I'd love to encourage `WhenActivated` usage as a Good Thing To Do
- `WhenActivated` seems to break for nested controls only - not surprising, maybe just a clue more than anything
[...]

PR #1598, opened 2018-04-30
feature: added integration tests

ghuntley:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

Adds integration tests that can be executed via VSTS on physical or simulated devices.

**What is the current behavior? (You can also link to an open issue here)**

Manual testing is the worst.

**What is the new behavior (if this is a feature change)?**

Automated testing is the best.

**What might this PR break?**

Nada. 

**Other information**:

We won't be hooking this into any CICD as part of this pull-request. The goal here is to scaffold this pattern:

image

On each and every platform we support.

Outstanding:
- [x] Share ViewModel between all platforms.
- [ ] Write tests for ViewModel assembly and ship them in `IntegrationTests.Shared.Tests`
- [ ] Add Xamarin Forms user interface
- [ ] Add UWP user interface
- [ ] Add WPF user interface
- [ ] Add Android user interface
- [ ] Add iOS user interface
- [ ] Add Mac user interface
- [ ] Add WinForms user interface.

PR #1651, opened 2018-05-19
Fix InitializeReactiveUI

Asesjix:
Bug fix

Currently use reactiveui a type from the splat Assembly to generate the type names for dependency regestration. This does not match the version to the specified assembly.
Fixes: #1016

Now a type from reactiveui is used.

**What might this PR break?**
Nothing

**Other information**:


ghuntley:
Will need to test on all platforms before merging.

Asesjix:
I think the test is done, but please check the test(https://github.com/reactiveui/ReactiveUI/pull/1598/commits/cd8c9e2a17efcd5226e0c4021f973ce6609fa111). The logic should be right, but I'm not sure with the platform adopt.

PR #1655, opened 2018-05-23
[XamForms] Allow navigation transition animations to be disabled

bratsche:
Xamarin.Forms navigation animations transitions by default, and apps
must explicitly opt out of the animation. RoutedViewHost hides these
APIs from the user.

This allows apps to apply DisableAnimation attribute to views in order
to instruct RoutedViewHost to not animate the transition.

Feature - right now the ReactiveUI.XamForms RoutedViewHost hides the Xamarin.Forms navigation API so that there's no way for apps to disable animation transitions. This change allows developers to mark views so that when they are pushed into the navigation stack there is no transition animation.

The current behavior is to call `PushAsync(x)`, which is the same as calling `PushAsync(x, true)`. There is currently no way to call `PushAsync(x, false)`.

This change would allow apps to apply a `[DisableAnimation]` attribute to their view, and when RoutedViewHost sees this attribute it will pass `false` to `PushAsync()`.

**What might this PR break?**

I don't think this is likely to break anything.

**Other information**:

I'd like some feedback from devs/maintainers if this seems like a decent approach. Originally I thought to have a new property on `IRoutableViewModel` but that was a little trickier to do and it would have the major downside of breaking backwards compatibility since everyone would have to add that property to their viewmodels.

### DIFF
--- a/src/ReactiveUI.XamForms/RoutedViewHost.cs
+++ b/src/ReactiveUI.XamForms/RoutedViewHost.cs
@@ -3,13 +3,14 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using Xamarin.Forms;
-using Splat;
-using ReactiveUI;
+using System.Reactive;
 using System.Reactive.Linq;
 using System.Reactive.Threading.Tasks;
-using System.Diagnostics;
-using System.Reactive;
+using System.Reflection;
+
+using Xamarin.Forms;
+
+using Splat;
 
 namespace ReactiveUI.XamForms
 {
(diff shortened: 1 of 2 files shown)


src/ReactiveUI.XamForms/RoutedViewHost.cs
```diff
@@ -94,7 +94,16 @@ public RoutedViewHost()
                         }
                         else
                         {
-                            await this.PushAsync(x);
+                            bool animated = true;
+                            var type = x.GetType();
+                            Attribute[] attrs = Attribute.GetCustomAttributes(type);
```
glennawatson: Can we get a unit test for this one at all? Know it's a been a while since you did the PR.

Also I think you can use the type to get the Attribute..

eg 

var attribute = type.GetCustomAttribute<DisableAnimationAttribute>();
if (attribute != null) type thing.

PR #1503, opened 2017-10-06
[WIP] feature: bring reactiveui.fody into the main project

ghuntley:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

feature

**What is the current behavior? (You can also link to an open issue here)**

@kswoll has been doing a stunning job over the last couple years maintaining his addin. I think it's high time to in-line his work and make it official. We had conversations back in May - this PR just makes it official and does the needed to make it happen.

**What is the new behavior (if this is a feature change)?**

We are in-lining https://github.com/kswoll/ReactiveUI.Fody and making it an official thing.  Kirk Woll will get commit bit access to the `reactiveui/reactiveui` repository as a contributor/maintainer so that he can still work autonomously but over time the core team will also increase their knowledge of things they maintain to include fody.

**What might this PR break?**

Nothing

**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Other information**:

* We need @kswoll to be the person that presses the merge button. 
* This PR still needs work before merging.

PR #1671, opened 2018-06-08
Finish up ReactiveUI.Fody

gtbuchanan:
This is an attempt to finish up the ReactiveUI.Fody integration in #1503.

This PR includes the following changes:
* Fix package versioning
* Fix ReactiveUI.Fody NuGet packaging
* Include ReactiveUI.Fody tests in the build
* Add ReactiveUI.Fody projects to ReactiveUI.sln

I have tested the NuGet package output in a project at my company and it's working great.

I don't think there is anything else remaining. The code ported over from the original project could use some cleanup, but I'd say that shouldn't hold up getting this out to the masses.


gtbuchanan:
I'm not entirely sure why the ReactiveUI tests failed when I didn't make any changes relating to them.

ghuntley:
Let's inline this to the fody branch. Ta.

PR #1680, opened 2018-07-01
 Add ResolveStrategy parameter to WireUpControls in ReactiveUI.AndroidSupport

### LINKED ISSUE #1679: WireUpControls for Android.Support.V4.App.Fragment is missing ResolveStrategy parameter

Bug

When using `[WireUpResource("some_id_ name")]` on properties inside a `Android.Support.V4.App.Fragment` fragment, then calling `this.WireUpControls();` in `OnCreateView`, an exception is thrown much like #1231

1. Create a new Fragment that is based on `Android.Support.V4.App.Fragment` 
1. Add a Control property to be wired up:

   ```C#
        [WireUpResource("text_some_name")]
        public TextInputEditText SomeTextInput { get; private set; }
   ``` 
1. Inflate with a layout with ids that do not match the property name in the fragmet's class but instead match with the `WireUpResource` attribute.
1. Override `OnCreateView(...)` with:

   ```C#
        public override View OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
        {
            var view = inflater.Inflate(Resource.Layout.some_fragment_dialog, container, false);

            if (view != null)
                this.WireUpControls(view);

            return view;
        }
   ```
1. Note that `this.WireUpControls(view);` does not have any overrides that accept a `ResolveStrategy ` parameter.

Supplying `ResolveStrategy.ExplicitOptIn` would allow use of properties with `WireUpResource` attributes

To be more consistent with `WireUpControls` method that is used with an `Activity` or `Fragment` class 

> Note that I'm using using the Fragment class from the Android support library instead of the default Fragment (as recommended by the Official Android Documentation)

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the [latest stable](https://www.nuget.org/profiles/reactiveui) and [development snapshot](https://reactiveui.net/docs/getting-started/installation/nuget-packages/development-package-feed)**

v8.3.1 (stable)

I've yet to test this against development

Suggest adding another `WireUpControls` extentionmethod in `ReactiveUI.AndroidSupport/ControlFetcherMixin.cs` that accepts the same (or similar) parameters much like `WireUpControls` for an `Activity` or `Fragment` (As seen in `ReactiveUI/Platforms/android/ControlFetcherMixins.cs`)

NZSmartie:

Fixes #1679

See https://github.com/reactiveui/ReactiveUI/issues/1679

Adds a `ResolveStrategy` parameter to `WireUpControls` for AppCompatActivity and Fragment in `Android.Support.V{4,7}...` 

**What might this PR break?**

`ReactiveUI.AndroidSupport.ControlFetcherMixins.WireUpControls(AppCompatActivity)` is removed in favour of a more capable `ReactiveUI.ControlFetcherMixins.WireUpControls(Activity, ResolveStrategy)`

**Other information**:

### DIFF
--- a/src/ReactiveUI.AndroidSupport/ControlFetcherMixin.cs
+++ b/src/ReactiveUI.AndroidSupport/ControlFetcherMixin.cs
@@ -12,6 +12,7 @@ using Android.App;
 using Android.Views;
 using Java.Interop;
 using Android.Support.V7.App;
+using static ReactiveUI.ControlFetcherMixin;
 
 namespace ReactiveUI.AndroidSupport
 {
@@ -67,20 +68,21 @@ namespace ReactiveUI.AndroidSupport
                 () => This.FindViewById(controlIds[propertyName.ToLowerInvariant()]).JavaCast<T>());
         }
 
-        /// <summary>
-        /// This should be called in the Fragement's OnCreateView, with the newly inflated layout
-        /// </summary>
-        /// <param name="This"></param>
-        /// <param name="inflatedView"></param>
-        public static void WireUpControls(this global::Android.Support.V4.App.Fragment This, View inflatedView)
+		/// <summary>
+		/// A helper method to automatically resolve properties in an <see cref="Android.Support.V4.App.Fragment"/> to their respective elements in the layout.
+		/// This should be called in the Fragement's OnCreateView, with the newly inflated layout
+		/// </summary>
+		/// <param name="This"></param>
+		/// <param name="inflatedView">The newly inflated <see cref="View"/> returned from <see cref="LayoutInflater.Inflate"/>.</param>
+		/// <param name="resolveMembers">The strategy used to resolve properties that either subclass <see cref="View"/>, have a <see cref="WireUpResourceAttribute"/> or have a <see cref="IgnoreResourceAttribute"/></param>
+		public static void WireUpControls(this global::Android.Support.V4.App.Fragment This, View inflatedView, ResolveStrategy resolveMembers = ResolveStrategy.Implicit)
 		{
-			var members = This.GetType().GetRuntimeProperties()
-				.Where(m => m.PropertyType.IsSubclassOf(typeof(View)));
+			var members = This.getWireUpMembers(ResolveStrategy.Implicit);
 
 			members.ToList().ForEach(m => {
 				try {
 					// Find the android control with the same name from the view
-					var view = inflatedView.getControlInternal(m.PropertyType, m.Name);
+					var view = inflatedView.getControlInternal(m.PropertyType, m.getResourceName());
 
 					// Set the activity field's value to the view with that identifier
 					m.SetValue(This, view);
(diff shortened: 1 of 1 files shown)


NZSmartie:
I do not see any tests for WireUpControls in either `ReactiveUI` or `ReactiveUI.AndroidSupport` so i'm not sure how to add them or if I should?

src/ReactiveUI.AndroidSupport/ControlFetcherMixin.cs
```diff
@@ -67,20 +68,21 @@ public static T GetControl<T>(this View This, [CallerMemberName]string propertyN
...
-        /// <param name="inflatedView"></param>
-        public static void WireUpControls(this global::Android.Support.V4.App.Fragment This, View inflatedView)
+		/// <summary>
+		/// A helper method to automatically resolve properties in an <see cref="Android.Support.V4.App.Fragment"/> to their respective elements in the layout.
+		/// This should be called in the Fragement's OnCreateView, with the newly inflated layout
+		/// </summary>
+		/// <param name="This"></param>
+		/// <param name="inflatedView">The newly inflated <see cref="View"/> returned from <see cref="LayoutInflater.Inflate"/>.</param>
+		/// <param name="resolveMembers">The strategy used to resolve properties that either subclass <see cref="View"/>, have a <see cref="WireUpResourceAttribute"/> or have a <see cref="IgnoreResourceAttribute"/></param>
+		public static void WireUpControls(this global::Android.Support.V4.App.Fragment This, View inflatedView, ResolveStrategy resolveMembers = ResolveStrategy.Implicit)
```
glennawatson: Might be worth adding some unit tests to the Android tests project just to make sure we don't have a regression in the future.

src/ReactiveUI.AndroidSupport/ControlFetcherMixin.cs
```diff
@@ -92,17 +94,18 @@ public static void WireUpControls(this global::Android.Support.V4.App.Fragment T
 		}
 
 		/// <summary>
-		///
+		/// A helper method to automatically resolve properties in an <see cref="AppCompatActivity"/> to their respective elements in the layout.
 		/// </summary>
-		public static void WireUpControls(this AppCompatActivity This)
+		/// <param name="This"></param>
+		/// <param name="resolveMembers">The strategy used to resolve properties that either subclass <see cref="View"/>, have a <see cref="WireUpResourceAttribute"/> or have a <see cref="IgnoreResourceAttribute"/></param>
+		public static void WireUpControls(this AppCompatActivity This, ResolveStrategy resolveMembers = ResolveStrategy.Implicit)
```
Qonstrukt: I don't think this method adds anything that wouldn't already work with the extension method on the regular `Activity`, since that is what `AppCompatActivity` extends from.

NZSmartie: I would agree, but unfortunately, `Android.Support.V4.App.Fragment` does not extend from Android.App.Fragment` in C# thanks to the Java bindings

```C#
namespace Android.Support.V4.App
{
    [Register("android/support/v4/app/Fragment", DoNotGenerateAcw = true)]
    public class Fragment : Java.Lang.Object, ILifecycleOwner, IJavaObject, IDisposable, IComponentCallbacks, Views.View.IOnCreateContextMenuListener
    {
        [Register(".ctor", "()V", "")]
        public Fragment();
        protected Fragment(IntPtr javaReference, JniHandleOwnership transfer);

        // ...
```

So these extension methods are required

Qonstrukt: I noticed so for Fragments, see my comments in the issue, but this is the case fro `AppCompatActivity` though.

NZSmartie: Ah, gotcha. Will get on to it

NZSmartie:
Until  #1598 is merged. There isn't a standard way to perform platform dependant unit tests. I'm reluctant too add one for WireUpContols without the testing framework in place.

NZSmartie:
@glennawatson: I've rebased onto the master branch

glennawatson:
Thanks @NZSmartie

Issue #1675, opened 2018-06-19
ActOnEveryObject appears not to track resets correctly


tcolemanld:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

ActOnEveryObject does not correctly track if `This` collection is reset

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

```csharp
public class Foo : ReactiveObject {
	public Foo(int id) {
		Id = $"Foo{id}";
	}
	public string Id { get; }
}

[Fact]
public void ActOnEveryObject_WithReset() {
	using (TestUtils.WithScheduler(ImmediateScheduler.Instance)) {
		var original = new ReactiveList<Foo>();

		var copy = new ReactiveList<Foo>();
		original.ActOnEveryObject(
			onAdd: x => {
				_output.WriteLine($"Adding {x.Id} to copy");
				copy.Add(x);
			},
			onRemove: x => {
				_output.WriteLine($"Removing {x.Id} from copy");
				copy.Remove(x);
			});

		var foo1 = new Foo(1);
		var foo2 = new Foo(2);
		original.Add(foo1);
		original.Add(foo2);

		Assert.Equal(2, original.Count);
		Assert.Equal(2, copy.Count);

		using (original.SuppressChangeNotifications()) {
			original.Remove(foo2);
		}

		Assert.Equal(1, original.Count);
		Assert.Equal(1, copy.Count); // fails. actually 2.

		/* output is
			Adding Foo1 to copy
			Adding Foo2 to copy
			Removing Foo1 from copy       <- curious
			Adding Foo1 to copy
		*/
	}
}
```

**What is the expected behavior?**

`copy` should contain the same items as `original`

**What is the motivation / use case for changing the behavior?**

ActOnEveryObject docs say that reset is handled.

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the [latest stable](https://www.nuget.org/profiles/reactiveui) and [development snapshot](https://reactiveui.net/docs/getting-started/installation/nuget-packages/development-package-feed)**

Windows. ReactiveUI 8.3.1

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

I suppose the problem is either with ActOnEveryObject, or possibly with ReactiveList<> emitting the wrong notifications.

Wouterdek:
Yes, I have noticed this bug too.
[...]

PR #1685, opened 2018-07-10
Fix ActOnEveryObject missing changes when SuppressChangeNotifications is used

Wouterdek:
Bug fix.

Previously `ActOnEveryObject()` would not handle list changes correctly if the changes were made while `SuppressChangeNotifications()` was used. This is because `SuppressChangeNotifications()` currently publishes a reset `Changing` notification _after_ the changes have been made. See issue #1675.

`SuppressChangeNotifications()` now publishes the reset `Changing` notification _before_ the changes are made, when the suppression is first enabled.

**What might this PR break?**
Protected virtual method `publishResetNotification()` is removed. It is not used in the class anymore and current subclasses might be overriding it because it affects Reset() and SuppressChangeNotifications(), which it doesn't anymore.

**Other information**:


src/ReactiveUI/ReactiveList.cs
```diff
@@ -579,7 +580,10 @@ public IDisposable SuppressChangeNotifications()
 
             return new CompositeDisposable(this.suppressChangeNotifications(), Disposable.Create(() => {
                 if (Interlocked.Decrement(ref _resetNotificationCount) == 0) {
-                    publishResetNotification();
+                    if (resetEventArgs == null) {
```
glennawatson: Potentially a threading issue here, since your dispose may not be called in the same thread as when it was created. Probably should use a Interlocked.Exchange or something?

glennawatson: It might be alright given that it's inside the Interlocked.Decrement() though.

Wouterdek: As `resetEventArgs` is set (if it is set) before the disposable is returned, and nothing can change its value until the disposable is called, I don't really see how a race condition could occur? Duplicate/missing changed notifications cannot occur since `Interlocked.Decrement` is still used. Am I missing something?

glennawatson: Yeah I agree, was just a initial reaction until I saw everything inside the Decrement statement.

src/ReactiveUI.Tests/ReactiveCollectionTest.cs
```diff
@@ -481,6 +481,64 @@ public void GetAResetWhenWeAddALotOfItems()
             Assert.Equal(1, reset.Count);
         }
 
+        [Fact]
+        public void ActOnEveryObjectShouldHandleAddRemoveAndClear()
```
olevett: I think from a clarity point of view, this test is doing too much, I think each separate case of ActOnEvery… should be a separate test.

src/ReactiveUI.Tests/ReactiveCollectionTest.cs
```diff
@@ -481,6 +481,64 @@ public void GetAResetWhenWeAddALotOfItems()
...
+
+            fixture.Add(testObj);
+            Assert.Equal(1, testObj.NullableInt);
+
+            fixture.Clear();
+            Assert.Equal(0, testObj.NullableInt);
+        }
+
+        [Fact]
+        public void ActOnEveryObjectShouldNotMissChangesUnderSuppressedNotifications()
```
olevett: Likewise, I think this would be clearer as a couple of tests

src/ReactiveUI/ReactiveList.cs
```diff
@@ -568,7 +563,13 @@ public bool ChangeTrackingEnabled
 
         public IDisposable SuppressChangeNotifications()
         {
-            Interlocked.Increment(ref _resetNotificationCount);
+            NotifyCollectionChangedEventArgs resetEventArgs = null;
+
+            if (Interlocked.Increment(ref _resetNotificationCount) == 1) {
+                // Changes were not suppressed before, publish a Changing reset notification.
+                resetEventArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset);
```
olevett: Is there a reason to not use the reset/publishResetNotification in here? Wondering if we can reduce the breaking changes to a minimum.

Wouterdek: Note that with this patch only a `Changed`, but no `Changing` notification, is pushed in the disposable. Using `publishResetNotification()` in here would cause a duplicate `Changing` notification after the changes have been made, which would be unexpected but perhaps less breaking. I removed `publishResetNotification()` because it was used only in `Reset()` and `SuppressChangeNotifications()` and since it is no longer used there I feel like it would be better to have a compiler error rather than unexpected behaviour when overriding this method.

I guess it comes down to this: Do you prefer a duplicate, technically incorrect, `Changing` notification but no breaking changes? Or would you rather have the (small) breaking change and no duplicate notifications?

olevett:
Only just spotted, can we change this PR to be against master rather than develop?

glennawatson:
That merge we definitely don't want. Now we got mass number of changes. You probably want to do some sort of rebase operation. Another option is close this pr, then you just cherry pick the commits into a new branch from master your changes and submit a new pr.

Wouterdek:
I'll close this one, and open a cherry-picked new one later.

Also, not sure if you always want your PRs against master, but if so you might want to update [the guidelines](https://reactiveui.net/contribute/features-and-patches/submitting-a-pull-request).

glennawatson:
Good point. I will do a pr for that later today. It was something that changed with the RxUI 8 release.

PR #1688, opened 2018-07-12
Fix ActOnEveryObject missing changes when SuppressChangeNotifications is used

Wouterdek:
_Copied from PR #1685, this is the same PR but against the master branch instead of develop._

Bug fix.

Previously `ActOnEveryObject()` would not handle list changes correctly if the changes were made while `SuppressChangeNotifications()` was used. This is because `SuppressChangeNotifications()` currently publishes a reset `Changing` notification _after_ the changes have been made. See issue #1675.

`SuppressChangeNotifications()` now publishes the reset `Changing` notification _before_ the changes are made, when the suppression is first enabled.

**What might this PR break?**
Protected virtual method `publishResetNotification()` is removed. It is not used in the class anymore and current subclasses might be overriding it because it affects Reset() and SuppressChangeNotifications(), which it doesn't anymore.

**Other information**:


glennawatson:
To pass the failing test you'll need to remove line 601 where it references publishResetNotification

We have API checker that confirms the API hasn't changed, and since you have a potentially breaking change we'll have to update the approved API. 

https://github.com/reactiveui/ReactiveUI/blob/master/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.approved.txt

olevett:
To echo what @glennawatson has said, if you run the tests locally, you'll get an "actual" output and an "approved" if you update the approved with the actual, it should make the test happier :)

Wouterdek:
I removed the line, but the error seems to persist. I'm not sure what the problem is...

glennawatson:
I'll grab your branch today and see what's going on.

glennawatson:
When I run build.cmd it shows the following error message: 

Failed Approval: Received file C:\OpenSource\branch\ReactiveUI\src\ReactiveUI.Tests\API\ApiApprovalTests.ReactiveUI.received.txt does not match approved file C:\OpenSource\branch\ReactiveUI\src\ReactiveUI.Tests\API\ApiApprovalTests.ReactiveUI.approved.txt.

The files have to byte match. Changed it for you.

Wouterdek:
Thank you!

PR #1694, opened 2018-07-23
housekeeping: header incorrectly said MS-PL instead of MIT

ghuntley:

house keeping

Headers incorrectly say MS-PL

Headers state the correct license (MIT)

**What might this PR break?**

It was a programmatic find and replace

image

**Other information**:

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,5 @@
 // Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MS-PL license.
+// The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
 //////////////////////////////////////////////////////////////////////
--- a/samples/stylecop.json
+++ b/samples/stylecop.json
@@ -6,7 +6,7 @@
       "documentationCulture": "en-US",
       "documentInterfaces": true,
       "documentInternalElements": false,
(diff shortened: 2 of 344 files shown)


glennawatson:
I've opened a bug over here https://github.com/onovotny/MSBuildSdkExtras/issues/98 about the weird msbuid.sdk.extras error.

ghuntley:
If we need to revert out that commit that updates msbuild.extras then happy to do so.

glennawatson:
msbuild.sdk.extras changed a bit in the latest version, you now need to add additional properties to wpf/winforms applications

```Support for WPF and Windows Forms requires an opt-in property to enable: Set ExtrasEnableWpfProjectSetup/ExtrasEnableWinFormsProjectSetup to true to include required references and default items.```

This is in regards to the recent upgrade to Msbuild.sdk.extras

but the other bug isn't explained by those either though

You may want to revert that until those items can be sorted out by Oren as well.

onovotny:
How are the extras being added? Will fix next week, but it should work as an SDK reference instead of as a PackageReference.

glennawatson:
Yeah at the moment it's set as a ProjectReference inside the Directory.Build.Props file.

onovotny:
I'd switch that to set the version in a global.json and then use the Sdk= attribute. 

I'll still fix the underlying issue though.

glennawatson:
@onovotny About to submit a new PR with the change to the newer format. Seemed to have solved those issues.

glennawatson:
See https://github.com/reactiveui/ReactiveUI/pull/1695 implemented your recommendations.

PR #1701, opened 2018-08-02
feat: Update ReactiveUI to use Reactive Extensions 4.0.0 as minimum

### LINKED ISSUE #1672: Update RxUI to use System.Reactive >= 4
Upgrade

Referencing System.Reactive 3.1.1

Update to System.Reactive >= 4.0.0

This avoids having NU1608 come up when compiling. 
warning NU1608: Detected package version outside of dependency constraint: ReactiveUI 8.2.1 requires System.Reactive (>= 3.1.1 && < 4.0.0) but version System.Reactive 4.0.0 was resolved.

RxUI is the main consumer of Rx and should maintain version parity.

**Which versions of ReactiveUI, and which platform / OS are affected by this issue?**
8.2.1, all

glennawatson:
Fixes #1672 

This changes the min version for ReactiveUI to use Rx 4.0.0

We are getting a few reports from users where upgrading to Rx 4.0.0 fixes their issues.

Also adding .NET .Core 2.1 as a target. It was partially done in a previous PR but this adds it to the other projects.

Use Rx 3.1.1

Use Rx 4.0.0 and also .NET Core 2.1 added to other projects.

**What might this PR break?**
Users using Rx 3.1.1. will have to upgrade their package reference in nuget.


src/ReactiveUI.Testing/ReactiveUI.Testing.csproj
```diff
@@ -1,14 +1,18 @@
...
     <PackageId>ReactiveUI.Testing</PackageId>
   </PropertyGroup>
   
   <ItemGroup>
-    <PackageReference Include="Microsoft.Reactive.Testing" Version="3.1.1" />
+    <PackageReference Include="Microsoft.Reactive.Testing" Version="4.0.0" />
     <ProjectReference Include="..\ReactiveUI\ReactiveUI.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'Xamarin.Mac20' ">
```
glennawatson: I had to reference netstandard to get the Xamarin.Mac20 to compile after updating. Bit weird.

onovotny:
You should not need a `netcoreapp2.1` target unless you're using new API's there (like `Span<T>`).

glennawatson:
Would it not make it easier for consuming users to have a .NET Core 2.1? Or would the JIT just go based on their API? Also my understanding is also they done a ton of performance operations on .NET Core 2.1 so potentially users targetting the framework could take advantage.

onovotny:
.NET Core 2.1 is just an API surface area. It's no different than releasing .NET 4.6 and 4.6.1 at the same time. No need to unless you're using API's not in the lower version.

glennawatson:
This is something that came up before when I was editing one of Geoff media announcements, eg which form of English we should use since he was using Australian english. We should probably document a standard in that regards.

PR #1702, opened 2018-08-02
Fix: Made POCOObservableForProperty include property name in warning

### LINKED ISSUE #864: The BindTo/OneWayBind/Bind POCO warning does not clarify which property the expression is referring to, just the owning type
Whenever I try to use BindTo/OneWayBind/Bind in the View to bind a VM property to a property of a XAML view element I get the following warning:

> POCOObservableForProperty: rx_bindto_test.MainWindow is a POCO type and won't send change notifications, WhenAny will only return a single value!

The thing is it still works correctly, even when I switch ViewModels.

Replacing the binding with a WhenAnyValue combined with a Susbcribe() to update the property gets rid of the warning.

Please check out http://stackoverflow.com/questions/30352447/using-reactiveuis-bindto-to-update-a-xaml-property-generates-a-warning for a short example to reproduce the issue.

glennawatson:
bug fix.
Fixes #864

Makes POCOObservableForProperty  error message include the property name.

Just lists the full type name.

Now includes the property name.

**What might this PR break?**
Could be performance issue but only if the error happens, and only once per instance of the type.

### DIFF
--- a/src/ReactiveUI/POCOObservableForProperty.cs
+++ b/src/ReactiveUI/POCOObservableForProperty.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -22,15 +22,13 @@ namespace ReactiveUI
             return 1;
         }
 
-        static readonly Dictionary<Type, bool> hasWarned = new Dictionary<Type, bool>();
-        public IObservable<IObservedChange<object, object>> GetNotificationForProperty(object sender, Expression expression, bool beforeChanged = false)
+        private static readonly Dictionary<(Type, string), bool> hasWarned = new Dictionary<(Type, string), bool>();
+        public IObservable<IObservedChange<object, object>> GetNotificationForProperty(object sender, Expression expression, string propertyName, bool beforeChanged = false)
         {
             var type = sender.GetType();
-            if (!hasWarned.ContainsKey(type)) {
-                this.Log().Warn(
-                    "{0} is a POCO type and won't send change notifications, WhenAny will only return a single value!",
-                    type.FullName);
-                hasWarned[type] = true;
+            if (!hasWarned.ContainsKey((type, propertyName))) {
+                this.Log().Warn($"The class {type.FullName} property {propertyName} is a POCO type and won't send change notifications, WhenAny will only return a single value!");
+                hasWarned[(type, propertyName)] = true;
             }
 
             return Observable.Return(new ObservedChange<object, object>(sender, expression), RxApp.MainThreadScheduler)
--- a/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.approved.txt
+++ b/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.approved.txt
@@ -175,7 +175,7 @@ namespace ReactiveUI
     public interface ICreatesObservableForProperty : Splat.IEnableLogger
     {
         int GetAffinityForObject(System.Type type, string propertyName, bool beforeChanged = False);
-        System.IObservable<ReactiveUI.IObservedChange<object, object>> GetNotificationForProperty(object sender, System.Linq.Expressions.Expression expression, bool beforeChanged = False);
+        System.IObservable<ReactiveUI.IObservedChange<object, object>> GetNotificationForProperty(object sender, System.Linq.Expressions.Expression expression, string propertyName, bool beforeChanged = False);
     }
     public interface IHandleObservableErrors
     {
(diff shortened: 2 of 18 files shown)


src/ReactiveUI/POCOObservableForProperty.cs
```diff
@@ -27,9 +27,9 @@ public int GetAffinityForObject(Type type, string propertyName, bool beforeChang
         {
             var type = sender.GetType();
             if (!hasWarned.ContainsKey(type)) {
-                this.Log().Warn(
-                    "{0} is a POCO type and won't send change notifications, WhenAny will only return a single value!",
-                    type.FullName);
+                var info = expression.GetMemberInfo();
```
olevett: Is it worth nullchecking expression here? 

Also, there's potentially a little oddity in the way the hasWarned cache is used (you'd get a notification for the first property of the type that's bound to, but no more) I think it's probably fine though :)

glennawatson:
I changed over to providing the different ICreatePropertyObservable to having the property name passed in. This already happens with the other method in the ICreatesPropertyObservable and the base class already asks for the GetMemberInfo().Name.

I did bench marking to confirm if this is indeed quicker. 

Here are the old times:
```
BenchmarkDotNet=v0.11.0, OS=Windows 10.0.17134.191 (1803/April2018Update/Redstone4)
Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
  [Host] : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3160.0
  Clr    : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3160.0
  Core   : .NET Core 2.0.9 (CoreCLR 4.6.26614.01, CoreFX 4.6.26614.01), 64bit RyuJIT


          Method |    Job | Runtime |      Mean |     Error |    StdDev | Rank |
---------------- |------- |-------- |----------:|----------:|----------:|-----:|
 PropertyBinding |    Clr |     Clr | 10.201 us | 0.2273 us | 0.2706 us |    2 |
 PropertyBinding |   Core |    Core |  9.241 us | 0.1816 us | 0.2162 us |    1 |
 PropertyBinding | CoreRT |  CoreRT |        NA |        NA |        NA |    ? |
```

Here are the new times:
```
BenchmarkDotNet=v0.11.0, OS=Windows 10.0.17134.191 (1803/April2018Update/Redstone4)
Intel Core i7-7700HQ CPU 2.80GHz (Max: 2.00GHz) (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
  [Host] : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3160.0
  Clr    : .NET Framework 4.7.2 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.3160.0
  Core   : .NET Core 2.0.9 (CoreCLR 4.6.26614.01, CoreFX 4.6.26614.01), 64bit RyuJIT


          Method |    Job | Runtime |     Mean |     Error |    StdDev | Rank |
---------------- |------- |-------- |---------:|----------:|----------:|-----:|
 PropertyBinding |    Clr |     Clr | 6.041 us | 0.1122 us | 0.1152 us |    2 |
 PropertyBinding |   Core |    Core | 5.516 us | 0.0478 us | 0.0399 us |    1 |
 PropertyBinding | CoreRT |  CoreRT |       NA |        NA |        NA |    ? |
```

As you can see it halves the time in that function. 

Here's my benchmark code for reference:
```
    [ClrJob, CoreJob, MonoJob, CoreRtJob]
    [RPlotExporter, RankColumn]
    public class INPCObservableForPropertyBenchmarks
    {
        private string propertyName;
        private Expression<Func<TestClassChanged, string>> expr = x => x.Property1;
        private Expression exp;
        private INPCObservableForProperty instance = new INPCObservableForProperty();


        [GlobalSetup]
        public void Setup()
        {
            exp = Reflection.Rewrite(expr.Body);
            propertyName = exp.GetMemberInfo().Name;
        }

        [Benchmark]
        public void PropertyBinding()
        {
            var testClass = new TestClassChanged();

            var changes = new List<IObservedChange<object, object>>();
            instance.GetNotificationForProperty(testClass, exp, false).Subscribe(c => changes.Add(c));
        }

        class TestClassChanged : INotifyPropertyChanged
        {
            string property;

            string property2;

            public string Property1
            {
                get { return property; }
                set
                {
                    property = value;
                    OnPropertyChanged();
                }
            }

            public string Property2
            {
                get { return property2; }
                set
                {
                    property2 = value;
                    OnPropertyChanged();
                }
            }

            public event PropertyChangedEventHandler PropertyChanged;

            public void OnPropertyChanged([CallerMemberName] string propertyName = null)
            {
                var handler = PropertyChanged;
                if (handler != null) {
                    handler(this, new PropertyChangedEventArgs(propertyName));
                }
            }
        }
    }
```

src/ReactiveUI/POCOObservableForProperty.cs
glennawatson: Yeah, I use a Tuple for storing if we have seen the value before now. 

Also the expression.GetMemberInfo() does have a performance hit and the code that gets the observable for the WhenAny code etc already passed the GetMemberInfo().Name into another function so now I cache it. As you can see in the bench marking below it reduces the time per call but is a wide spread change now.

PR #1705, opened 2018-08-03
housekeeping: Update the testing projects and cake file to have latest nuget packages.

glennawatson:

Updated to the latest versions of the test frameworks. This potentially eliminates some issues I been having with the tests.

The only one I couldn't upgrade easily was the ApiGenerator so decided to leave that one out.

old versions.

new versions.

**What might this PR break?**
Since it's only the testing framework it's limited to the tests. The tests are all running correctly.


glennawatson:
@RLittlesII is going to upgrade one of our Cake dependencies so that we can use Cake 0.26.0, then I will include that upgrade as part of this PR.

PR #1709, opened 2018-08-07
Bug Make SourceLink the latest version.

glennawatson:
This is because the SourceLink data doesn't seem to be uploading properly.

Updates to the latest SourceLink

Be a older version

Be a newer version

**What might this PR break?**
Shouldn't unless there is a breaking change in SourceLink.

### DIFF
--- a/src/Directory.build.props
+++ b/src/Directory.build.props
@@ -34,7 +34,6 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(IsTestProject)' != 'true' and '$(SourceLinkEnabled)' != 'false'">
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.1" />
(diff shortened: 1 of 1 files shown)


glennawatson:
This is broken at the moment, the Rx 4 project uses the new Microsoft libraries as well.

src/Directory.build.props
```diff
@@ -34,6 +34,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(IsTestProject)' != 'true' and '$(SourceLinkEnabled)' != 'false'">
-    <PackageReference Include="SourceLink.Create.CommandLine" Version="2.8.0" PrivateAssets="All" /> 
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.1" />
```
olevett: Somewhat surprised we need the extra dependency, but 👍

glennawatson: I copied pasted Rx to be honest. Will try without and see if it works still.

PR #1715, opened 2018-08-12
(wip) Animate RecyclerView changes and migrate to DynamicData

cabauman:
Feature:
- Take advantage of the free "collection changed" animations provided by RecyclerView.
- Improve efficiency.
- Migrate to using DynamicData.

No animation when an item is added, removed, changed, or moved.

Currently, _NotifyDataSetChanged()_ is called for every little change, "forcing LayoutManagers to fully rebind and relayout all visible views." So, the RecyclerView is unable to infer any small changes that it would otherwise be able to animate.

Taken from the official RecyclerView.Adapter documentation:

> If you are writing an adapter it will always be more efficient to use the more specific change events if you can. Rely on notifyDataSetChanged() as a last resort.

Source: [https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter#notifydatasetchanged](https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter#notifydatasetchanged)

Animations for all the aforementioned scenarios and eliminate unneeded rebinds/relayouts.

**What might this PR break?**
Nothing

**Other information**:
Tested on my Redmi Android phone.


src/ReactiveUI.AndroidSupport/ReactiveRecyclerViewAdapter.cs
```diff
@@ -28,13 +28,21 @@ public abstract class ReactiveRecyclerViewAdapter<TViewModel> : RecyclerView.Ada
     {
         readonly IReadOnlyReactiveList<TViewModel> list;
 
-        IDisposable _inner;
+        IDisposable inner1;
+        IDisposable inner2;
+        IDisposable inner3;
```
glennawatson: Use a CompositeDisposable and DisposeWith()

Issue #1615, opened 2018-05-06
"We should get your projects into the Xamarin QA harness to catch regressions"


ghuntley:
# Background

The community feels (rightly or wrongly) that they aren't getting enough support from the product teams and that they are are using data to solve the most common use case. If true, the problem with that is we the maintainers ("the power users") are the least common use case but your most common use case depends upon us. When you break us, you break the ecosystem.

ReactiveUI v8 could have been released on the 24th of May 2017 but was [only released a couple of days ago](https://reactiveui.net/blog/2018/05/reactiveui-v8.0.0-released) as I was waiting for the toolchain to settle on down. The historical specifics can be discussed in private - that's not the purpose of this thread. 

Let's focus on the future - this future

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr">We should get your projects into@our QA harness to catch these.</p>&mdash; Miguel de Icaza (@migueldeicaza) <a href="https://twitter.com/migueldeicaza/status/992563477398081536?ref_src=twsrc%5Etfw">May 5, 2018</a></blockquote>

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr">It&#39;s been offered before <a href="https://twitter.com/migueldeicaza?ref_src=twsrc%5Etfw">@migueldeicaza</a> but the project was in a deep branch due to netstandard20.<br><br>We are green and on VSTS.<br><br>I want to take you up on this offer.<br><br>How can we?</p>&mdash; 𝓖eoffrey 𝓗untley (@GeoffreyHuntley) <a href="https://twitter.com/GeoffreyHuntley/status/992567117554761729?ref_src=twsrc%5Etfw">May 5, 2018</a></blockquote>

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr">Email me</p>&mdash; Miguel de Icaza (@migueldeicaza) <a href="https://twitter.com/migueldeicaza/status/992567267903852545?ref_src=twsrc%5Etfw">May 5, 2018</a></blockquote>

# Scope
- Xamarin Forms
- Xamarin Android
- Xamarin iOS
- Xamarin Mac
- Visual Studio for Windows
- NuGet 

Visual Studio for Mac is out of scope because it does not have the necessary features implemented to build ReactiveUI. When it becomes possible to compile ReactiveUI using this product I'd like to see it integrated as well.

# Measurement of success
[...]

PR #1717, opened 2018-08-13
Allow building against prerelease versions of VS

kdubau:

feature

build only targets stable VS instances

include prerelease instances of VS when using vswhere.exe

**What might this PR break?**

There is no way, with this implementation, to prefer Stable VS if Preview is also installed.

**Other information**:
 
Needed to address https://github.com/reactiveui/ReactiveUI/issues/1615

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -32,6 +32,8 @@ if (string.IsNullOrWhiteSpace(target))
     target = "Default";
 }
 
+var includePrerelease = Argument("includePrerelease", false);
+
 //////////////////////////////////////////////////////////////////////
 // PREPARATION
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


kdubau:
I can imagine you folks might want to be able to enable this with either an argument or environment variable (current implementation provides no way to prefer Stable given a machine that also has Preview installed). If so, let me know the preference and I can add it.

glennawatson:
My preference would be a cake script parameter. Makes it easier for devs to turn on and off independently of their environment. We can modify what we pass to to the VSTS instance then.

glennawatson:
Oh btw, can we make that parameter turned off by default?

kdubau:
@glennawatson updated.

glennawatson:
Thanks @kdubau

PR #1729, opened 2018-08-23
Added Benchmark Harness (#1729)

### LINKED ISSUE #1276: ReactiveList creation performance

Neither, I guess it would be an improvement request.

Currently `ReactiveList<T>` is quite expensive to create in comparison to `ObservableCollection<T>` which is a problem in case you have thousands of them created at the same time. 

            var count = 20000;
            var rls = new List<ReactiveList<string>>();
            var ocs = new List<ObservableCollection<string>>();

            var sw = Stopwatch.StartNew();
            for (var i = 0; i < count; ++i)
            {
                rls.Add(new ReactiveList<string>());
            }
            sw.Stop();
            Console.WriteLine(sw.ElapsedMilliseconds);

            sw.Restart();
            for (var i = 0; i < count; ++i)
            {
                ocs.Add(new ObservableCollection<string>());
            }
            sw.Stop();
            Console.WriteLine(sw.ElapsedMilliseconds);

On my machine it takes about 1200ms to create ReactiveLists and 25ms to create ObservableCollections (that's almost 50 times longer).

It would be nice if creation time of ReactiveLists would be comparable to that of ObservableCollections.

I'd like to lessen the performance impact of creating ReactiveLists in my code. Currently it takes around 20% of overall execution time of that specific code to create them.

We're using ReactiveUI 7.1.0 on Windows for WPF desktop application development.

Performance profiling shows that the bottleneck sits in the `ReactiveList.setupRx` method and more specifically in this place (`Observable.Subscribe`, [Profiler Screenshot](https://www.screencast.com/t/EOMobga5b)):

    CountChanging.Where(_ => this.areChangeNotificationsEnabled()).Subscribe(_ => this.RaisePropertyChanging("Count"));
    CountChanged.Where(_ => this.areChangeNotificationsEnabled()).Subscribe(_ => this.RaisePropertyChanged("Count"));
    IsEmptyChanged.Where(_ => this.areChangeNotificationsEnabled()).Subscribe(_ => this.RaisePropertyChanged("IsEmpty"));
    Changing.Where(_ => this.areChangeNotificationsEnabled()).Subscribe(_ => this.RaisePropertyChanging("Item[]"));
    Changed.Where(_ => this.areChangeNotificationsEnabled()).Subscribe(_ => this.RaisePropertyChanged("Item[]"));

### LINKED ISSUE #1711: Create a benchmark project.
_We will close your issue without comment if you delete or do not fill out the issue checklist and provide ALL the requested information. This repository's issues are reserved for feature requests and bug reports. For support questions, please ask on StackOverflow: https://stackoverflow.com/questions/tagged/reactiveui_

_Please delete this note before pressing the submit button_

feature

Lets add a bench marking framework based on BenchmarkDotNet

So that we can get performance differences with new implementations of our code base. Focus more on getting something that performs well for end users.

RLittlesII:
- Added NavigationStack Benchmarks
- Added AutoPersist Benchmarks
- Added ReactiveList Operations Benchmarks
- Added RoutableViewModelMixin Benchmarks

Resolves #1711 
Resolves #1276

There are no performance tests

Adds a harness for ReactiveUI to execute performance benchmarks

**What might this PR break?**
Nothing.  This PR is additive.  Nothing in this PR affects the core solution.

### DIFF
--- a/benchmarks/ReactiveUI.Benchmarks/AutoPersistBenchmark.cs
+++ b/benchmarks/ReactiveUI.Benchmarks/AutoPersistBenchmark.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Reactive;
+using System.Reactive.Linq;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Attributes.Exporters;
+using BenchmarkDotNet.Attributes.Jobs;
+
+namespace ReactiveUI.Benchmarks
+{
+    [CoreJob]
+    [MarkdownExporterAttribute.GitHub]
+    public class AutoPersistBenchmark
+    {
+        private ReactiveList<string> _collection;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            _collection = new ReactiveList<string>(new[]
+            {
+                "ReactiveUI",
+                "ReactiveUI.XamForms",
+                "ReactiveUI.WPF",
+                "ReactiveUI.Events"
+            });
+        }
+
+        [Benchmark]
+        public void AutoPersistCollection() => _collection.AutoPersist(x => Observable.Return(Unit.Default), TimeSpan.FromMilliseconds(200));
+    }
+}
--- a/benchmarks/ReactiveUI.Benchmarks/Harness.cs
+++ b/benchmarks/ReactiveUI.Benchmarks/Harness.cs
@@ -1,0 +1,45 @@
+﻿using System;
+using BenchmarkDotNet.Running;
+using Xunit;
+
+namespace ReactiveUI.Benchmarks
+{
+    public class Harness
+    {
+        [Fact]
+        public void Auto_Persist_Benchmark()
+        {
+            BenchmarkRunner.Run<AutoPersistBenchmark>();
+        }
+
+        [Fact]
+        public void Create_Reactive_List_Benchmark()
+        {
+            BenchmarkRunner.Run<CreateReactiveListBenchmark>();
+        }
+
+        [Fact]
+        public void Reactive_List_Operation_Benchmark()
+        {
+            BenchmarkRunner.Run<ReactiveListOperationBenchmark>();
+        }
+
+        [Fact]
+        public void Navigation_Reactive_List_Benchmark()
+        {
+            BenchmarkRunner.Run<NavigationStackBenchmark>();
+        }
+
+        [Fact]
+        public void Routable_View_Model_Mixin_Benchmark()
+        {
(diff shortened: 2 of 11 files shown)


RLittlesII:
A few things to point out.

1. Some other projects that have implemented Benchmark DotNet have pulled their benchmarks into the main solution.  I elected to go a different route based on our current set of integration tests having separate solutions
2. xUnit is used to create the test harness.  We could have gone with just a dotnet core app, but I tend to prefer tests using test frameworks.
3. There are several output formats.  We should probably discuss what output format we want to use for the future of these types of tests.  [List of Bechmark exporters](https://benchmarkdotnet.org/articles/configs/exporters.html).
4. This PR does not add the benchmarks to the build process
5. If you find something that you know uses dynamic data that should be added before this PR is completed, call it out.

benchmarks/ReactiveUI.Benchmarks/AutoPersistBenchmark.cs
```diff
@@ -0,0 +1,31 @@
...
+            _collection = new ReactiveList<string>(new[]
+            {
+                "ReactiveUI",
+                "ReactiveUI.XamForms",
+                "ReactiveUI.WPF",
+                "ReactiveUI.Events"
+            });
+        }
+
+        [Benchmark]
```
glennawatson: I wonder if there is any penalty having the xunit fact in this file. I imagine that harness file getting pretty large

benchmarks/ReactiveUI.Benchmarks/Harness.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        [Fact]
+        public void Create_Reactive_List_Benchmark()
+        {
+            BenchmarkRunner.Run<CreateReactiveListBenchmark>();
+        }
+
+        [Fact]
+        public void Reactive_List_Operation_Benchmark()
```
glennawatson: Can you add my bench marks for the inpc property observable in the framework. In the pr I linked the other day.

RLittlesII: Added.

benchmarks/ReactiveUI.Benchmarks/AutoPersistBenchmark.cs
RLittlesII: Because it's xUnit and it should simply run all `[Fact]` in the project, we should be able to break up the harness files into smaller files in the future as needed.

Issue #1739, opened 2018-09-01
Add support for Avalonia properties in binding.


glennawatson:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
There is no support for Avalonia properties https://github.com/AvaloniaUI/Avalonia

**What is the expected behavior?**
To allow use of Avalonia's based properties.

**What is the motivation / use case for changing the behavior?**
To extend the number of users that can use RxBinding features.

worldbeater:
Just to mention, Avalonia team has an <a href="https://github.com/AvaloniaUI/Avalonia/tree/master/src/Avalonia.ReactiveUI">Avalonia.ReactiveUI</a> package published in NuGet, it would be great if we could transfer it to ReactiveUI organization, rename to `ReactiveUI.Avalonia` and add support for type-safe bindings.

glennawatson:
Want to open a issue on their website to see if they are willing?

worldbeater:
Great news. AvaloniaUI now supports WhenActivated! https://github.com/AvaloniaUI/Avalonia/pull/1951

RLittlesII:
@worldbeater Based on your assessment closing this issue.

PR #1748, opened 2018-09-14
[WIP] feature: Add Avalonia support

worldbeater:

This PR adds `WhenActivated` and <a href="https://reactiveui.net/docs/handbook/data-binding/">type-safe bindings</a> support for <a href="https://github.com/avaloniaui">AvaloniaUI</a> views. 

ReactiveUI doesn't fully support <a href="https://github.com/avaloniaui">AvaloniaUI</a> now.

ReactiveUI should fully support <a href="https://github.com/avaloniaui">AvaloniaUI</a> in the future. Closes https://github.com/reactiveui/ReactiveUI/issues/1739

**What might this PR break?**

Nothing.

**Todo:**

- [x] Implement `IActivationForViewFetcher` for Avalonia
- [x] Implement `ICreatesObservableForProperty` for Avalonia (removed, see https://github.com/reactiveui/ReactiveUI/pull/1748#issuecomment-427606487)
- [x] Implement `IWantsToRegisterStuff` and register all stuff there
- [x] Add tests for `IActivationForViewFetcher`
- [x] Add tests for `ICreatesObservableForProperty`
- [x] Add an Avalonia project to the <a href="https://github.com/reactiveui/ReactiveUI/tree/master/integrationtests">IntegrationTests</a> folder
- [ ] Create a `ReactiveUI.Avalonia` package (see https://github.com/AvaloniaUI/Avalonia/issues/1852)


src/.nuget/NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
   	<!--<add key="MyGet" value="https://www.myget.org/F/reactiveui/api/v3/index.json" />-->
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Avalonia MyGet" value="https://www.myget.org/F/avalonia-ci/api/v2" />
```
glennawatson: This won't be able to remain this way. You force any user to have to have a myget reference as well in their projects. Anything special you use in the beta not already in the nuget.org version? Eg if not you can just reference the latest nuget.org version with ="*" in the csproj, and you're not stopping people from using the myget anyway.

worldbeater: When I reference Avalonia `0.6.1` (latest on nuget), I get into a version hell and get exceptions like this one when I try to build the project: `System.IO.FileNotFoundException: Could not load file or assembly 'System.Reactive.Interfaces`. Latest beta is the only version where it *just worked*. So we may continue working on it like this until Avalonia team releases a new version on NuGet or until we figure out how to fix that version hell...

worldbeater:
Hmm, Avalonia doesn't seem to generate references to controls that have an `x:Name` attribute.

It seems to me the only thing we can do is something like this:
```cs
this.Bind(ViewModel,
    viewModel => viewModel.Title,
    view => view.FindControl<TextBox>("TitleTextBox").Text)
    .DisposeWith(subscriptions);
```

src/.nuget/NuGet.Config
glennawatson: You add a reference to System.Reactive.Compatibility to the ReactiveUI.Avalonia.csproj file.

glennawatson:
That will cause an exception what you have above btw. The ICreatesObservables you should be able to use the FindControl method. You may want to check that first then check for a Avalonia property?

glennawatson:
You could potentially also make two ICreatesObservable methods maybe? One for View Controls and the other for Avalonia Properties. Might be overkill but could separate out the logic a bit. Maybe given a higher waiting to if FindControl exists.

src/ReactiveUI.Avalonia/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
+using Avalonia.Controls;
+using Avalonia.Threading;
+using ReactiveUI;
+using Splat;
+
+namespace ReactiveUI.Avalonia
+{
+    public static class AppBuilderExtensions
+    {
+        public static TAppBuilder UseReactiveUI<TAppBuilder>(this TAppBuilder builder)
```
glennawatson: I suspect you can eliminate this by adding a IWantsToRegisterStuff() btw. 

To access as discussed in Slack you just need to add a entry to src/ReactiveUI/Properties/AssemblyInfo.cs file to this new assembly.

glennawatson:
BTW just to let you know, a big change is coming which will affect this PR, it will just auto-detect code style and force compliance. Wouldn't be too dramatic for this PR though.

worldbeater:
It turned out, that all <a href="https://github.com/AvaloniaUI/Avalonia/blob/b196f00bea3cd39d4514104df8491423f964a67e/src/Avalonia.Base/AvaloniaObject.cs#L23">AvaloniaObject</a>s implement the `INotifyPropertyChanged` interface:

```csharp
public class AvaloniaObject : IAvaloniaObject, IAvaloniaObjectDebug, INotifyPropertyChanged
```

This means we don't need any extra `ICreatesObservableForProperty` implementations. I'll remove the existing one with the next few commits, but will leave tests for `AvaloniaObject`s.

integrationtests/IntegrationTests.Shared/IntegrationTests.Shared.csproj
```diff
@@ -6,7 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="genesis.ensure" Version="2.0.0" />
-    <PackageReference Include="ReactiveUI" Version="*" />
+    <ProjectReference Include="..\..\src\ReactiveUI\ReactiveUI.csproj" />
```
worldbeater: I also don't like this change. We might want to wait until the new `ReactiveUI.Avalonia` package ships, and then reference `ReactiveUI *` and `ReactiveUI.Avalonia *` instead of playing with paths.

worldbeater:
I've just had a conversation with one of AvaloniaUI core maintainers who said that Avalonia API is extremely unstable now and will dramatically change over time, so they'd like to continue supporting `Avalonia.ReactiveUI` package and publishing instant fixes and updates when they need. So we should consider closing this PR and moving `WhenActivated` support to <a href="https://github.com/avaloniaui/avalonia">AvaloniaUI main repository</a>.

kekekeks:
We are heavily using ReactiveUI in our sample/demo apps that live in our main repo. Having them using an external nuget package that depends on Avalonia package would be really inconvenient.

PR #1837, opened 2018-11-12
ReactiveDemo hack around VS xaml designer issue with inherits from generic classes

Nilox:
Update ReactiveDemo sample

ReactiveDemo sample solution does not build. After adding missing dependencies, Views cannot be opened in designer.

In ReactiveDemo sample:
NugetDetailsView.xaml can be opened in Design view.
Fixes missing references to PresentationCore, PresentationFramework and WindowsBase.

**What might this PR break?**
Nothting. Just wonder now if missing dependencies are not actually missing. I mainly work in VS for Windows, but I don't know how the sample hs to be used to target e.g. Xamarin.

**Other information**:

### DIFF
--- a/samples/getting-started/ReactiveDemo/NugetDetailsView.xaml.cs
+++ b/samples/getting-started/ReactiveDemo/NugetDetailsView.xaml.cs
@@ -4,11 +4,14 @@ using System.Windows.Media.Imaging;
 
 namespace ReactiveDemo
 {
-    // The class derives off ReactiveUserControl which contains the ViewModel property.
+    /// Nongeneric class that allows us to edit NugetDetailsView in Design view
+    public class NugetDetailsViewBase : ReactiveUserControl<NugetDetailsViewModel> { }
+
+    // Second level derived class off ReactiveUserControl which contains the ViewModel property.
     // In our MainWindow when we register the ListBox with the collection of 
     // NugetDetailsViewModels if no ItemTemplate has been declared it will search for 
     // a class derived off IViewFor<NugetDetailsViewModel> and show that for the item.
-    public partial class NugetDetailsView : ReactiveUserControl<NugetDetailsViewModel>
+    public partial class NugetDetailsView : NugetDetailsViewBase
     {
         public NugetDetailsView()
         {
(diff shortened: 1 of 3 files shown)


samples/getting-started/ReactiveDemo/NugetDetailsView.xaml.cs
```diff
@@ -4,11 +4,14 @@
...
 {
-    // The class derives off ReactiveUserControl which contains the ViewModel property.
+    /// Nongeneric class that allows us to edit NugetDetailsView in Design view
+    public class NugetDetailsViewBase : ReactiveUserControl<NugetDetailsViewModel> { }
+
+    // Second level derived class off ReactiveUserControl which contains the ViewModel property.
     // In our MainWindow when we register the ListBox with the collection of 
     // NugetDetailsViewModels if no ItemTemplate has been declared it will search for 
     // a class derived off IViewFor<NugetDetailsViewModel> and show that for the item.
-    public partial class NugetDetailsView : ReactiveUserControl<NugetDetailsViewModel>
```
glennawatson: Where's this NugetDetailsViewBase come from. Seems redundant and unneccessary and the old approach was better.

glennawatson:
Hi @Nilox  thanks for the contribution. Due to being part of the .NET Foundation we require every contributor to do legal paperwork. Also the Reference updates to the csproj make sense, the deriving of the control the way you did it is not necessarily the standard way to use the ReactiveUserControl<T> so I would just remove that for the moment.

samples/getting-started/ReactiveDemo/NugetDetailsView.xaml.cs
Nilox: Fact is that NugetDetailsView.xaml cannot be edited in view designer. There is an issue in VS that does not support editing controls directly inherited from a generic class, so I proposed using a non generic class in between.
`public class NugetDetailsViewBase : ReactiveUserControl<NugetDetailsViewModel> { }`

`public partial class NugetDetailsView : NugetDetailsViewBase`
`{`
`    public NugetDetailsView()`
`    {...`
Other than that, I agree that the class is redundant from a functional point of view. However it is a matter of taste which approach is better - for a beginner, a MVVM framework with a sample project in which "a View cannot be viewed" may not look very promising.

Nevertheless, I have removed the intermediate class as requested.

glennawatson:
That designer is so flakey. Bain of a lot of people's existance.

Issue #1853, opened 2018-12-06
Cannot create an instance of "ReactiveUserControl`1"


M0n7y5:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
XAML Designer not working 
![obrazek](https://user-images.githubusercontent.com/17201053/49590126-5359a180-f96b-11e8-8d74-fe340dfa0f32.png)

But compilation work. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Simply create Reactive User Control

**What is the expected behavior?**

Working XAML Designer

**What is the motivation / use case for changing the behavior?**

...

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the [latest stable](https://www.nuget.org/profiles/reactiveui) and [development snapshot](https://reactiveui.net/docs/getting-started/installation/nuget-packages/development-package-feed)**

ReactiveUI 9.4.1
Visual Studio 2017 CE 15.9.3
Net Framework 4.7.2

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**
Control XAML

```XAML
<reactiveui:ReactiveUserControl x:Class="WavinModbusTester.Views.TestControl" 
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"          
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
        xmlns:reactiveui="http://reactiveui.net"
        xmlns:viewmodel="clr-namespace:WavinModbusTester.ViewModels" 
        x:TypeArguments="viewmodel:TestControlViewModel"
        mc:Ignorable="d" 
        d:DesignHeight="450" d:DesignWidth="800">
    <Grid>

    </Grid>
</reactiveui:ReactiveUserControl>
```
Control CS

```cs
using ReactiveUI;
using WavinModbusTester.ViewModels;

namespace WavinModbusTester.Views
{
    /// <summary>
    /// Interaction logic for TestControl.xaml
    /// </summary>
    public partial class TestControl : ReactiveUserControl<TestControlViewModel>
    {
        public TestControl()
        {
            InitializeComponent();
        }
    }
}
```
Control ViewModel

```cs
using ReactiveUI;

[...]

PR #1855, opened 2018-12-10
Add documentation explaining how to use ReactiveUserControl in XAML

jorisvergeer:
Add documentation explaining how to use ReactiveUserControl in XAML in UWP projects and the WPF designer window.

docs update

https://github.com/reactiveui/ReactiveUI/issues/1853

### DIFF
--- a/src/ReactiveUI/Platforms/windows-common/ReactiveUserControl.cs
+++ b/src/ReactiveUI/Platforms/windows-common/ReactiveUserControl.cs
@@ -36,6 +36,33 @@ namespace ReactiveUI
     /// ]]>
     /// </code>
     /// </para>
+    /// <para>
+    /// Note that UWP projects do not support the <c>TypeArguments</c> attribute. The XAML designer window in WPF projects also do not
+    /// support generic types. To use <see cref="ReactiveUserControl{TViewModel}"/> in XAML documents you need to create a base class
+    /// where you derive from <see cref="ReactiveUserControl{TViewModel}"/> with the type argument filled in.
+    /// <code>
+    /// <![CDATA[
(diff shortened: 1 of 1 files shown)


src/ReactiveUI/Platforms/windows-common/ReactiveUserControl.cs
```diff
@@ -36,6 +36,33 @@ namespace ReactiveUI
     /// ]]>
     /// </code>
     /// </para>
+    /// <para>
+    /// Note that UWP projects do not support the <c>TypeArguments</c> attribute. The XAML designer window in WPF projects also do not
```
glennawatson: Change "do not" to "does not"

src/ReactiveUI/Platforms/windows-common/ReactiveUserControl.cs
```diff
@@ -36,6 +36,33 @@ namespace ReactiveUI
...
+    /// support generic types. To use <see cref="ReactiveUserControl{TViewModel}"/> in XAML documents you need to create a base class
```
glennawatson: You don't have to do it this way. The xaml designer is do fragile I gave up on it years ago. So might be worth saying more like "to use the base class"

jorisvergeer: It doesn't compile in an UWP (it does in WPF) project when you use ReactiveUserControl. The xaml designer is an extra for me.

src/ReactiveUI/Platforms/windows-common/ReactiveUserControl.cs
jorisvergeer: You're right

PR #1870, opened 2018-12-27
housekeeping: remove code analyzers from base directory. (#1870)

glennawatson:


glennawatson:
Basically, the theory being the /src folders will have their own settings, and integration tests/samples will have their own. Eg the samples have some by other members eg @giusepe who may have different coding standards.

RLittlesII:
Shouldn't you remove the `CodeAnalysisRuleSet ` xml tag earlier in this file?

PR #1877, opened 2018-12-30
housekeeping: update msbuild.sdk.extras 1.6.65, and eventbuilder updates (#1877)

glennawatson:
- Now supports Tizen 4 types, separated into Tizen 4 so we can support Tizen 5 in the future.
- NuGet is updated with the latest framework
- MsBuild.Sdk.Extras is updated, allowing the latest Tizen NugetPackages

### DIFF
--- a/src/EventBuilder/Program.cs
+++ b/src/EventBuilder/Program.cs
@@ -4,14 +4,17 @@
 
 using System;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading.Tasks;
+using CommandLine;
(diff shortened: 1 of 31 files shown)


src/EventBuilder/Program.cs
```diff
@@ -21,133 +24,128 @@ internal static class Program
...
+                            Debugger.Break();
+                        }
                     }
-                }
-            }
 
-            Environment.Exit((int)ExitCode.Error);
+                    Environment.Exit((int)ExitCode.Error);
+                },
+                _ => Task.CompletedTask).ConfigureAwait(false);
```
timotei: Really? Ignoring the errors? :(

PR #1919, opened 2019-01-18
housekeeping: Add ReactiveCommand usage examples

worldbeater:

This PR adds usage examples to [ReactiveCommand](https://reactiveui.net/api/reactiveui/reactivecommand/) API docs. Addresses https://github.com/reactiveui/rfcs/issues/5#issuecomment-389593613

There are no examples in our API docs showing up how to create and execute a command.

**What is the new behavior?**

Examples are present.


giusepe:
Hey @worldbeater!

Thank you for one more PR :)

One question, instead of using `Execute()` that doesn't respect the `canExecute`, shouldn't we use an `Observable.Return(42).InvokeCommand(command)` in the samples as a way to reinforce it as a best practice?

PR #1932, opened 2019-01-26
housekeeping: Make ReactiveUI compile on VisualStudio for Mac

glennawatson:
* Modified the Events projects to handle compiling on mac
* Android events now look for desired version on both windows/mac
* Modified the azure to run tests/ci on both windows/mac
* If not compiling on windows exclude uwp/net461
* Fix the build.sh to use the same as what the cake build team uses, and uses a global tool
* Turn off sourcelink for mac building temporarily. We don't use it for generating the NuGet packages anyway
* Modified Fody to not rely on TargetFramework for where to copy files from.
* Also fixed Fody compilation when it copies from the helper project, to wait until the sub-projects have a chance to compile first.


src/ReactiveUI.Fody.Tests/ReactiveUI.Fody.Tests.csproj
```diff
@@ -1,26 +1,33 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
   <PropertyGroup>    
-    <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);net461</TargetFrameworks>
+    <FodyTargetFramework>netstandard2.0</FodyTargetFramework>
+    <FodyTargetFramework Condition=" $(TargetFramework.StartsWith('net4')) ">$(TargetFramework)</FodyTargetFramework>
```
RLittlesII: Should this be `$(FodyTargetFramework)` ?

glennawatson: No. In this case I want net461 (or whatever future versions we use)

PR #1951, opened 2019-02-14
(wip) housekeeping: Update EventBuilder to use System.Reflection.Metadata

glennawatson:
Rewrite of the EventBuilder to start using System.Reflection.Metadata.

Has dramatically reduced time to generate now. 

More modular facing in the mind of making it focused on allowing it to be consumed by other libraries.


worldbeater:
Sorry for marking this as ready. Saw a button saying something like "Review" and pressed it with a hope to look though the changes to see how this is implemented, but it appears that I marked this PR as ready for review... 🌚

glennawatson:
Yeah. I specifically opened this PR to try that feature too since it's a little early.

src/EventBuilder/Templates/StaticTemplate.mustache
```diff
@@ -0,0 +1,19 @@
+﻿
+{{#Namespaces}}
+namespace {{Name}}
+{
+    public static class Events
```
RLittlesII: So I noticed while using the ReactiveUI.Events.XamEssentials package that the static class that allows access to the Events is `Events`.  I am thinking it should be a bit more specific to the domain it is surfacing events for.  I know this would be a breaking change for this library, but it opens us up for further Events in the future.

PR #1953, opened 2019-02-16
housekeeping: Make ReactiveUI use the new ReactiveUI.Cake.Recipe

glennawatson:
To keep the builds consistent between all projects, we are now using a common cake recipe.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,11 +12,12 @@ jobs:
 - job: Windows
   pool:
     vmImage: vs2017-win2016
+
   steps:
   - task: DotNetCoreInstaller@0
     displayName: Install Dot Net Core v2.2.1
     inputs:
-      version: '2.2.103'
+        version: '2.2.103'
 
   - task: BatchScript@1
     inputs:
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -28,34 +29,61 @@ jobs:
   - powershell: .\build.ps1
     displayName: Build 
     env:
-      SIGNCLIENT_SECRET: $(SignClientSecret)
-      SIGNCLIENT_USER: $(SignClientUser)
-      ArtifactDirectory: $(System.DefaultWorkingDirectory)\artifacts
+       SIGNCLIENT_SECRET: $(SignClientSecret)
+       SIGNCLIENT_USER: $(SignClientUser)
+       CODECOV_TOKEN: $(CODECOV_TOKEN)
+       ArtifactDirectory: $(System.DefaultWorkingDirectory)\Artifacts
```
RLittlesII: ArtifactsDirectory?

glennawatson: That variable isn't used but I'm going to swap over to using these soon so this file will be dramatically changing as well https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops

glennawatson:
the change in codecov is due to the fact we use a Release build without optimisation now. (use Release so we don't have #debug stuff happening but no optimisations so we get full code coverage)

PR #1956, opened 2019-02-21
feature: Add winforms auto binding to Tables/Flow based controls

glennawatson:
The following has been added:
* A interface called `ISetMethodBindingConverter` which allows the bind engine to perform a custom set operation. This is because WinForms often gives you collections where you are meant to do 'Add' operations directly rather than setting them.
* Added two new implementations of the `ISetMethodBindingConverter`. One for general ContentControls, the other for tables.
* Added a winforms demo based on the main sample.


axgdev:
Hi there! Thank you for the work on ReactiveUI it is great! :)

This commit unfortunately breaks some functionality. If a property value of a binding was null, the framework could (in v9.10.7) cleanly convert it into something appropriate. Example:
```csharp
this.WhenActivated(disposable =>
{
    this.WhenAnyValue(x => x.FormText)
        .BindTo(this, x => x.Text)
        .DisposeWith(disposable);
});
```

Please not that `this` refers to a Form. If for some reason the property `FormText` is `null` then an exception is thrown like:
 
PropertyBinderImplementation: x.Text Binding received an Exception!: System.NullReferenceException: Object reference not set to an instance of an object.
   at ReactiveUI.PropertyBinderImplementation.<>c__DisplayClass10_0`3.<BindToDirect>g__SetThenGet|0(Object paramTarget, Object paramValue, Object[] paramParams) in somepath\src\ReactiveUI\Bindings\Property\PropertyBinderImplementation.cs:line 459
   at ReactiveUI.PropertyBinderImplementation.<>c__DisplayClass10_0`3.<BindToDirect>b__1(TObs x) in somepath\src\ReactiveUI\Bindings\Property\PropertyBinderImplementation.cs:line 475
   at System.Reactive.Linq.ObservableImpl.Select`2.Selector._.OnNext(TSource value) in somepath\src\System.Reactive\Linq\Observable\Select.cs:line 39

glennawatson:
It's worth probably taking this sort of feedback straight to an issue btw. Closed PR's have a habit of getting lost. Moved it over to one ;)

PR #1988, opened 2019-03-31
housekeeping: Update to UAP 10.0.17763.0

glennawatson:
Update to UAP 10.0.17763.0
Now using the generic conditions for the csproj.
Update various nuget packages in the Directory.Build.Props

This allows us to compile under visual studio 2019 on azure devops.


worldbeater:
Feels like analyzers changed the behavior of `CA2213` check:
```
'ScheduledSubject<T>' contains field '_defaultObserverSub' that is of IDisposable type 'IDisposable', but it is never disposed. Change the Dispose method on 'ScheduledSubject<T>' to call Close or Dispose on this field.
```
Probably worth changing action to `None` or `Warning`:
https://github.com/reactiveui/ReactiveUI/blob/eb53eee4bd1865b3a226ec468c0a344620349bee/src/analyzers.ruleset#L63

glennawatson:
Seems to be a legitimate bug tbh so will likely fix the warnings/errors. I think before they were just checking you had a dispose method now they check to make sure you dispose the contents

PR #1999, opened 2019-04-11
feature: re-add dot net core 3.0 wpf/winforms support

worldbeater:

Re-apply changes from https://github.com/reactiveui/ReactiveUI/commit/17558daa7983087d3626bdc9c66941987c925499 missed due to `git lfs` conversion, which caused `master` force push

image

**What is the new behavior?**

Should be only one commit ahead and none commits behind master.

**What might this PR break?**

Need to go through all the files and check if it compiles on Azure Pipelines. 
This is an early draft, please do not merge.


worldbeater:
Temporary changed base branch to `master` to avoid those 3K commits included in diff.
Need to change base branch to `preview/10.0.0` if we are going to merge this one.

PR #2007, opened 2019-04-23
Update Program.cs

timotei:
Improvement fix :D

No error logged on errors.

**What is the new behavior?**
Error logged on errors.

**What might this PR break?**
Nothing.

**Other information**:

### DIFF
--- a/src/EventBuilder/Program.cs
+++ b/src/EventBuilder/Program.cs
@@ -141,7 +141,24 @@ namespace EventBuilder
 
                     Environment.Exit((int)ExitCode.Error);
                 },
-                _ => Task.CompletedTask).ConfigureAwait(false);
+                errors =>
+                {
+                    Log.Fatal("Had these errors parsing the command line:");
+                    foreach (var error in errors)
+                    {
+                        switch (error)
+                        {
+                            case MissingRequiredOptionError opt:
+                                Log.Fatal($"Missing required parameter: {opt.NameInfo.NameText}");
+                                break;
+                            default:
+                                Log.Fatal(error.ToString());
+                                break;
+                        }
+                    }
+
+                    return Task.CompletedTask;
+                }).ConfigureAwait(false);
         }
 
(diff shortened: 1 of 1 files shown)


glennawatson:
Thanks for the pr but event builder is changing soon. So will be closing this one.

glennawatson:
It's basically getting a full rewrite so will leave as is for the moment. There's also some splat extensions that can an make that logging easier.

timotei:
@glennawatson Hmm, I think would be awesome if you could expose the eventbuilder somehow (prebuilt binary), as a nuget package for example. We want to generate the events for some DevExpress libraries but for that we have to maintain our own fork, instead of just using an existing exe (https://github.com/gluck/ReactiveUI/blob/DevExpress/ReactiveUI.Events/ReactiveUI.Events.Dx15.Xpf_Net40.csproj)

glennawatson:
Yeah that's the direction I am going. You can specify nuget or assembly to generate events for.

glennawatson:
Part of the motivation for the rewrite we had users who were ahead on xamarin forms so their events weren't matching. So by rewriting and allowing it to be a msbuild task or global tool we can generate events for specific versions. 

I had to speed up the generation to do it so using Roslyn and system.reflection.metadata now

timotei:
@glennawatson Btw, how about renaming the .events packages to something more proper, that does *not* include "ReactiveUI"? Cause technically, apart from the repo, they don't use anything reactiveui. 

Maybe they should be renamed to just "Reactive.Events.X". What do you think?

glennawatson:
Worth asking about that one on slack under the #reactiveui-dev channel. 

Generally I shy away from package renames since there is quite a bit of work involved.

glennawatson:
https://github.com/reactiveui/pharmacist is going to be released in the next 24 hours. It's the replacement for event builder.

glennawatson:
So for the devexpress libraries, you can probably just use Pharmacist.MsBuild now, similar to what we do with Xamarin.Essentials -- I'm integrating some of @gluck performance improvements he tried to implement about 2-3 years ago now.

Issue #2013, opened 2019-04-28
[BUG] exception swallowed in RaiseAndSetIfChanged


philippdoersek:
**Describe the bug**

Exceptions thrown in an OnNext handler attached to an observable created via WhenAnyValue and RaiseAndSetIfChanged are swallowed and don't propagate, in contrast to the documentation of the default exception handler (https://reactiveui.net/docs/handbook/default-exception-handler/).

**Steps To Reproduce**

RoslynPad reproduction case:
```c#
#r "nuget:ReactiveUI/9.13.1"

using System.Reactive.Linq;
using ReactiveUI;

class ReactiveExample : ReactiveObject
{
    public ReactiveExample()
    {
        var o = this.WhenAnyValue(x => x.S).Skip(1);
        o.Subscribe(s => 
        {
            Console.WriteLine("received value " + s);
            throw new Exception("throw on value " + s);
        });
        //this.ThrownExceptions.Subscribe(e => throw new Exception("uncaught", e));
    }
    
    public string S
    {
        get => _s; 
        set => this.RaiseAndSetIfChanged(ref _s, value);
    }
    
    private string _s = "";
}

var r = new ReactiveExample();
r.S = "bar";
```

**Expected behavior**

The exception thrown in the above code should propagate.

**Screenshots**

N/A

**Environment**

- OS:       Windows
- Version   10
- Device:   PC

**Additional context**

I believe this is caused by ReactiveUI.IReactiveObjectExtensions.NotifyObservable, which doesn't rethrow the exception in case ThrownExceptions isn't subscribed.

glennawatson:
As per your stack overflow question some additional details where to track it down. 

I believe that this is caused by the try/catch in ReactiveUI.IReactiveObjectExtensions.NotifyObservable, which I would have expected to rethrow the exception in case the ThrownExceptions observable doesn't have a subscriber instead of just logging it (see https://github.com/reactiveui/ReactiveUI/blob/a4ee168dade8b5e3f34337fabd56eca10eca5200/src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs from line 382).

glennawatson:
Not sure when we can get to it, hopefully soon. Everyone on the project has a lot of work related pressure at the moemnt.

PR #2035, opened 2019-05-17
fix: Rethrow exception on RaiseAndSetIfChanged

RLittlesII:

Resolves: #2013 

Exceptions thrown on `RaiseAndSetIfChanged` are not being rethrown if there is no "ThrownExceptions" subscribed.

**What is the new behavior?**

Exceptions thrown on `RaiseAndSetIfChanged` are rethrown if there is no "ThrownExceptions" observable subscriptions.

**What might this PR break?**

### DIFF
--- a/src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
+++ b/src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
@@ -391,6 +391,7 @@ namespace ReactiveUI
                     if (_thrownExceptions.IsValueCreated)
                     {
                         _thrownExceptions.Value.OnNext(ex);
+                        return;
                     }
 
                     throw;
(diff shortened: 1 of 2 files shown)


RLittlesII:
@glennawatson Seems that adding the code for this broke a preexisting unit test.  Seems that fixing this is in direct violation with how we are marshaling exceptions.  It may require additional conversation.

src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
```diff
@@ -392,6 +392,8 @@ internal void NotifyObservable<T>(TSender rxObj, T item, ISubject<T> subject)
                     {
                         _thrownExceptions.Value.OnNext(ex);
                     }
+
```
glennawatson: I would only throw this if there is no _thrownExceptions.IsValueCreated

PR #2051, opened 2019-05-26
Feature:  ReactiveRecyclerViewAdapter.GetItemViewType now passed view model type

clintonrocksmith:
Feature

The GetItemViewType is standard with only a position ID, which cannot be used to work out what item belongs in the sourceList.

**What is the new behavior?**
The GetItemViewType addition, now bubbles up the current position as well as the ViewModel so that the cell view type can be worked out.  This is important following best practices when a RecyclerView has different ViewHolders inside it.  With different View Cells.

**What might this PR break?**
Unsure

**Other information**:

### DIFF
--- a/src/ReactiveUI.AndroidSupport/ReactiveRecyclerViewAdapter.cs
+++ b/src/ReactiveUI.AndroidSupport/ReactiveRecyclerViewAdapter.cs
@@ -90,7 +90,7 @@ namespace ReactiveUI.AndroidSupport
 
         private TViewModel GetViewModelByPosition(int position)
         {
-            return _list.Count > position ? null : _list.Items.ElementAt(position);
+            return position > _list.Count ? null : _list.Items.ElementAt(position);
         }
 
         private void UpdateBindings(Change<TViewModel> change)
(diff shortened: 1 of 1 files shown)


src/ReactiveUI.AndroidSupport/ReactiveRecyclerViewAdapter.cs
```diff
@@ -90,7 +90,7 @@ protected override void Dispose(bool disposing)
 
         private TViewModel GetViewModelByPosition(int position)
         {
-            return _list.Count > position ? null : _list.Items.ElementAt(position);
+            return position > _list.Count ? null : _list.Items.ElementAt(position);
```
worldbeater: Won't this throw `IndexOutOfBoundsException` here? E.g. in case when `position` equals to `_list.Count`. Probably worth changing `>`  to `>=` and covering this stuff with some unit tests if possible to verify such cases

glennawatson: Hard to do unit tests at the moment with Android. Since we are limited to azure DevOps. There is a task for adding them with tutorial how to achieve it.

clintonrocksmith: Position is zero based whereas count starts from 1, if it hits null then there is some other problem
The usual code is _list.Items.ElemantAt(position) without the check.  At least it returns an Empty ViewModel for the implementation.  I might have to re-think how this works though.  Hmmmmm  fresh eyes.

ReacvityRecyclerViewAdapter cannot be used directly, it needs to be used in-directly

Thanks for the comment.

Issue #590, opened 2014-05-11
A modest proposal: Let users initialize RxUI.


jlaanstra:
### RxUI 2.4

Service location was first added in commit https://github.com/reactiveui/ReactiveUI/commit/38ba9c3eaf8cdf33f8e77a14d052ea66b8ac67d1, to locate views for RxUI.Routing.
### RxUI 3.0

Service location moved into `RxApp` in commit https://github.com/reactiveui/ReactiveUI/commit/783a008f00eb4516b53df94e0c018410b4ee282f. That commit removed the dependency on the common service locator. This most likely happened because the CSL was too limited.
### RxUI 4.0

The service location infrastructure added in RxUI 3.0 gets a little smarter in commit https://github.com/reactiveui/ReactiveUI/commit/f7bf19a3453b264dfa9b83363a82984602107262. It can now service as a very simple locator for RxUI internally while still allowing an external locator to be configured.

In RxUI 4 we also started to call into the `ServiceLocationRegistration` class in RxUI.Xaml via reflection to register stuff (https://github.com/reactiveui/ReactiveUI/commit/296c7a4789c679c8665e2c0bae8ad06a2e45ebdb). This class got more and more responsibilities such as getting the Dispatcher scheduler (https://github.com/reactiveui/ReactiveUI/commit/56346913fff42162727d125630ebcba3b1a013dc) and command bindings.

This registration stuff was expanded to check all RxUI dlls and load them early in commit https://github.com/reactiveui/ReactiveUI/commit/e088a9af6d519644f43edde6b09ba6322e4a188e. 
The registration stuff was fragily because especially Windows Phone and Windows Store apps were not able to detect all assemblies and load them. Therefore instead of loading assemblies we switched to do `Type.GetType()` with the fully qualified type name (https://github.com/reactiveui/ReactiveUI/commit/19c450e3034b0993ea08079617ec264ae6901368). This broke a lot of times, for example in commit (https://github.com/reactiveui/ReactiveUI/commit/5d40c60322a932bd9ccf11877e3da9b09a6fcf5e).
### RxUI 5.0

[...]

PR #2067, opened 2019-06-12
feature: Added Uno support

weitzhandler:
This PR adds initial support for the Uno Platform.


src/ReactiveUI.Wpf/obj/Debug/net461/GeneratedInternalTypeHelper.g.cs
```diff
@@ -0,0 +1,62 @@
+﻿//------------------------------------------------------------------------------
```
worldbeater: Looks like this `.g.cs` file was included here accidentally ¯ \ _ (ツ) _ / ¯
Probably worth removing?

weitzhandler:
Indeed. Either remove or I'll do tmr

src/ReactiveUI.Wpf/obj/Debug/net461/GeneratedInternalTypeHelper.g.cs
glennawatson: I'm surprised the file got included since it wouldn't pass the git ignore file.

src/ReactiveUI.Uno/GlobalSuppressions.cs
```diff
@@ -0,0 +1,10 @@
+﻿
+// This file is used by Code Analysis to maintain SuppressMessage 
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given 
+// a specific target and scoped to a namespace, type, member, etc.
+
```
glennawatson: Let's not use a global suppression file, instead move to just attributes above the particular targets.

Eg
```
[SupressMessage("Design", "CA1010:Collections should implement generic interface", Justification = "Deliberate usage")]
public class ReactivePage<T>
```

weitzhandler:
Why is my build failing?

src/ReactiveUI.Uno/GlobalSuppressions.cs
weitzhandler: Done.

weitzhandler:
Heads up, looks like generic type arguments are not supported.
So what I suggest is having a property in `ReactiveUserControl` instead, on the UWP XAML based platforms.
Does this make sense?

glennawatson:
https://github.com/reactiveui/ReactiveUI/blob/e6e8b3cdc7060b9cc44d37aaf3e6530283ab09aa/src/ReactiveUI/Platforms/windows-common/ReactiveUserControl.cs#L45 that's the current proposal we have for users in our docs.

weitzhandler:
@jeromelaban can u take a look around here?

src/ReactiveUI.Uno/ReactiveUI.Uno.csproj
```diff
@@ -0,0 +1,33 @@
+﻿<Project Sdk="MSBuild.Sdk.Extras">
+  <PropertyGroup>
+    <TargetFrameworks>xamarinmac20;MonoAndroid90;xamarinios10;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);uap10.0.17763</TargetFrameworks>
+    <PackageId>ReactiveUI.Uno</PackageId>
+    <Title></Title>
+    <Description>Uno-Platform specific extensions to ReactiveUI</Description>
```
jeromelaban: Is it _for ReactiveUI_ ? My english may be at fault :)

src/ReactiveUI.Uno/ReactiveUI.Uno.csproj
```diff
@@ -0,0 +1,33 @@
...
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+  </PropertyGroup>  
+  <PropertyGroup>
+    <DefineConstants>HAS_UNO</DefineConstants>
```
jeromelaban: This is automatically defined by the Uno package. See here:

https://github.com/nventive/Uno/blob/d9d01e7a06f289d1188807573d3ce5487db0a35d/build/uno.ui.targets#L23

src/ReactiveUI.Uno/ReactiveUI.Uno.csproj
ghuntley: It should be _for_ ReactiveUI

jeromelaban: Also, we're using "Uno Platform" without the dash.

weitzhandler: I see you resolved this @ghuntley, thank you so much!
was actually doubting if i should cc u cuz i didn't wanna bother
you guys rock 👑

src/ReactiveUI.Uno/ReactiveUI.Uno.csproj
weitzhandler: Left it in Debug

weitzhandler:
@glennawatson Perhaps revert this commit, I opened #2078 for that.

src/ReactiveUI/Mixins/DependencyResolverMixins.cs
```diff
@@ -31,7 +31,8 @@ public static void InitializeReactiveUI(this IMutableDependencyResolver resolver
             {
                 "ReactiveUI.XamForms",
                 "ReactiveUI.Winforms",
-                "ReactiveUI.Wpf"
+                "ReactiveUI.Wpf",
+                "ReactiveUI.Uno"
```
glennawatson: This will likely have to change in the near future based on talking to others.

weitzhandler: IMHO we should rather make a `Bootstrapper` that inherits `Application` and should be used to initialize app, and provide virtual initialization and registration stuff. Static is evil because it's sealed. That's why I am not the biggest fun of `Locator.Current`. Event it's mobile, we should improve DI performance, not use static instead. Just my opinion.
In Prism, there is an abstract service-locator, that Unity, DryIoc etc. are all ported to, and the bootstrapper can in essence be agnostic of it. We also need to make it easier to bring in modules and external V/VM assemblies and provide them with the container for type registration.

Anyway what else do we need to get this PR going so we can start using ReactiveUI in Uno apps off NuGet?

README.md
```diff
@@ -79,6 +79,7 @@ Install the following packages to start building your own ReactiveUI app. <b>Not
 [TestDoc]: https://reactiveui.net/docs/handbook/testing/
 
 [UniDoc]: https://reactiveui.net/docs/getting-started/installation/universal-windows-platform
+[UnoDoc]: https://reactiveui.net/docs/getting-started/installation/universal-windows-platform
```
glennawatson: Remove please, until we are ready with actual docs.

src/ReactiveUI/Mixins/DependencyResolverMixins.cs
weitzhandler: Can I leave this one in?

glennawatson:
Leave it in, I'm going to check some stuff the blazer guys did which would affect that portion of the code.

src/ReactiveUI/Mixins/DependencyResolverMixins.cs
weitzhandler: I see this now: https://github.com/reactiveui/ReactiveUI/issues/590.

src/ReactiveUI.sln
```diff
@@ -1,6 +1,7 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26730.15
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
```
RLittlesII: Do we want to change the Visual Studio version?

glennawatson: I think it's fine yeah. We are going down the route of requiring it anyway for maintainers.

weitzhandler:
Thank you guys for the hard work, you're amazing!

PR #2080, opened 2019-06-18
Fix: Restore case-insensitive WireUpControls feature

### LINKED ISSUE #2061: Controls not Wiring Up in Android
Updated to the latest on android (9.16.6), and it is no longer wiring up controls that are camel cased in the axml, but are upper cased in the class. If I change the axml to match the class, it wires up.

Example:
android:id="@+id/authPassword" used to map to "AuthPassword" when I call WireUpControls(), but will now not map unless I change to android:id="@+id/AuthPassword".

cabauman:

Fixes #2061

Not wiring up controls that are camel cased in the axml, but are upper cased in the class. If the axml is changed to match the class, it wires up.

**What is the new behavior?**

Wiring up controls won't depend on property case.

**What might this PR break?**
Anyone calling the View overload of GetControl because the signature changed.
Anyone calling the Fragment overload of GetControl because it was removed.

**Other information**:


src/ReactiveUI.AndroidSupport/ControlFetcherMixin.cs
```diff
@@ -20,108 +15,20 @@ namespace ReactiveUI.AndroidSupport
     /// Fragments via property names, similar to Butter Knife, as well as allows
     /// you to fetch controls manually.
     /// </summary>
-    public static class ControlFetcherMixin
+   public static class ControlFetcherMixin
     {
-        private static readonly ConditionalWeakTable<object, Dictionary<string, View>> viewCache =
-            new ConditionalWeakTable<object, Dictionary<string, View>>();
-
         static ControlFetcherMixin()
```
glennawatson: Do we need the empty constructor?

cabauman: No, I can remove it.

PR #2097, opened 2019-06-28
feature: add a UAP target in ReactiveUI.Uno with namespace differences.

weitzhandler:
A user friendly feature.

Allow the users to use the ReactiveUI.Uno namespace for UAP namespaces if they include it in. This will allow UAP users to use the same XAML files both all platforms.


glennawatson:
Closing, going to reopen with some more fixes.

Issue #2092, opened 2019-06-27
Let ReactiveUI RxApp handle multiple windows


glennawatson:
On UWP each new window gets its own ui thread and the way `WaitForDispatcherScheduler` (which is the default scheduler for `RxApp.MainThreadScheduler`) works it is inherently only going to work on a single window app. As @anaisbetts put it in Slack, UWP UI thread situation is a "tire fire" and to quote "ReactiveUI is not designed to handle multiple window apps" - which essentially means no `ProjectionManager`, no `ShareTarget`, no `ApplicationViewSwitcher`, nothing multi-window. While I agree with the first part (and begrudgingly the second part), I am not so sure that RxUI shouldn't be considering these issues.

This can cause issues with RxApp.MainThreadScheduler, where each one of these windows runs its own dispatcher thread and is an “Application Single Threaded Apartment” (ASTA). Essentially the framework is blocking what it thinks is a potential deadlock. This particular exception is thrown when code running under one STA (1) calls out to another STA (2) and that STA (2) might have to re-enter STA (1) in order to do its work. The framework throws that exception at this point to try and stop you getting into a mess with deadlocks.

I am currently testing a scheduler that can handle multiple windows/dispatchers and will hopefully manage this as well, but we are still testing it and I'm not sure if I'll be able to block this exception yet (which would make my new scheduler useless... 👎). 

_Originally posted by @hinterlandsupplyco in https://github.com/reactiveui/ReactiveUI/issues/2034#issuecomment-504558031_

glennawatson:
Fixed by #2100

Issue #2034, opened 2019-05-16
[BUG] ReactiveCommand throws exception in UWP app when activated through sharing


harvinders:
**Describe the bug**
In a UWP app, if the app is activated through [receive data api](https://docs.microsoft.com/en-us/windows/uwp/app-to-app/receive-data) the ReactiveCommand invocation raises exception 

**Steps To Reproduce**
I am providing a bare minimum app that reproduces this problem. 
[ReactiveUITestApp1 code](https://github.com/harvinders/ReactiveUITestApp1)

1. Start the Debugging the app in visual studio
2. Using firefox or edge browser, click on the share menu
3. Select ReactiveUITestApp1 from the list of apps
4. From the new application window, click the `Complete` button
5. After the command's `Action` is executed an exception is thrown
 
**Expected behavior**
There should not be an exception or there should be a way to safely invoke the command

**Environment**

- OS:       Windows 10
- Version   1903 

**Additional context**
The project target was set to 1903 with min target of 1809. The ReactiveUI package 9.15.1 was used.

If I remove the command creation in MainViewModel in the main UI of the app, I do not get the exception.

Others have faced similar exception whilst sharing/receiving in UWP, here are some links to provide some pointers why it might be happening

https://stackoverflow.com/a/18643977/1139856
https://github.com/luberda-molinet/FFImageLoading/issues/166

Exception Stacktrace

  | Name | Value | Type
-- | -- | -- | --
[...]

Issue #1221, opened 2017-01-05
ReactiveUI crashes on initialization in a UWP app


pingzing:
**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
ReactiveUI 7.0 crashes on application start with a TypeInitializationException in UWP apps.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Create a new, blank UWP app, and add ReactiveUI  7.0 via NuGet. Add some element to your code in, say, App.xaml that makes use of Reactive UI, like `private ReactiveList<string> test = new ReactiveList<string>();`.

Observe crash.

**What is the expected behavior?**
No crashes.

**What is the motivation / use case for changing the behavior?**
Crashes are bad =)

**Which versions of ReactiveUI, and which platform / OS are affected by this issue? Did this work in previous versions of ReativeUI? Please also test with the latest stable and snapshot (http://docs.reactiveui.net/en/contributing/snapshot/index.html) versions.**
ReactiveUI 7.0, on Windows 10 Anniversary Update, creating a UWP app. Presumably it worked in previous versions, as I see references to people using it for UWP in 6.5.2.

I see the same issue in the latest snapshot (7.1.0-alpha0003).

**Other information (e.g. stacktraces, related issues, suggestions how to fix)**

The exception message: `Value cannot be null. Parameter name: dispatcher`

Here's the stack trace:
Inner Exception:
   ```
at System.Reactive.Concurrency.CoreDispatcherScheduler..ctor(CoreDispatcher dispatcher)
   at System.Reactive.Concurrency.CoreDispatcherScheduler.get_Current()
   at ReactiveUI.PlatformRegistrations.<>c.<Register>b__0_5()
   at ReactiveUI.WaitForDispatcherScheduler.attemptToCreateScheduler()
   at ReactiveUI.PlatformRegistrations.Register(Action`2 registerFunction)
   at ReactiveUI.DependencyResolverMixins.InitializeReactiveUI(IMutableDependencyResolver resolver)
   at ReactiveUI.RxApp.<>c.<.cctor>b__0_0()
   at Splat.Locator.RegisterResolverCallbackChanged(Action callback)
   at ReactiveUI.RxApp..cctor()

```
Outer Exception:
   ```
at ReactiveUI.RxApp.get_MainThreadScheduler()
[...]

Issue #1001, opened 2015-12-21
RxUI application as a target for "sharing content"


ghuntley:
ghuntley [10:31 PM] @flagbug: I'm doing UWP currently, as is @moswald so
    between us we got that covered.
    
    flagbug [10:34 PM] @ghuntley: Beware, there are some major issues with RxUI
    if your app is a target for a share operation: 

    https://msdn.microsoft.com/en-us/library/windows/apps/hh465251.aspx
    because of the way that RxApp.MainThreadScheduler and friends work.

    ghuntley [10:39 PM] @flagbug: can you give some examples of what you have
    observed? Thanks for the heads up, not using sharing in this app fortunately.
    
    flagbug [10:41 PM] @ghuntley: You have to set `RxApp.MainThreadScheduler`
    when opening the app as share target, since the share target UI gets its own UI
    thread and then, when you close the share UI, you have to set
    `RxApp.MainThreadScheduler` back to its previous value, or else your main
    application will crash.
    

See also https://github.com/reactiveui/ReactiveUI/issues/965

flagbug:
As a workaround, I'm doing something like this in the `OnShareTargetActivated` method:

```cs
// Capture the UI thread scheduler of the main application, so we can reset it, when we
// exit the share flyout
IScheduler realMainThreadScheduler = RxApp.MainThreadScheduler;
RxApp.MainThreadScheduler = new WaitForDispatcherScheduler(() => CoreDispatcherScheduler.Current);

// Pretty much the same goes for the DefaulltItemTemplate, since the template is
// generated on a different UI thread, we have to reset it here and change it back when
// closing the share target
// 
// See https://github.com/reactiveui/ReactiveUI/issues/965 for more information
var previousTemplate = AutoDataTemplateBindingHook.DefaultItemTemplate;
AutoDataTemplateBindingHook.DefaultItemTemplate = new Lazy<DataTemplate>(() =>
{
    const string template = "<DataTemplate xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation' xmlns:xaml='using:ReactiveUI'>" +
        "<xaml:ViewModelViewHost ViewModel=\"{Binding}\" VerticalContentAlignment=\"Stretch\" HorizontalContentAlignment=\"Stretch\" IsTabStop=\"False\" />" +
    "</DataTemplate>";
    var assemblyName = "";

[...]

PR #2100, opened 2019-06-29
fix: Replace default `MainThreadScheduler` on UAP

hinterlandsupplyco:
This replaces the usage of `WaitForDispatcherScheduler` and `CoreDispatcherScheduler` with a scheduler `SingleWindowDispatcherScheduler` that explicitly dispatches to the main app window regardless of how many windows are created by the application.

See #2092, #2034 for more info. Should also resolve or at least improve the situation in #1001. Every window created by the `CoreApplication` gets it's own shiny new UI thread and `CoreDispatcher`. Example of this would be applications which use casting, `ShareTarget`, `ProjectionManager` or are just snazzy enough to have an app with multiple windows.

Because `WaitForDispatcherScheduler` uses a factory to get an instance of `CoreDispatcherScheduler` and `CoreDispatcherScheduler` uses `Window.Current` to obtain a `CoreDispatcher` there are multiple ways for `MainThreadScheduler` to be hijacked by a second window. For instance, if the thread being used by that window is returned to the thread pool (e.g. - by that window being closed or launching your main window from a `ShareTargetActivated`) then forever that `[ThreadStatic]` instance of `MainThreadScheduler` will attempt to dispatch to an non-existent window. This will surface in thread marshaling exceptions which will very likely crash the user's application (or cause nasty UI issues like controls which no longer update).

Of course, in these situations most devs will simply inject a scheduler into their view models for anything that used on a separate window. But this still won't block the hijacking because if code from the main window is ever executed on the thread separate window and then attempts to get an instance of `MainThreadScheduler` you could still have this problem. An example of this is code executing from the UI thread of window 2 updates an observable which is subscribed to on a view model or service being used in window 1. If in that code now in window 1 attempts to `ObserveOn(RxApp.MainThreadScheduler)` and `MainThreadScheduler` does not have an instance for that thread (which it likely wouldn't since we aren't using it on our other window in this example) it will grab the wrong dispatcher because for that thread `Window.Current.Dispatcher` is not in fact null, but is the dispatcher for window 2.

Since that is a very likely scenario in Rx, the problem should be solved in ReactiveUI.

**What is the new behavior?**

`SingleWindowDispatcherScheduler` does not use `CoreDispatcherScheduler` as its underlying scheduler, but does use code and patterns from that implementation.

It also does not use `Window.Current` to find the correct dispatcher. Instead it uses `CoreApplication.Views` to find the dispatcher. This has several advantages: 1) It is always available and you can get the correct dispatcher from any thread, 2) It is not possible to get the wrong dispatcher because your app's main window will always exist at `CoreApplication.Views[0]`

**What might this PR break?**
If for some reason a user was expressly relying on `RxApp.MainThreadScheduler` being of a specific type they will be discouraged from doing that _ever_ again.

- [] Tests for the changes have been added (for bug fixes / features)
- [] Docs have been added / updated (for bug fixes / features)

**Other information**:
Unfortunately this code is not testable in a unit test, but then again neither are other platform schedulers (see `NSRunloopScheduler`), but we are running this code on our beta channel in an app that supports multiple windows and it is performing well without crashes. Since most of the actual dispatching code is taken from `CoreDispatcherScheduler` I think we are ok.

A separate PR will be added to the website documentation to make it **VERY** clear that ReactiveUI supports one window UWP environments only out of the box and suggest appropriate workarounds for multiple windows.

Also tested that this will not cause a regression on fixes from c5b9462 (original issue: #1221)


hinterlandsupplyco:
@glennawatson I did give a multiple window scheduler the good ole' college try but after about a week of testing I do not think it's possible. See an attempt here: https://gist.github.com/hinterlandsupplyco/c8e0396f6d837fdc9cd2a5b00688b451

glennawatson:
I'm going to get multiple reviewers on this one who have some uwp experience. So hang in there.

glennawatson:
Thanks for this one @hinterlandsupplyco sure it will assist others.

hinterlandsupplyco:
@glennawatson happy to contribute! :-)

jasonwurzel:
this fix does not work in the constellation WPF +  WindowsXamlHost + UWP content. 
CoreApplication.Views.Count is > 0, yet accessing CoreApplication.Views[0].Dispatcher throws an Exception. I'm looking into it, maybe I can implement a fallback for XamlIslands...

glennawatson:
Hey @jasomwurzel if you can make a new issue even linking back to this one that would be great.

KAS1990:
Hello. I have tested new SingleWindowDispatcherScheduler and found bug. One of my view models has such code
Observable.Interval(TimeSpan.FromSeconds(.5)).Select(_ => State)
    .DistinctUntilChanged()
    .SubscribeOn(RxApp.MainThreadScheduler)
    .Subscribe(_ => this.RaisePropertyChanged(nameof(State)))
    .DisposeWith(DisposableOnDestroy);
My app allows user to create new view by ProjectionManager.StartProjectingAsync. Sometimes when my app executes 
await CoreApplication.CreateNewView().Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () => {...});
COMException "Element not found" is fired, and execution is broken on this.RaisePropertyChanged(nameof(State) or outside my code (somewhere in System.Private.CoreLib.dll).
In first situation stack trace leads to 
SingleWindowDispatcherScheduler.ScheduleOnDispatcherNow<TState>(TState state, Func<IScheduler, TState, IDisposable> action) line 152, where scheduler tries to get CoreApplication.Views[0].Dispatcher, but it cannot do it. I do not know why?
In second situation i cannot get stack trace but i think the problem is the same.
Can you help me?

PR #2101, opened 2019-06-30
feature: add monoandroid80 TargetFramework

weitzhandler:
Enables ReactiveUI.Uno support for MonoAndroid80.

The reason for adding this is that otherwise, an obsolete Droid project of that platform will fallback to .NET Standard 2.0, requiring `Reactive.Wasm` package.


glennawatson:
Closing. ReactiveUI minimum is 8.1

jeromelaban:
There is no need for `monoandroid80`, and monoandroid90 will be a requirement by the end of the year anyways. See https://developer.android.com/distribute/best-practices/develop/target-sdk

glennawatson:
We tend to let xamarin forms dictate our mono Android version. So being one major revision behind we tend to bounce up when they do.

PR #2133, opened 2019-08-15
feature: add view model change observables to android ReactiveRecyclerViewViewHolder 

clintonrocksmith:
Thus eliminating the need to have a collection to get the ViewModel from.
This is helpful when using IObservable<IChangeSet<TViewModel>> as we don't have a list if it has been sorted during the Dynamic Data connection

Feature to return the ViewModel of the item that is selected in the adaptor for Android

**What might this PR break?**
None

**Other information**:

### DIFF
--- a/src/ReactiveUI.AndroidSupport/ReactiveRecyclerViewViewHolder.cs
+++ b/src/ReactiveUI.AndroidSupport/ReactiveRecyclerViewViewHolder.cs
@@ -63,7 +63,7 @@ namespace ReactiveUI.AndroidSupport
             SelectedWithViewModel = Observable.FromEvent<EventHandler, TViewModel>(
                                 eventHandler =>
                                 {
-                                    void Handler(object sender, View.LongClickEventArgs e) => eventHandler(ViewModel);
+                                    void Handler(object sender, EventArgs e) => eventHandler(ViewModel);
                                     return Handler;
                                 },
                                 h => view.Click += h,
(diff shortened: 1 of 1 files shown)


src/ReactiveUI.AndroidSupport/ReactiveRecyclerViewViewHolder.cs
```diff
@@ -51,6 +51,16 @@ protected ReactiveRecyclerViewViewHolder(View view)
                 h => view.LongClick += h,
                 h => view.LongClick -= h)
                     .Select(_ => AdapterPosition);
+
+            SelectedWithViewModel = Observable.FromEventPattern(
```
glennawatson: We are attempting to use FromEvent() these days for performance reasons. FromEventPattern is double the performance.

glennawatson: you can use similar format to this 

```cs
Observable.FromEvent<StartupEventHandler, Unit>(
                    eventHandler =>
                    {
                        void Handler(object sender, StartupEventArgs e) => eventHandler(Unit.Default);
                        return Handler;
                    },
                    x => app.Startup += x,
                    x => app.Startup -= x);
```

clintonrocksmith: Roger that, I'll have look at this

PR #2163, opened 2019-09-09
fix: remove exception in .netstandard 2.0 code

### LINKED ISSUE #2162: [BUG] Client-side Blazor loads plain netstandard PlatformRegistrations.cs and throws an exception
**Describe the bug**
Blazor client-side is technically a netstandard2.0 project.  Even with the `ReactiveUI.Blazor` nuget package, the `PlatformRegistrations` class from reactiveui/platforms/netstandard2.0 is still loading and throwing an `Exception`.   

**Steps To Reproduce**
Either load this repo (and make sure latest VS preview with aspnetcore-prev9 is installed)
https://github.com/limefrogyank/RxUIBlazorError

OR
Provide the steps to reproduce the behavior:
1. Create Blazor client-side (wasm) project.
2. Install ReactiveUI.Blazor nuget package.
3. In App.razor at the bottom, add the following lines:
```
@code{
    protected override Task OnAfterRenderAsync(bool firstRender)
    {
        System.Diagnostics.Debug.WriteLine("About to call RxApp instance...");
        var threadScheduler = ReactiveUI.RxApp.MainThreadScheduler;
        return base.OnAfterRenderAsync(firstRender);
    }
}
```
4. Run the project and open the debug window in the resulting browser to see error messages.

**Expected behavior**
No error messages!  ReactiveUI should initialize just fine.

**Screenshots**

Annotation 2019-09-09 084006

**Environment**
- OS:       Windows10/IISExpress/Chrome,Edgium
- Version   10.1.5

**Additional context**
Simply commenting out that `throw Exception` line from the netstandard2.0 `PlatformRegistrations` would make this work.  Other than that, there's not a good way to identify a Blazor wasm app from a regular netstandard2.0 project.  Perhaps check for blazor assembies before throwing that error??

glennawatson:
WASM projects will run under .net standard so the exception will likely cause issues in the future.

This removes the exception in the platform registrations.

This fixes #2162

### DIFF
--- a/src/ReactiveUI/Platforms/netstandard2.0/PlatformRegistrations.cs
+++ b/src/ReactiveUI/Platforms/netstandard2.0/PlatformRegistrations.cs
@@ -4,6 +4,7 @@
 // See the LICENSE file in the project root for full license information.
 
 using System;
+using System.Reactive.Concurrency;
 
 namespace ReactiveUI
 {
@@ -16,7 +17,13 @@ namespace ReactiveUI
         /// <inheritdoc/>
         public void Register(Action<Func<object>, Type> registerFunction)
         {
-            throw new Exception("You are referencing the Portable version of ReactiveUI in an App. Please change your reference to the specific version for your platform found here: https://reactiveui.net/docs/getting-started/installation");
+            if (registerFunction == null)
+            {
+                throw new ArgumentNullException(nameof(registerFunction));
+            }
+
+            RxApp.TaskpoolScheduler = TaskPoolScheduler.Default;
+            RxApp.MainThreadScheduler = DefaultScheduler.Instance;
         }
     }
 }


src/ReactiveUI/Platforms/netstandard2.0/PlatformRegistrations.cs
```diff
@@ -16,7 +17,13 @@ public class PlatformRegistrations : IWantsToRegisterStuff
         /// <inheritdoc/>
         public void Register(Action<Func<object>, Type> registerFunction)
         {
-            throw new Exception("You are referencing the Portable version of ReactiveUI in an App. Please change your reference to the specific version for your platform found here: https://reactiveui.net/docs/getting-started/installation");
```
worldbeater: Probably worth leaving a log message here for those newbies who are experiencing issues with threading and start looking into the debug output to find answers about what's going wrong with scheduling. Although not so critical and shouldn't stop the PR :+1:

PR #2168, opened 2019-09-16
Prioritize DependencyObject over INPC for UWP based projects

dr1rrb:
Bug fix

When building an Observable to observe a property of an object, if it inherits for `DependencyObject` and implements `INotifyPropertyChanged` the observable will be built using the `INotifyPropertyChanged.PropertyChanged` event.

The issue is that on iOS and Android the `UIElement` implements the `INotifyPropertyChanged` but it never raise the event: https://github.com/unoplatform/uno/issues/1551.

**What is the new behavior?**
On UNO projects, if an object inherit from `DependencyObject` we will create a observable from the `DependencyProperty` if found and fallback to the `INotifyPropertyChanged` otherwise.

**What might this PR break?**
If a `DependencyObject` has a `DependencyProperty` which is bypassed by the object's property getter / setter, the resulting observable won't produce any value.

**Other information**:
The real issue is on the Uno side (https://github.com/unoplatform/uno/issues/1551) however it's safer / lighter to change the priority in ReactiveUI for now, and rollback this commit once the issue has been fixed in Uno.

### DIFF
--- a/src/ReactiveUI/Platforms/uap/DependencyObjectObservableForProperty.cs
+++ b/src/ReactiveUI/Platforms/uap/DependencyObjectObservableForProperty.cs
@@ -37,7 +37,13 @@ namespace ReactiveUI
                 return 0;
             }
 
+#if HAS_UNO
+            // DependencyObject must be prioritized compared to INotifyPropertyChanged (score == 5)
+            // until https://github.com/unoplatform/uno/issues/1551 is resolved.
+            return 6; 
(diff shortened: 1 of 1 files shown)


glennawatson:
Some minor code errors being detected there. I'd be happy to make this change for all platforms tbh so feel free to remove the ifdef and just make dependency props win

dr1rrb:
Here it is :)

PR #2180, opened 2019-09-26
fix: the WPF designer by removing abstract

glennawatson:
Chris Pullman on Slack noticed that if classes aren't marked as abstract, the designer will function for WPF, where previously they wouldn't.

Removing the abstract on these classes to allow for the designer to function.


ChrisPulman:
Thanks for the mention @Glennawatson

PR #2186, opened 2019-10-01
Feature: Add warning when running .NET Standard

### LINKED ISSUE #2164: Add warning about .net standard project
Add warning message if using the .NET Standard platform.

It was previous happening before this version came in this file https://github.com/reactiveui/ReactiveUI/blob/08503a73818157705765a0fa6742718ec4de65dd/src/ReactiveUI/Platforms/netstandard2.0/PlatformRegistrations.cs#L20

We have to be careful so we don't confuse WASM users as well.

rgomez90:
Fixes #2164 

* Adds a warning message when running .NET Standard informing the user a host package needs to be installed.

None

**What is the new behavior?**

None

**What might this PR break?**
Nothing

**Other information**:

### DIFF
--- a/src/ReactiveUI/RxApp.cs
+++ b/src/ReactiveUI/RxApp.cs
@@ -149,7 +149,7 @@ namespace ReactiveUI
                     _hasSchedulerBeenChecked = true;
                     LogHost.Default.Warn("It seems you are running .NET Standard, but there is no host package installed!\n");
                     LogHost.Default.Warn("You will need to install the specific host package for your platform (ReactiveUI.WPF, ReactiveUI.Blazor, ...)\n");
-                    LogHost.Default.Warn("You can install the needed package via NuGet");
+                    LogHost.Default.Warn("You can install the needed package via NuGet, see https://reactiveui.net/docs/getting-started/installation/");
                 }
 
                 return _mainThreadScheduler;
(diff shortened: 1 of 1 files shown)


src/ReactiveUI/RxApp.cs
```diff
@@ -135,7 +136,25 @@ static RxApp()
...
+                    return _unitTestMainThreadScheduler;
+                }
+
+                // If Scheduler is DefaultScheduler, user is likely using .NET Standard
+                if (!_hasSchedulerBeenChecked && _mainThreadScheduler == Scheduler.Default)
+                {
+                    _hasSchedulerBeenChecked = true;
+                    LogHost.Default.Warn("It seems you are running .NET Standard, but there is no host package installed!\n");
+                    LogHost.Default.Warn("You will need to install the specific host package for your platform (ReactiveUI.WPF, ReactiveUI.Blazor, ...)\n");
+                    LogHost.Default.Warn("You can install the needed package via NuGet");
```
worldbeater: Looks good, but how about referencing the [installation instructions](https://reactiveui.net/docs/getting-started/installation/)? :book: 
```suggestion
                    LogHost.Default.Warn("You can install the needed package via NuGet, see https://reactiveui.net/docs/getting-started/installation/");
```

rgomez90: @worldbeater you're right, a reference should be added

Issue #2182, opened 2019-09-29
Different NuGet info for each package


TonyGravagno:
The description for all ReactiveUI packages in NuGet is the same. It's difficult to look at the packages and work out which ones to use for a project. I believe this makes the onboarding experience more difficult for the newcomer, myself included.

> A MVVM framework that integrates with the Reactive Extensions for .NET to create elegant, testable User Interfaces that run on any mobile or desktop platform. Supports Xamarin.iOS, Xamarin.Android, Xamarin.Mac, Xamarin Forms, Xamarin.TVOS, Tizen, WPF, Windows Forms, Universal Windows Platform (UWP) and the Uno Platform.

The [general information](https://reactiveui.net/docs/getting-started/installation/) for installation on the website is very helpful, as are individual pages like [this one](https://reactiveui.net/docs/getting-started/installation/xamarin-forms).

I'm requesting and recommending that the NuGet descriptions should have individual descriptions which are more in line with the project-specific info, including recommendations for the project type which they support and the other packages recommended for that project type.

Further, the tags for packages don't accurately describe them. Consider for ReactiveUI.WPF the tags include several related to xamarin and ios.

Finally, the project URL for all packages goe to reactiveui.net, and while accurate, that doesn't provide information about the package and its specific usage.

I'm guessing the content for the NuGet packages was created early-on and that this is all just a suggestion for taking the next step beyond that initial and successful effort.

Examples: Here are two identical descriptions for two different packages:
![image](https://user-images.githubusercontent.com/499473/65838696-ad40a480-e2ba-11e9-9270-9df23d0de78a.png)

![image](https://user-images.githubusercontent.com/499473/65838718-f5f85d80-e2ba-11e9-8fc1-e6fc7e6ac8f8.png)

glennawatson:
Hey @TonyGravagno while the core maintainers won't be able to address this one immediately it's a new contributor friendly task. Essentially in the .csproj for each projects you would customise the metadata. 

At the moment it picks it up from the Directory.build.props file if not overridden in the project files.

rgomez90:
@glennawatson I could give this a try, if still needed.

If so, should I pull the descriptions from somewhere or should I define them by myself?

rgomez90:
@glennawatson I think using the descriptions at [https://reactiveui.net/docs/getting-started/installation/#release-packages](https://reactiveui.net/docs/getting-started/installation/#release-packages) should be the way to go....

What do you think?

[...]

PR #2189, opened 2019-10-02
Housekeeping: Msbuild nuget package description validation

dpvreony:

Housekeeping for build process to enforce #2182 and ensure nuget package descriptions are set AND not the default.

#2182 all packages have same description

**What is the new behavior?**
msbuild task in targets file to validate pre-compile stage.

**What might this PR break?**

**Other information**:

merge after #2182 as it will break the build prior to that being completed.

### DIFF
--- a/src/Directory.build.props
+++ b/src/Directory.build.props
@@ -12,7 +12,8 @@
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://reactiveui.net</PackageProjectUrl>
     <PackageIconUrl>https://github.com/reactiveui/styleguide/blob/master/logo/main.png?raw=true</PackageIconUrl>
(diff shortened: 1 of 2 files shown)


src/Directory.build.props
```diff
@@ -4,6 +4,7 @@
     <NoWarn>$(NoWarn);1591;1701;1702;1705;VSX1000</NoWarn>
     <Platform>AnyCPU</Platform>
     <IsTestProject>$(MSBuildProjectName.Contains('Tests'))</IsTestProject>
+    <IsTestHelperProject>$(MSBuildProjectName.Contains('Testing'))</IsTestHelperProject>
```
glennawatson: Testing is actually released as a NuGet package. It contains testing utilities for client's.

glennawatson: https://www.nuget.org/packages/ReactiveUI.Testing/

dpvreony: doh ok i'll remove that

PR #2196, opened 2019-10-04
Fix: interactions can have reentry

### LINKED ISSUE #1396: Modal WPF dialog from Interaction handler breaks ReactiveUI viewmodel

anaisbetts:

Fixes #1396 by ensuring calls to ShowDialog aren't reentrant

**What might this PR break?**

Might make WPF/WinForms code ever slightly slower.

## TODO:

- [ ] Make sure this actually fixes the repro of this bug

### DIFF
--- a/src/ReactiveUI.Winforms/Registrations.cs
+++ b/src/ReactiveUI.Winforms/Registrations.cs
@@ -41,6 +41,8 @@ namespace ReactiveUI.Winforms
                 WindowsFormsSynchronizationContext.AutoInstall = true;
                 RxApp.MainThreadScheduler = new WaitForDispatcherScheduler(() => new SynchronizationContextScheduler(new WindowsFormsSynchronizationContext()));
             }
+
+            Interaction.DefaultInteractionScheduler = RxApp.MainThreadScheduler;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


anaisbetts:
This fix didn't actually fix the underlying issue, even though the code was correctly scheduled on the thread I expected :-/  We'll revisit this later

Issue #1939, opened 2019-01-31
Android: add unit tests


glennawatson:
**Is your feature request related to a problem? Please describe.**
At the moment we don't run the unit tests in the Android platform of the unit tests. We need to run them and run a emulator to run all the tests.

**Describe the solution you'd like**
That unit tests run through the android emulator

**Describe suggestions on how to achieve the feature**
Something along the lines of https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/android?view=azure-devops#test-on-the-android-emulator

Kavignon:
So you want to run the Android test in an emulator with the Asure DevOps pipeline to enhance the CI workflow?

glennawatson:
I would follow what's in the https://github.com/reactiveui/splat project -- Add a android project for the tests, add a target framework to ReactiveUI.Tests as netstandard2.0, and then derive all run the tests through the android test project.

dpvreony:
the runner in splat is https://github.com/xunit/devices.xunit automation would be tied to https://github.com/xunit/devices.xunit/issues/79

glennawatson:
This was added as part of #2201

PR #2201, opened 2019-10-09
feature: add Android unit tests

duracellko:
Implements #1939 Android: add unit tests

* Added **ReactiveUI.Android.Tests** project that runs unit tests **ReactiveUI.Tests** in Android.
* `LocalizableAttribute` was enabled for Android target framework, because it is used in some tests.
* `PocoObservableForPropertyTests`
  * `TestLogger` is unregistered after test run. In Android AppDomain is not unloaded after test run, so `TestLogger` keeps registered.
  * `NotificationPocoErrorOnBind` test is executed twice to verify that message is logged first time only.
* `DefaultViewLocatorTests.NoErrorIsRaisedIfATypeCannotBeFound` adds assembly name in "DoesNotExist" type name, because it was failing on Mono.

**Other information**

* **Resource.designer.cs** is excluded from Git repository. It is automatically generated by Visual Studio.
  * Maybe **Cake** build should create empty **Resource.designer.cs** file. MSBuild requires empty file to generate Android resources. It fails, when the file is missing.
* Tests **ApiApprovalTests** are failing on Android. I don't think **PublicApiGenerator** is supported on Android.

**Open questions**

1. Does **ReactiveUI.Android.Tests** project need to be added to **Cake** build? I don't see it in **ReactiveUI.Splat** repository.


glennawatson:
I have been planning on replacing the API approval so feel free to leave those out. 

Thanks for doing this.

Issue #2170, opened 2019-09-18
[BUG] PropertyBindingMixins.Bind crashes


setallunder:
**Describe the bug**
`ReactiveUI.PropertyBindingMixins.Bind` throws `android.runtime.JavaProxyThrowable: System.Reflection.TargetParameterCountException: Parameter count mismatch.`

**Steps To Reproduce**
1. Create new Xamarin.Android project;
2. Add layout with TextView;
3. Invoke `ReactiveUI.PropertyBindingMixins.Bind` inside activity to bind `TextView.Text` with ViewModel for example:
```
this.Bind(
                ViewModel,
                x => x.ImportantText,
                x => x.SuperTextView.Text);
```
4. Launch app on device;
5. Observe crash.

**Expected behavior**
No crash

**Stacktrace**
```
09-18 13:00:36.914 E/AndroidRuntime(12346): FATAL EXCEPTION: main
09-18 13:00:36.914 E/AndroidRuntime(12346): Process: ReaciveUITest.ReaciveUITest, PID: 12346
09-18 13:00:36.914 E/AndroidRuntime(12346): android.runtime.JavaProxyThrowable: System.Reflection.TargetParameterCountException: Parameter count mismatch.
09-18 13:00:36.914 E/AndroidRuntime(12346):   at System.Delegate.CreateDelegate (System.Type type, System.Object firstArgument, System.Reflection.MethodInfo method, System.Boolean throwOnBindFailure, System.Boolean allowClosed) [0x000ee] in <ff07eae8184a40a08e79049bbcb31a0e>:0 
09-18 13:00:36.914 E/AndroidRuntime(12346):   at System.Delegate.CreateDelegate (System.Type type, System.Object firstArgument, System.Reflection.MethodInfo method) [0x00000] in <ff07eae8184a40a08e79049bbcb31a0e>:0 
09-18 13:00:36.914 E/AndroidRuntime(12346):   at System.Reactive.ReflectionUtils.CreateDelegate[TDelegate] (System.Object o, System.Reflection.MethodInfo method) [0x00000] in <f9e209095f5a4628a0b7a00d8a83c3b9>:0 
09-18 13:00:36.914 E/AndroidRuntime(12346):   at System.Reactive.Linq.ObservableImpl.FromEvent`2[TDelegate,TEventArgs].GetHandler (System.Action`1[T] onNext) [0x00025] in <f9e209095f5a4628a0b7a00d8a83c3b9>:0 
[...]

PR #2221, opened 2019-10-18
fix: RefCount of ValueTuple breaks Xamarin.Android

RLittlesII:

Fixes Xamarin.Anrdroid `this.Bind` implementation.

Resolves: #2170 

**What is the new behavior?**

This removes the `RefCount` which isn't generally someone cares about during binding.

**What might this PR break?**

`this.Bind`

**Other information**:

### DIFF
--- a/src/ReactiveUI/Bindings/Property/PropertyBinderImplementation.cs
+++ b/src/ReactiveUI/Bindings/Property/PropertyBinderImplementation.cs
@@ -628,9 +628,9 @@ namespace ReactiveUI
                 return null;
             }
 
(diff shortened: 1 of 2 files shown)


ajhuntsman:
@glennawatson: Any update on this? :-)

RLittlesII:
@ajhuntsman Seems all the fixes we have tried (including rolling back to previous state) don't seem to affect the problem.  I was able to reproduce a similar issue on Xamarin.iOS using 9.12.1.  We hope to have a resolution soon.  This one has been an interesting issue.

glennawatson:
Closing this PR since I don't think it solves the issue. @RLittlesII was going to lodge an issue with the rx team because it seems to be happening if you compile against MonoAndroid 9.0

Issue #2154, opened 2019-09-03
[BUG] UWP - SingleWindowDispatcherScheduler causes issues in XamlIsland


jasonwurzel:
**Describe the bug**
My UWP project lives inside a XamlIsland, that is, I have a WPF application having a WindowsXamlHost that in turn hosts a UWP control. Since #2100 any access to RxApp.MainThreadScheduler results in an Exception. 
The root cause is that in the newly introduced `SingleWindowDispatcherScheduler` the dispatcher is fetched via `CoreApplication.Views[0].Dispatcher`, but this is not present inside the XamlIsland Context. `CoreApplication.Views.Count` is > 0, but subsequently accessing `.Dispatcher` results in an Exception.

**Steps To Reproduce**

see https://github.com/jasonwurzel/XamlIslandDispatcherProblem for a repo. Start either ControlPlaygroundPackage (WPF - XamlIsland - Uwp) or UwpApp (UWP App) (in Debug - x64). Starting UwpApp results in no exception, starting the WPF App results in an exception when CoreApplication.Views[0].Dispatcher is accessed.
- Min Windows version is 1903 
- dotnet core sdk 3.0.100-preview5-011568 is expected to be present

**Expected behavior**
Accessing `RxApp.MainThreadScheduler` should not result in an exception. If it is correct that the Views in the Views enumeration have no dispatcher in the XamlIsland case, we have to find a workaround in that special case.

**Additional context**
I'm not sure whether the views in the Views enumeration should have a Dispatcher present or not. I'll check with the Windows.Toolkit guys as well. 
For simplicity's sake I ommitted referencing the actual ReactiveUI package in my repro (XamlIslands with 3rd party nuget references is still uncharted territory, unfortunately and requires some - more - fiddling with the csproj files.)

glennawatson:
I'll have a look on the weekend anyway. I would like to get this solved for rxui 10 release

jasonwurzel:
That's great! Maybe I'll have an answer until then from windows-toolkit ( https://github.com/windows-toolkit/Microsoft.Toolkit.Win32/issues/174 )

jasonwurzel:
Alexandre was swift to answer my issue on windows-toolkit, see [here ](https://github.com/windows-toolkit/Microsoft.Toolkit.Win32/issues/174#issuecomment-527524578) for his answer.
So basically this behaviour is by design and we have to find another way.
[...]

PR #2255, opened 2019-11-21
fix: sometimes the core dispatcher can't be fetched in constructor

jasonwurzel:
this fix is related to #2154 , some more testing showed that the former bugfix also has to be in place in ScheduleOnDispatcherNow

fix related to #2154 

When accessing the RxApp.MainThreadScheduler too early (i.e. no MainWindow yet), the CoreDispatcher can't be fetched in SingleWindowDispatcherScheduler's ctor. Thus, ScheduleOnDispatcherNow has to contain the original fix as well (check if _dispatcher is still null and if so, try again fetching the Dispatcher.)

**What is the new behavior?**

**What might this PR break?**

**Other information**:

### DIFF
--- a/src/ReactiveUI/Platforms/uap/SingleWindowDispatcherScheduler.cs
+++ b/src/ReactiveUI/Platforms/uap/SingleWindowDispatcherScheduler.cs
@@ -43,17 +43,7 @@ namespace ReactiveUI
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


glennawatson:
Thanks @jasonwurzel I will put out another release tomorrow.

PR #2275, opened 2019-12-08
feature: DefaultViewLocator now public for users to override ViewModelToViewFunc

fdoyon:
This is a feature update that allows users to override `DefaultViewLocator.ViewModelToViewFunc`  if they are using custom naming conventions.

Currently the users have to copy/paste the source of `DefaultViewLocator` and then change their `ViewModelToViewFunc` implementation. This is dangerous because if the logic of `DefaultViewLocator` changes, the copy-pasted code won't be able to carry the feature.

**What is the new behavior?**
Allow users to get a reference to the locator thusly:
```csharp
var viewLocator = Locator.Current.GetService<IViewLocator>() as DefaultViewLocator;
viewLocator.ViewModelToViewFunc = x => x.Replace("ViewModel", "MyViewModel");
```
**What might this PR break?**
nothing

**Other information**:
Discussed with @anaisbetts on slack.


fdoyon:
For some reason an unrelated test is broken in the CI build: https://dev.azure.com/dotnet/ReactiveUI/_build/results?buildId=25723&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=a8035825-ecc2-53a4-dc8f-13b782b0dc20&l=418

RLittlesII:
This isn't an unrelated test.  These tests verify changes to the public API of the code.  You changing from internal to public changes the API surface so the test failed.  The api approval change should be uploaded to the artifacts of the build so you can see the difference between the old API surface and the new API surface.  You should also be able to this to reproduce this locally running `.\build.ps1 --t=Test`

glennawatson:
Install a diff program like `beyondcompare` and it should be easy to modify with the API changes.

glennawatson:
You can copy <api>.received.txt to <api>.approved.txt -- that might get around your current issues.

fdoyon:
Thanks - now the winforms tests are crashing - I don't have a windows machine to test on I am afraid.

glennawatson:
We'll do the rest for you. Thanks for getting it this far.

PR #2293, opened 2019-12-29
fix: warnings about package downgrade

weitzhandler:
Bug fix

When targeting `ReactiveUI.Uno`, I'm getting warnings such as this one:

image

**What is the new behavior?**
Let's make it equal to the main ReactiveUI project, that depends on `Splat.Drawing` version `9.*`, rather than a specific version.
This causes warnings.


glennawatson:
This wouldn't solve your issue. Eg at the time the NuGet gets built the version gets locked in.

glennawatson:
Also we are using Dependabot

PR #2294, opened 2019-12-30
feature: ReactiveUI.Events.Uno support

weitzhandler:
Preparation for `ReactiveUI.Events.Uno` support.

There is currently no support for Uno Platform on ReactiveUI.Events.

**What is the new behavior?**
- In UAP project types, simply ports to ReactiveUI.Events.
- All other projects - generate and map to events from Uno.UI.

**What might this PR break?**
No breaking changes. Just new experiment.

### DIFF
--- a/src/ReactiveUI.Events.Uno/ReactiveUI.Events.Uno.csproj
+++ b/src/ReactiveUI.Events.Uno/ReactiveUI.Events.Uno.csproj
@@ -1,0 +1,29 @@
+<Project Sdk="MSBuild.Sdk.Extras">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard20;MonoAndroid90;Xamarin.iOS10;Xamarin.Mac20</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);uap10.0.16299</TargetFrameworks>
+    <AssemblyName>ReactiveUI.Events.Uno</AssemblyName>
+    <RootNamespace>ReactiveUI.Events</RootNamespace>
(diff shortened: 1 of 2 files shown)


glennawatson:
We won't be supporting any new Events projects. We are pushing users towards Pharmacist which solves a lot of support overhead for us.

glennawatson:
See https://github.com/reactiveui/pharmacist#msbuild for instructions how to use it.

weitzhandler:
Oh wow you mean we should be using Pharmacist in our own projects to have all events as Observables? Well that's awesome, haven't thought it should be that powerful...
So it's going to work for all Uno.UI dlls too, is that right?

glennawatson:
If you add Pharmacist MSBUILD and Uno NuGet's in your project it'll generate all the events for Uno.

Basically the project you add the Pharmacist MsBuild NuGet and any other NuGet it'll scan those NuGet's for available DLLs.

glennawatson:
Also if you look at this PR it was using Pharmacist anyway. The difference is if users use it directly they'll get a 100% match for their Uno version.

weitzhandler:
Thanks for all Glenn and Merry xmas!

glennawatson:
Btw if you want to see the generated file go into /obj/ directory and see Pharmacist.Nuget.g.cs file -- it's what is generated by Pharmacist for NuGet packages.

weitzhandler:
Does it work in the old csproj too?
I'm getting `The "PharmacistNuGetTask" task was not given a value for the required parameter "TargetFramework".` error.

glennawatson:
That flag is unlikely to do what you want.  What is you target framework set to.

weitzhandler:
It's an Uno platform project, so it's 4 projects: UWP, Droid & iOS (via Xamarin), and WASM (.NET Standard).

PR #2295, opened 2019-12-30
feature: Enable WASM on Blazor

richbryant:

Bug fix.  Changes to Registrations class to enable ReactiveUI.Blazor to determine operating platform effectively, Scheduler initialization

mono-wasm throws errors when initializing ReactiveCommand

**What is the new behavior?**

No errors, commands work.

**What might this PR break?**

Potential to break ReactiveUI.Blazor on Server *IF* scheduling changes.  Potential to become obsolete once mono-wasm utilizes wasm threading.

**Other information**:

Requires latest PR raised against System.Reactive.Wasm

### DIFF
--- a/src/ReactiveUI.Blazor/ReactiveUI.Blazor.csproj
+++ b/src/ReactiveUI.Blazor/ReactiveUI.Blazor.csproj
@@ -1,5 +1,4 @@
 ﻿<Project Sdk="MSBuild.Sdk.Extras">
-
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <PackageDescription>Contains the ReactiveUI platform specific extensions for Blazor</PackageDescription>
@@ -8,8 +7,10 @@
 
   <ItemGroup>
     <PackageReference Include="Reactive.Wasm" Version="1.*" />
-    <PackageReference Include="Microsoft.AspNetCore.Components" Version="3.1.0" />
-    <PackageReference Include="ReactiveUI" Version="11.0.6" />
+    <PackageReference Include="Microsoft.AspNetCore.Components" Version="3.1.0" />    
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\ReactiveUI\ReactiveUI.csproj" />    
+  </ItemGroup>
 </Project>
--- a/src/ReactiveUI.Blazor/ReactiveUI.Blazor.csproj
+++ b/src/ReactiveUI.Blazor/ReactiveUI.Blazor.csproj
@@ -7,10 +7,10 @@
 
   <ItemGroup>
     <PackageReference Include="Reactive.Wasm" Version="1.*" />
-    <PackageReference Include="Microsoft.AspNetCore.Components" Version="3.1.0" />    
+    <PackageReference Include="Microsoft.AspNetCore.Components" Version="3.1.0" />
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\ReactiveUI\ReactiveUI.csproj" />    
+    <ProjectReference Include="..\ReactiveUI\ReactiveUI.csproj" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


src/ReactiveUI.Blazor/ReactiveUI.Blazor.csproj
```diff
@@ -7,11 +7,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Reactive.Wasm" Version="1.0.1-preview.12" />
     <PackageReference Include="Microsoft.AspNetCore.Components" Version="3.1.0" />
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\..\Users\richardb\source\repos\richbryant\Reactive.Wasm\src\System.Reactive.Wasm\System.Reactive.Wasm.csproj" />
```
glennawatson: This is likely to be an issue. I would recommend just doing a NuGet reference.

richbryant: on it

glennawatson: You can use NuGet wildcard versioning. Eg 1.* and then we can set the version of system.reactive.wasm to match.

richbryant: Nuts.  Okay. Will commit again.

richbryant: Build failed with wildcard versioning so I had to revert that.  Please feel free to amend this if you can make it build :(

src/ReactiveUI.Blazor/Registrations.cs
```diff
@@ -27,18 +25,13 @@ public void Register(Action<Func<object>, Type> registerFunction)
...
-            else
-#endif
-            {
-                RxApp.TaskpoolScheduler = TaskPoolScheduler.Default;
-                RxApp.MainThreadScheduler = new WaitForDispatcherScheduler(() => CurrentThreadScheduler.Instance);
+                PlatformEnlightenmentProvider.Current.EnableWasm();
             }
+
+            RxApp.TaskpoolScheduler = TaskPoolScheduler.Default;
+            RxApp.MainThreadScheduler = new WaitForDispatcherScheduler(() => CurrentThreadScheduler.Instance);
```
anaisbetts: We should never have to wait for the dispatcher to get CurrentThread, there is no dispatcher!

richbryant: I tried to change as little as possible but I can remove that line and test again if you like.

src/ReactiveUI.Blazor/ReactiveUI.Blazor.csproj
glennawatson: You should be able to wildcard now that I released 1.2.1

richbryant: it's 01:13Hrs here.  I'll do that tomorrow :) Probably mistake-prone right now.

src/ReactiveUI.Blazor/ReactiveUI.Blazor.csproj
```diff
@@ -9,10 +9,7 @@
   <ItemGroup>
     <PackageReference Include="Reactive.Wasm" Version="1.*" />
     <PackageReference Include="Microsoft.AspNetCore.Components" Version="3.1.0" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <ProjectReference Include="..\ReactiveUI\ReactiveUI.csproj" />
```
glennawatson: It should be a ProjectReference for the ReactiveUI.Csproj file.

src/ReactiveUI.Blazor/Registrations.cs
glennawatson: The name of `WaitForDispatcherScheduler` is deceptive. It just checks for an exception, if there is one just returns the `CurrentThreadScheduler` instead.

glennawatson: So it has no reliance on the dispatcher directly. It was intended to be used with the DispatcherScheduler though.

PR #2303, opened 2020-01-06
fix: Blazor StateHasChanged not firing on ViewModel change

### LINKED ISSUE #2298: [BUG] ReactiveComponentBase doesn't update properties when ViewModel is set/changed
**Describe the bug**
If you first create your view implementing `ReactiveComponentBase<SomeViewModel>` and then later set the `ViewModel` property to an instance of `SomeViewModel`, the page does not update with the contents of the viewmodel.  The page will only update when you manually call `StateHasChanged()`.

Related to the end of #2240 

**Steps To Reproduce**
1.  Create a `ReactiveComponentBase<>` view with the ViewModel property set to null.
2.  Set some properties in the view with the null-conditional operator.  `Text=@(ViewModel?.SampleText)`
3.  Create a button that creates a ViewModel and sets it to the view's ViewModel property.
4.  Run, then hit the button:  None of the properties will show on the screen.

**Expected behavior**
As soon as the ViewModel is set, the page should update to show the new contents.

**Additional context**
https://github.com/reactiveui/ReactiveUI/blob/master/src/ReactiveUI.Blazor/ReactiveComponentBase.cs#L40-L52

The problem is here:
```
 var propertyChangedObservable = this.WhenAnyValue(x => x.ViewModel)
                .Select(x => Observable.FromEvent<PropertyChangedEventHandler, Unit>(
                    eventHandler =>
                    {
                        void Handler(object sender, PropertyChangedEventArgs e) => eventHandler(Unit.Default);

                        return Handler;
                    },
                    eh => x.PropertyChanged += eh,
                    eh => x.PropertyChanged -= eh))
                .Switch();

            propertyChangedObservable.Do(_ => StateHasChanged()).Subscribe();
```
and here:
```
public T ViewModel
        {
            get => _viewModel;
            set
            {
                if (EqualityComparer<T>.Default.Equals(_viewModel, value))
                {
                    return;
                }

                _viewModel = value;
                OnPropertyChanged();
            }
        }
```
When the `ViewModel` setter is run, the `OnPropertyChanged` is triggering both `WhenAnyValue` observable and the inner event-observable.  Somehow, the Switch() statement is being triggered first... and the inner event-observable is being triggered on the old observable for ViewModel, rather than the new observable for the new ViewModel. 
http://reactivex.io/documentation/operators/switch.html
Apparently, those old events just go nowhere... so they are not triggering the Do() statement.

A fix would be to stick a StateHasChanged() into the ViewModel's setter after the `OnPropertyChanged` call.

glennawatson:
Fixes #2298 where the view model property being changed doesn't trigger a state refresh.

### DIFF
--- a/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
+++ b/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
@@ -37,7 +37,8 @@ namespace ReactiveUI.Blazor
         /// </summary>
         public ReactiveComponentBase()
         {
-            var propertyChangedObservable = this.WhenAnyValue(x => x.ViewModel)
+            this.WhenAnyValue(x => x.ViewModel).Subscribe(_ => StateHasChanged());
+            var viewModelsPropertyChanged = this.WhenAnyValue(x => x.ViewModel)
                 .Select(x => Observable.FromEvent<PropertyChangedEventHandler, Unit>(
                     eventHandler =>
                     {
@@ -49,7 +50,7 @@ namespace ReactiveUI.Blazor
                     eh => x.PropertyChanged -= eh))
                 .Switch();
 
-            propertyChangedObservable.Do(_ => StateHasChanged()).Subscribe();
+            viewModelsPropertyChanged.Do(_ => StateHasChanged()).Subscribe();
         }
 
         /// <inheritdoc />
--- a/src/ReactiveUI.Blazor/ReactiveLayoutComponentBase.cs
+++ b/src/ReactiveUI.Blazor/ReactiveLayoutComponentBase.cs
@@ -33,7 +33,8 @@ namespace ReactiveUI.Blazor
         /// </summary>
         public ReactiveLayoutComponentBase()
         {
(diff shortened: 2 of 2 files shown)


src/ReactiveUI.Blazor/ReactiveComponentBase.cs
```diff
@@ -37,7 +37,8 @@ public class ReactiveComponentBase<T> : ComponentBase, IViewFor<T>, INotifyPrope
         /// </summary>
         public ReactiveComponentBase()
         {
-            var propertyChangedObservable = this.WhenAnyValue(x => x.ViewModel)
+            this.WhenAnyValue(x => x.ViewModel).Subscribe(_ => StateHasChanged());
+            var viewModelsPropertyChanged = this.WhenAnyValue(x => x.ViewModel)
```
michaelstonis: Doesn't this still need a check for null? When I was testing this out, part of the issue is that the event chaining failed when the ViewModel was null.

Adding a `.Where(x => x != null)` or a default check should resolve that.

limefrogyank: or a `.WhereNotNull()`

PR #2323, opened 2020-01-20
Fix and refactor Visual Studio snippets

Lixfeld:
Fix interaction snippet: `=>` replaced with `{ get; } =`
Some minor final syntax adjustments

Failed to find a registration for a Interaction.

**What is the new behavior?**
No exception. The property returns the registered Interaction instead of a new one.

### DIFF
--- a/snippets/Visual Studio/ruiiv4.snippet
+++ b/snippets/Visual Studio/ruiiv4.snippet
@@ -20,7 +20,7 @@
       </Declarations>
       <Code Language="csharp">
         <![CDATA[public static readonly DependencyProperty ViewModelProperty =
-            DependencyProperty.Register(nameof(ViewModel), typeof($classname$Model), typeof($classname$), new PropertyMetadata(null));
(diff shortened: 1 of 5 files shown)


snippets/Visual Studio/ruiiv4.snippet
```diff
@@ -20,18 +20,18 @@
       </Declarations>
       <Code Language="csharp">
         <![CDATA[public static readonly DependencyProperty ViewModelProperty =
-            DependencyProperty.Register("ViewModel", typeof($classname$Model), typeof($classname$), new PropertyMetadata(null));
+            DependencyProperty.Register(nameof(ViewModel), typeof($classname$Model), typeof($classname$), new PropertyMetadata(null));
```
glennawatson: I seen people do `PropertyMetadata(typeof($classname$Model)))` instead of `PropertyMetadata(null))`

This way it handles value types.

Lixfeld: Thanks for the info.

Issue #2317, opened 2020-01-17
[BUG] Router.NavigateBack is still broken in some cases


garyng:
**Describe the bug**

This is related to https://github.com/reactiveui/ReactiveUI/issues/2286 and https://github.com/reactiveui/ReactiveUI/pull/2315, with some more test cases

**Steps To Reproduce**

Sample: [Test.zip](https://github.com/reactiveui/ReactiveUI/files/4077642/Test.zip)

The sample is based on the one provided by https://github.com/reactiveui/ReactiveUI/issues/2286

There are three cases (all can be found inside `ItemViewModel`):
1. `NavigateBack` inside `Subscribe`
    ```cs
    var completedInternally = Observable.Amb(
        Confirm.Select(_ => _),
        Cancel.Select(_ => _),
        Delete.Select(_ => _));

    completedInternally.Subscribe(_ =>
    {
        if (HostScreen.Router.NavigationStack.LastOrDefault() == this)
        {
            HostScreen.Router.NavigateBack.Execute().Subscribe(); 
            if (NextViewModel != null)
                HostScreen.Router.Navigate.Execute(NextViewModel).Subscribe();
        }
    });
    ```
1. `NavigateBack` inside `SelectMany`
    ```cs
    var completedInternally = Observable.Amb(
            Confirm.Select(_ => _),
            Cancel.Select(_ => _),
            Delete.Select(_ => _))
        .Where(_ => HostScreen.Router.NavigationStack.LastOrDefault() == this)
        .SelectMany(_ => HostScreen.Router.NavigateBack.Execute());
    completedInternally.Subscribe(_ =>
    {
        if (NextViewModel != null)
            HostScreen.Router.Navigate.Execute(NextViewModel).Subscribe();
    });
    ```
1. `NavigateBack` inside `Do`
    ```cs
    var completedInternally = Observable.Amb(
            Confirm.Select(_ => _),
            Cancel.Select(_ => _),
            Delete.Select(_ => _))
        .Where(_ => HostScreen.Router.NavigationStack.LastOrDefault() == this)
        .Do(_ => HostScreen.Router.NavigateBack.Execute().Subscribe());
    completedInternally.Subscribe(_ =>
    {
        if (NextViewModel != null)
            HostScreen.Router.Navigate.Execute(NextViewModel).Subscribe();
    });
    ```

| Case | RxUI 11.1.11 | RxUI 11.1.6 |
| ---- | ------------ | ----------- |
[...]

PR #2331, opened 2020-01-26
fix: NavigateBack wrong values on the NavigationStack

RLittlesII:

@garyng provided a solution for fixing the `RoutedViewHost` Xamarin.Forms implementation.

fixes: #2317 

**What is the new behavior?**

The stack values do not get distorted after the navigation pattern provided.

**What might this PR break?**

Xamarin.Forms navigation

**Other information**:
I tested this in a sample and saw that the `NavigationStack` seems to be correct.  I am going to look into creating some unit tests for this behavior so we can make sure it continues to function as expected.

### DIFF
--- a/src/ReactiveUI.XamForms/RoutedViewHost.cs
+++ b/src/ReactiveUI.XamForms/RoutedViewHost.cs
@@ -135,6 +135,7 @@ namespace ReactiveUI.XamForms
                         popToRootPending = false;
                         return page;
                     })
+                    .Concat()
                     .Subscribe()
                     .DisposeWith(disposable);
 
--- a/src/ReactiveUI.XamForms/RoutedViewHost.cs
+++ b/src/ReactiveUI.XamForms/RoutedViewHost.cs
@@ -73,7 +73,7 @@ namespace ReactiveUI.XamForms
                     })
                     .Where(_ => !userInstigated)
                     .Where(x => x.Delta > 0)
-                    .SelectMany(
+                    .Select(
                         async x =>
                         {
                             // XF doesn't provide a means of navigating back more than one screen at a time apart from navigating right back to the root page
@@ -107,6 +107,7 @@ namespace ReactiveUI.XamForms
 
                             return Unit.Default;
                         })
+                    .Concat()
                     .Subscribe()
                     .DisposeWith(disposable);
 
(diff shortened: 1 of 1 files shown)


src/ReactiveUI.XamForms/RoutedViewHost.cs
```diff
@@ -94,14 +99,19 @@ public RoutedViewHost()
...
+                                if (CurrentPage is IViewFor page && x.CurrentViewModel != null)
+                                {
+                                    page.ViewModel = x.CurrentViewModel;
+                                }
                             }
 
                             return Unit.Default;
                         })
-                    .Subscribe());
+                    .Subscribe()
```
garyng: I think a `.Concat()` is needed here (https://gist.github.com/garyng/4f0a461eb7524b37b8a440d8bbff5a2a#file-routedviewex-cs-L142)? So that a new `NavigateBack` request will wait for the previous one to complete instead of running concurrently with each other (which is the main reason why the `NavigationStack` will go out of sync).

RLittlesII: Resolved.

garyng:
@RLittlesII Is your original fix of adding `.ObserveOn(Router.Scheduler)` still required? After thinking about it, the navigation changed notification might fire on any thread, and `Pop*Async` needs to be on the UI thread right?

I am not sure what is the default value for `Router.Scheduler` though...

RLittlesII:
@garyng I am not sure if its required.  I honestly thought my original fix resolved it.  Default for `Router.Scheduler` is `RxApp.MainThreadScheduler` but any consumer can override that I believe.

That was my thought that the notifications were not returning to the UI thread thus causing the issue.

garyng:
Ohh okay! I asked because I bumped into the issue where all the page-poppings are executed on other non-ui thread - it weirdly swallowed all the exceptions.

RLittlesII:
Well.  The code changed in the PR works with the previous scenario we were testing. So you're thinking the Popped notifications should happen on the MainThreadScheduler?

garyng:
I mean this chunk of code:
https://github.com/reactiveui/ReactiveUI/blob/877ddce4530d23f55992f36c7cb07b382160cdba/src/ReactiveUI.XamForms/RoutedViewHost.cs#L85-L97
Should be executed on the main ui thread right? Since all the `Pop*Async`s are all ui operations.

RLittlesII:
Yes.  Those are the Xamarin.Forms navigation methods.

PR #2353, opened 2020-02-25
feature: Add ReactiveUI.Drawing

### LINKED ISSUE #2342: Remove Dependency on Splat.Drawing
When the components of Splat.Drawing were created, they solved a very definite problem in the .NET ecosystem, especially on the Xamarin side of things. Now, we have a pretty good set of translation libraries on the Xamarin side (Xamarin.Essentials, etc.) and with other updates to .NET that open up the ability to share drawing primitives easily across platform.

Initializing Splat.Draw comes at a cost for something that I would imagine _most users don't use and perhaps don't even really know exists_. This appears to be especially true for Xamarin.Android which loads through all images in the bundle.

Split it out so that Splat.Drawing is some kind of opt-in. Either, make it a separate package that only loads when included or make it a flag that can be enabled, if needed.

Perhaps keep things the way that they are more or less today, but make it so that it can be disabled. Right now everything there is statically registered, so it is always called and doesn't give a way to opt-out, so that portion would need to be updated.

glennawatson:

feature

**What's changed?**

This adds ReactiveUI.Drawing.

This resolves #2342 

Splat.Drawing isn't required by all users and can cause a lot of startup processing for Android users.

Now only users who want to opt into the functionality provided by Splat.Drawing need to add the new assembly.

This also updates the tests to use netcoreapp3.1

**What might this PR break?**

Users will have to use ReactiveUI.Drawing if they are using the Splat.Drawing functionality.


src/ReactiveUI.Android.Tests/ReactiveUI.Android.Tests.csproj
```diff
@@ -71,7 +71,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="xunit.runner.devices">
-      <Version>2.*</Version>
+      <Version>2.5.25</Version>
```
RLittlesII: Specific version?

src/ReactiveUI.Tests/ReactiveUI.Tests.csproj
```diff
@@ -22,7 +23,7 @@
   </ItemGroup>
 
   <Choose>
-    <When Condition=" $(TargetFramework.StartsWith('netcoreapp3.0')) ">
+    <When Condition=" $(TargetFramework.StartsWith('netcoreapp3.0')) AND '$(OS)' == 'Windows_NT'">
```
RLittlesII: Should this check for `netcoreapp3.x`?

PR #2356, opened 2020-02-27
Improve ability of RegisterViewsForViewModels to find the correct interface

FreddyD-GH:

Bug fix

If you create an interface ```IMoreSpecificViewFor<T> : IViewFor<T>``` and have a view that implements that interface instead of ```IViewFor<T>```, calling ```RegisterViewsForViewModels``` registers the view using the wrong interface, and ReactiveUI fails to find the proper view at runtime.

**What is the new behavior?**

```RegisterViewsForViewModels``` will register the proper interface and allow views to be created properly.

**What might this PR break?**
Realistically, should not break anyone.  However, I suppose it could break somebody who has a class that directly implements ```IViewFor<T>```, as well as one or more other classes that implement an extended version of the interface, and is specifically relying on only the first class being registered.

**Other information**:
I did not add any additional tests for the scenario.  The existing DependencyResolverTests should still pass.  No additional documentation should be necessary.

### DIFF
--- a/src/ReactiveUI/Mixins/DependencyResolverMixins.cs
+++ b/src/ReactiveUI/Mixins/DependencyResolverMixins.cs
@@ -75,8 +75,8 @@ namespace ReactiveUI
             foreach (var ti in assembly.DefinedTypes
                 .Where(ti => ti.ImplementedInterfaces.Contains(typeof(IViewFor)) && !ti.IsAbstract))
             {
-                // grab the first _implemented_ interface that also implements IViewFor, this should be the expected IViewFor<>
-                var ivf = ti.ImplementedInterfaces.FirstOrDefault(t => t.GetTypeInfo().ImplementedInterfaces.Contains(typeof(IViewFor)));
+                // grab the first _implemented_ interface that matches IViewFor<T>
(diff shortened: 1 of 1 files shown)


glennawatson:
I don't think this is a pr we would want in its current form sorry since a lot of users implement their own IViewFor. Happy to discuss this more in an issue if you want to find alternatives.

FreddyD-GH:
I can open an issue if you'd prefer, but I'm not sure I fully understand where the problem lies.  If you are referring to users implementing the non-generic ```IViewFor```, I did not remove the null check on the next line, which filters that out; it's the same as it was before.  If you are referring to users implementing the generic ```IViewFor<T>``` in their own class or struct, that shouldn't be affected either.  This would only affect _interfaces_ that extend ```IViewFor<T>```.  I'm sure you know the code base better than I do (especially since 3 people seem to agree), but I can't figure out how those would get correctly registered now.

FreddyD-GH:
Ah, I think I may understand.  Are you saying that users might depend on the types being registered incorrectly and are currently already implementing their own IViewLocator as a workaround?

Issue #2377, opened 2020-03-31
[BUG] Memory leak using ReactiveUI with Blazor Web Assembly


RyanMelenaNoesis:
**Describe the bug**
While setting up a simple test app to explore the usage of ReactiveUI with Blazor (client-side web assembly) I ran into what appears to be a memory leak causing excessive garbage collection followed by an application crash with an `OutOfMemoryException`.  Any insight would be appreciated.

**Steps To Reproduce**
1. Pull down example project [here](https://github.com/RyanMelenaNoesis/BlazorReactiveUI)
2. Run project
3. Open browser debug console
4. Notice that after some time the console will begin showing GC messages and the counter will slow down and eventually the page will crash

**Expected behavior**
Counter will increment without memory leak

**Screenshots**
![image](https://user-images.githubusercontent.com/7570707/78060333-f8fb0600-7350-11ea-8963-addb7b8f57a0.png)

**Environment**
- OS: Windows
- Version: 10
- Device: Dell XPS-15

**Additional context**
[console.log](https://github.com/reactiveui/ReactiveUI/files/4410823/console.log)

glennawatson:
I will see if i can early next week get some more information on this issue, so we can track it down to either rxui or rx libraries.

ronaldvdv:
@glennawatson I faced the same issue. It can be fixed by

- moving the second `WhenAnyValue` case in `OnAfterRender` into the `if`-close
- caching the results of both `WhenAnyValue`-statements into an `CompositeDisposable`
- disposing this in `Dispose(bool)` method

Also, I believe the statement `InvokeAsync(StateHasChanged)` leads to converting `StateHasChanged` into a new `Action` instance for every call, which can be avoided by creating a single `Action` instance in the constructor for this purpose.

RyanMelenaNoesis:
@ronaldvdv Thank you for digging into this, do you happen to have a fork with your changes that I could use for testing?

RLittlesII:
@ronaldvdv Thanks for that assessment.  I have noticed some issues with the way this control operates, although I have not encountered this memory leak.  I had some code I was working on to clean up some of the state changes logic.  I will test those changes against this sample to verify it resolves this issue.

ronaldvdv:
@RyanMelenaNoesis Nope, I happened to find this close before a deadline so I just added a copy of the {{ReactiveComponentBase}} class with above changes applied to my project and used it as a base for the components that ran out of memory.

[...]

PR #2391, opened 2020-04-25
fix: ReactiveComponentBase creating unnecessary subscriptions

RLittlesII:

Relates To: #2377 

The current implementation of the `ReactiveComponentBase` is potentially creating unnecessary subscriptions.  Currently, we are wiring up `this.WhenAnyValue` in the `OnAfterRender` outside the `firstRender` which causes additional subscriptions every render.

We currently have our logic checking for the first render, doing some subscriptions for state changes, and then doing others outside of that if block.  That code is potentially leaking subscriptions.  We also do not call that state change correctly in the observable pipeline.

**What is the new behavior?**

This change moves all subscription logic to inside the `firstRender` if statement.  This will reduce the number of subscriptions generated and should reduce leaks.

**What might this PR break?**

ReactiveUI.Blazor

**Other information**:

### DIFF
--- a/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
+++ b/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
@@ -85,6 +85,7 @@ namespace ReactiveUI.Blazor
         {
             if (firstRender)
             {
+                // The following subscriptions are here because if they are done in OnInitialized, they conflict with certain JavaScript frameworks.
                 var viewModelChanged =
                     this.WhenAnyValue(x => x.ViewModel)
                         .Publish()
--- a/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
+++ b/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Reactive;
+using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


glennawatson:
We need to have the `WhenAnyValue` to be in `OnAfterRender()` due to javascript engines not being able to handle it earlier. Several users had bugs. Move both the WhenAnyValue check inside the if statement. Also we can probably use  `RefCount()` on the shared observable.

Please update the description about `OnAfterRender()` above.

src/ReactiveUI.Blazor/ReactiveComponentBase.cs
```diff
@@ -76,32 +76,44 @@ public void Dispose()
...
+        {
+            if (firstRender)
+            {
+                var viewModelChanged =
+                    this.WhenAnyValue(x => x.ViewModel)
+                        .Publish()
+                        .RefCount();
+
+                viewModelChanged
+                    .Skip(1)
```
glennawatson, line 94: We should probably remove the `Skip(1)`, this makes sense in constructors where the value couldn't be any value but `null` but in this case it can be.

b-straub:
I applied the changes for testing purpose to my project. Unfortunately the fix will break any rerendering of the component from a PropertyChangedEvent. You can verify the problem by using the ReactiveUI Blazor serverside sample and applying the following changes:

FetchDataViewModel.cs:
```
private readonly ObservableAsPropertyHelper<bool> _loading;
  public bool Loading => _loading.Value;
….
_loading = LoadForecasts.IsExecuting.ToProperty(this, x => x.Loading, scheduler: RxApp.MainThreadScheduler);

….
private async Task<WeatherForecast[]> LoadWeatherForecastsAsync()
{
   await Task.Delay(TimeSpan.FromSeconds(2));
   return await _weatherForecastService.GetForecastAsync(DateTime.Now);
}
```
FetchDataView.razor:
```
@if (ViewModel.Loading)
{
    <p><em>Loading...</em></p>
}
else if (ViewModel.Forecasts.Any())
{
```

After loading has finished the page will not update. This can be fixed by changing 
`.RefCount(2);`

I'm not such a reactive expert that I can judge if this makes sense.

RLittlesII:
@b-straub Thanks for the input.  I am continuing conversation on #2377 as it relates to this change.

b-straub:
As far I have tested this PR must be fixed, otherwise view updates neither for SPA nor for WASM will work anymore. Please see my comment here: #2377

RLittlesII:
We are aware that this is breaking.  No packages have been released.  This will be resolved this week in a subsequent Pull Request.  No need to comment further on this pull request as it is already merged.

Issue #2395, opened 2020-04-27
Allow configuring or detecting current UI framework to avoid exceptions during dependency registrations


JesperTreetop:
**Is your feature request related to a problem? Please describe.**

When I start up my ReactiveUI WPF project in the debugger and have exceptions set to always break on first throw (which is very helpful sometimes), I always see a number of FileNotFoundExceptions thrown in assembly loading, because of  [DependencyResolverMixins.InitializeReactiveUI](https://github.com/reactiveui/ReactiveUI/blob/b7df987e1808b04ea7cc2da0b8d8ecc1477cc6f2/src/ReactiveUI/Mixins/DependencyResolverMixins.cs#L31-L39) making sure that all dependencies for all UI frameworks are registered. ([This line](https://github.com/reactiveui/ReactiveUI/blob/b7df987e1808b04ea7cc2da0b8d8ecc1477cc6f2/src/ReactiveUI/Mixins/DependencyResolverMixins.cs#L119) is where the throwing originates.)

**Describe the solution you'd like**

For registrations to not happen which would always knowably result in exceptions being thrown. At most half of them could be loaded at the same time (WPF, WinForms and Drawing) - the registration path has to always go down 3-5 blind alleys every launch for every ReactiveUI application. Not having all UI frameworks ReactiveUI supports loaded in every application is the opposite of *exceptional* (and it's the result of [vexing](https://ericlippert.com/2008/09/10/vexing-exceptions/) exceptions being handled down the stack), so if possible the exceptions should be avoided.

**Describe alternatives you've considered**

I'm seeing this because I have exceptions break on first throw and I could turn it off. I have a habit of having it enabled since WPF is spectacular about swallowing exceptions, and I'd rather know. This issue is mainly about developer ergonomics for me. There are performance implications to both throwing the exceptions and whatever other work happens on the road to those exceptions being thrown, but that's not my primary issue.

**Describe suggestions on how to achieve the feature**

For ReactiveUI to detect which assemblies are loaded by that time (AppDomain.CurrentDomain.GetAssemblies()) and opt into the namespaces to register based on that. Maybe there's a timing problem where the assemblies may not yet have loaded. In that case, being able to somehow provide a list of which UI frameworks would be good.

**Additional context**
Stepping past these exceptions takes 8-12 seconds, about as much as disabling and re-enabling first-throw exception breaking. Over a day of stopping and starting my program it really adds up.

[...]

Issue #1737, opened 2018-09-01
Improve the Initialization of RxUI


glennawatson:
The Initialization of RxUI is a bit clunky at the moment

https://github.com/reactiveui/rfcs/issues/14

Implement a change around the discussion in the RFC.

glennawatson:
Ana came up with a point that at the moment on the platforms it works it works. 

On the original RFC there was the following comment

> So instead of something like options.UseDefaults() we would add some additional verbosity by splitting options out into their own functions?
> 
> Like
> 
> RxUIOptions options = new RxUIOptions();
> options.UsePlatformSchedulers();
> options.UsePlatformBindingConverters();
> options.UsePlatformActivationForViewFetcher();
> // etc..
> Presumably the equivalent to options.UseDefaults() is simply RxApp.Start()? So what's the difference between this:
> 
> RxApp.Start(new RxUIOptions());
> and this:
> 
> RxApp.Start();
> or even this:
> 
> RxApp.Start(null);
> Is the first just an "empty" RxApp whereas the second is an RxApp with all of the default platform registrations? (and the last is an ArgumentNullException?) I think if that's the case we might want to look at something like RxApp.StartDefault() instead of RxApp.Start() just to help make the behavior less surprising.
> 
> Also, would we be able to call RxApp.Start() multiple times with different options? What sort of bugs would we run into if that was supported? I would like to be able to call Start() multiple times so I can use different options for different unit tests, but calling Start() twice in a normal application seems like something that we probably don't want to support.

Just thinking we'd probably want to take advantage of the Target Framework where we can with a generic "Start()", maybe we can make this a extension method against our NuGet packages.

Also our error messages have to be really consistent if we are in a non-initialized state if possible since at the moment the users really have to hunt the documentation.

RLittlesII:
I think the over all plan is to give the flexibility of some of the .NET Core Startup concepts, but keep a simple `RxApp.InitializeWPF()` option.  This would institute a huge breaking change and for consumers that go from not having to do anything to initialize ReactiveUI to having to provide an entire Startup file might feel overwhelming.  So providing a default `init` method per platform that gives you OOTB what you get today is ideal.

InquisitorJax:
I know this is an old thread, but do we have an eta on delivering this?
I think it preferable to have an init() method implemented rather than having to click continue 3 or 4 times for FileNotFoundException every time I start an app with CLR exception management turned on.
[...]

PR #2396, opened 2020-04-27
Allow configuring current UI framework to avoid exceptions during dependency registrations

JesperTreetop:
A small feature: A configuration knob to cull the list of registration classes that dependency registration attempts to find, so as to avoid exceptions that are thrown and then caught in the background on every application launch, which has visible effects sometimes if first-chance exceptions are being caught. See #2395 for related discussion.

This is a short-term feature, and the problem will hopefully be addressed in greater detail by construction as part of init being improved (#1737).

Speculatively load all registration classes in all assemblies.

**What is the new behavior?**
Have a list of "registration namespaces" which can be overridden by the user. The API is not ideal because the API in reality ends up being called from RxApp's static constructor. In real application usage, any parameters has to be passed out-of-band in some way by a method called before the RxApp static constructor and separate from the method where they are used, and so that is what the code does.

**What might this PR break?**
If setting the registration namespaces was mandatory it would break every app. The list defaults to all of them, which is backwards compatible with the current behavior.

**Other information**:
I have added a test which verifies that specifying WPF and running the DependencyResolverMixin will work just the same. It does not check that other registration classes are excluded.

### DIFF
--- a/src/ReactiveUI/Mixins/DependencyResolverMixins.cs
+++ b/src/ReactiveUI/Mixins/DependencyResolverMixins.cs
@@ -4,6 +4,7 @@
 // See the LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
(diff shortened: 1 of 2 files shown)


src/ReactiveUI/Mixins/DependencyResolverMixins.cs
```diff
@@ -19,6 +20,73 @@ namespace ReactiveUI
...
+        private static IReadOnlyList<RegistrationNamespace> defaultRegistrationNamespaces =
+            (RegistrationNamespace[])Enum.GetValues(typeof(RegistrationNamespace));
+
+        private static HashSet<RegistrationNamespace> registrationNamespacesToInitialize =
+            new HashSet<RegistrationNamespace>(defaultRegistrationNamespaces);
+
+        /// <summary>
+        /// Platforms or other registration namespaces for the dependency resolver to consider when initializing.
+        /// </summary>
+        public enum RegistrationNamespace
```
glennawatson, line 32: Would prefer this to be a separate non-nested enum.

src/ReactiveUI/Mixins/DependencyResolverMixins.cs
```diff
@@ -28,16 +96,22 @@ public static class DependencyResolverMixins
         [SuppressMessage("Globalization", "CA1307: operator could change based on locale settings", Justification = "Replace() does not have third parameter on all platforms")]
         public static void InitializeReactiveUI(this IMutableDependencyResolver resolver)
         {
-            var extraNs = new[]
+            var possibleNamespaces = new Dictionary<RegistrationNamespace, string>
             {
-                "ReactiveUI.XamForms",
```
glennawatson, line 33: The order we previously looked was used as part of our previous tests. It may be best to instead of using a dictionary to use a List. It's only 6 entries so a dictionary probably wouldn't perform better anyway.

src/ReactiveUI/Mixins/DependencyResolverMixins.cs
JesperTreetop: I considered that and didn't know in which file it would live - is a neighboring file okay or should it live in some other folder?

JesperTreetop:
It looks like I'll have to continue this in a bit. The full build and therefore the test don't run on my machine because I'm missing a lot of the frameworks, and the test did do something other than what I was expecting, so I have some more work to do before this is ready.

PR #2400, opened 2020-04-29
fix: Additional Subscriptions causing out of memory exception

RLittlesII:

Resolves: #2377

When you run the sample provided in the report after a few updates, Blazor will throw an `OutOfMemoryException`.

**What is the new behavior?**

The sample no longer throw the exception.

**What might this PR break?**

`ReactiveComponentBase`

**Other information**:

I tested my changes on a branch of the provided sample.  I will note that even though I am now disposing of the subscription with a composite disposable, I stopped seeing the issue as soon as I moved it into the `firstRender` check.

I let the app run for over an hour and did not experience the issue.  The most I did see logged on the memory footprint was a call to the GC that the nursery is full.

`GC_MINOR: (Nursery full) time 5.00ms, stw 5.00ms promoted 287K major size: 1152K in use: 409K los size: 1024K in use: 16K`

https://github.com/RLittlesII/BlazorReactiveUI/tree/fix/memory-leak

### DIFF
--- a/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
+++ b/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
@@ -27,6 +27,7 @@ namespace ReactiveUI.Blazor
         private readonly Subject<Unit> _initSubject = new Subject<Unit>();
         [SuppressMessage("Design", "CA2213: Dispose object", Justification = "Used for deactivation.")]
         private readonly Subject<Unit> _deactivateSubject = new Subject<Unit>();
+        private readonly CompositeDisposable _compositeDisposable = new CompositeDisposable();
 
         private T _viewModel;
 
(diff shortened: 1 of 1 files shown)


src/ReactiveUI.Blazor/ReactiveComponentBase.cs
```diff
@@ -87,15 +88,16 @@ protected override void OnAfterRender(bool firstRender)
             {
                 // The following subscriptions are here because if they are done in OnInitialized, they conflict with certain JavaScript frameworks.
                 var viewModelChanged =
-                    this.WhenAnyValue(x => x.ViewModel)
-                        .Where(x => x != null)
-                        .Publish()
-                        .RefCount();
+                    this.WhenAnyValue(x => x.ViewModel);
```
glennawatson, line 92: Both need Where(x != null). Avoid the Skip below. 

Add RefCount(2) and that should avoid the two observables being made.

PR #2402, opened 2020-05-02
Improved TransitioningContentControl animations

luisllamasbinaburo:
Improved TransitioningContentControl animations in WPF
- Added more transitions (Fade, Move, Slide, Drop, Bounce)
- Added DirectionProperty (Left, Right, Down, Up)
- Added DurationProperty
- Remove TransitionPart

```xaml
 <reactiveUi:RoutedViewHost x:Name="RoutedViewHost" Transition="SlideLeft"/>
```

**What is the new behavior?**
```xaml
 <reactiveUi:RoutedViewHost x:Name="RoutedViewHost" Transition="Slide" Direction="Right" />
```

**What might this PR break?**

**Other information**:

### DIFF
--- a/src/ReactiveUI.Wpf/Themes/Generic.xaml
+++ b/src/ReactiveUI.Wpf/Themes/Generic.xaml
@@ -30,7 +30,7 @@
                                     </Storyboard>
                                 </VisualState>
 
-                                <VisualState x:Name="FadeTransition_OutIn">
+                                <VisualState x:Name="Transition_Fade">
                                     <Storyboard>
                                         <DoubleAnimation 
                                             BeginTime="00:00:00" Duration="00:00:00.3"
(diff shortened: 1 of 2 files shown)


glennawatson:
Hey there,

Is there any way of supporting the old animations so not to break existing users? I don't think many users would but be worth having backwards compatibility unless there is a great reason not to.

luisllamasbinaburo:
Hi, I think a lot about that, but found no "clean" solution.

Currently we have;
Fade,SlideLeft

glennawatson:
No problem, I will merge it in and bump the version number then. Thanks for the contribute.

luisllamasbinaburo:
Hi, I think a lot about that, but found no "clean" solution.

Currently we have;
- TransitionType: Fade, SlideLeft, FadeDown
- TransitionPart: In, Out, InOut

But it is a little mess. For example, Fade can only be InOut, FadeDown can be In, Out, SlideLeft can be In, Out, but no InOut. 

Now we have:
- TransitionType: Fade, Slide, Move, Drop, Bounce
- TransitionDirection: Left, Right, Up, Down

Of course, new Slide + Left is the same that old SlideLeft
But to keep full compatibility, we will need to keep SlideLeft in TransformationTypeEnum
So it will be
- TransitionType: Fade, Slide, Move, Drop, Bounce + SlideLeft, FadeDown

And I think is most confusing.

What I have done is to keep default values as similar as I can.
For example, Default transistion is Fade, so is user have no change de property Transition in his RoutedViewHost, he will obtain the same behavior.
Or, if the user used SlideLeft transition, he only have to change "SlideLeft" with new "Slide", and will have the same animation again.

PR #2403, opened 2020-05-06
Fix Bug #2382 DelayChangeNotifications does not delay PropertyChanged events

### LINKED ISSUE #2382: [BUG] DelayChangeNotifications does not delay PropertyChanged events
**Describe the bug**
The `ReactiveObject.DelayChangeNotifications()` method does not delay the firing of `PropertyChanged` events. (Please note `ReactiveObject.SuppressChangeNotifications()` does suppress them.)

**Steps To Reproduce**
The following C# console app demonstrates the bug. All assertions are expected to pass, but 2 of 3 fail.
```csharp
using System.Diagnostics;
using ReactiveUI;

class Example : ReactiveObject
{
    private int _value;

    public int Value
    {
        get => _value;
        set => this.RaiseAndSetIfChanged(ref _value, value);
    }
}

class Program
{
    static void Main(string[] args)
    {
        var obj = new Example();
        var changeEvents = 0;
        obj.PropertyChanged += (s, e) => changeEvents++;
        using (obj.DelayChangeNotifications())
        {
            obj.Value = 1;
            obj.Value = 2;
            Debug.Assert(changeEvents == 0, "Property change raised during delay");
        }
        Debug.Assert(changeEvents < 2, "Property change not deduped");
        Debug.Assert(changeEvents != 0, "Property change not raised");
    }
}
```

**Expected behavior**
Similar to how `ReactiveObject.SuppressChangeNotifications()` affects property changed events and `ReactiveObject.DelayChangeNotifications()` delays and removes duplicate change events from the `ReactiveObject.Changed` observable. I expect:

1. No property changing and changed events will be raised between the call of `ReactiveObject.SuppressChangeNotifications()` and the disposing of the returned `IDisposable`.
2. When the `IDisposable` is disposed of, any delayed changing and changed events will be raised
3. Changing and changed events that are delayed will be deduplicated so that if a property is changed more than once during the delayed time, only one property change event will be raised for it.

**Environment**
- OS:       Windows Pro
- Version   10.0.18363
- Device:   Dell Vostro 7590
- ReactiveUI: v11.3.1
- .NET Framework: v4.7.2

WalkerCodeRanger:
Fixes #2382 DelayChangeNotifications does not delay PropertyChanged events

Calling `DelayChangeNotifications()` on a `ReactiveObject` delays the `Changing` and `Changed` observable notifications, but not the actual `PropertyChanging` and `PropertyChanged` events.

**What is the new behavior?**
`DelayChangeNotifications()` now delays `PropertyChanging` and `PropertyChanged` events the same as the observables. Duplicate events are also removed the same as with the observables.

**What might this PR break?**
*Code relying on the current behavior of `PropertyChanging` and `PropertyChanged` events not being delayed.*

While I don't think this will break stuff, the most likely places for other bugs to be introduced:

* Exception handling for exceptions thrown from `PropertyChanging` and `PropertyChanged` events
* The ordering of change notifications in the `Changing` and `Changed` observables
* The raising of `PropertyChanging` and `PropertyChanged` events

  * API docs for `DelayChangeNotifications()` was already not very specific, so didn't need to change.

**Other Information**:
The approach used here was discussed in issue #2382. See specifically [my comment](https://github.com/reactiveui/ReactiveUI/issues/2382#issuecomment-619299798) and those follow.

### DIFF
--- a/src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
+++ b/src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
@@ -44,12 +44,13 @@ namespace ReactiveUI
             IObservable<IReactivePropertyChangedEventArgs<TSender>> Changed { get; }
 
             /// <summary>
-            /// Gets a observable when a exception is thrown.
+            /// Gets a observable for when an exception is thrown.
             /// </summary>
             IObservable<Exception> ThrownExceptions { get; }
 
             /// <summary>
-            /// sdfsdg.
+            /// Subscribe raise property changing events to a property changing
+            /// observable. Must be called before raising property changing events.
             /// </summary>
             void SubscribePropertyChangingEvents();
 
@@ -60,7 +61,8 @@ namespace ReactiveUI
             void RaisePropertyChanging(string propertyName);
 
             /// <summary>
-            /// sdf sdfs fgdf.
+            /// Subscribe raise property changed events to a property changed
+            /// observable. Must be called before raising property changed events.
             /// </summary>
             void SubscribePropertyChangedEvents();
 
--- a/src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
+++ b/src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
@@ -5,8 +5,8 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel;
 using System.Diagnostics.Contracts;
+using System.Linq;
 using System.Reactive;
 using System.Reactive.Concurrency;
 using System.Reactive.Disposables;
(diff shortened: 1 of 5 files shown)


src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
```diff
@@ -48,12 +48,22 @@ private interface IExtensionState<out TSender>
             /// </summary>
             IObservable<Exception> ThrownExceptions { get; }
 
+            /// <summary>
+            /// sdfsdg.
```
glennawatson, line 52: We probably want a real comment here

src/ReactiveUI/ReactiveObject/ReactiveObject.cs
```diff
@@ -29,14 +33,54 @@ public ReactiveObject()
...
+                _propertyChanged += value;
+            }
+            remove => _propertyChanged -= value;
+        }
+
+        [SuppressMessage(
+            "StyleCop.CSharp.NamingRules",
+            "SA1300:Element should begin with upper-case letter",
+            Justification = "Event is private backing field for PropertyChanging")]
+        private event PropertyChangingEventHandler _propertyChanging;
```
glennawatson, line 77: Probably best to stick to the naming scheme regardless if it's private.

src/ReactiveUI/ReactiveObject/IReactiveObjectExtensions.cs
WalkerCodeRanger: I absolutely agree. That was an oversight on my part. Sorry about that. I'll fix this one and the other one below.

glennawatson:
I'll have a full review over the next hour or so. Thanks for submitting the pr

glennawatson:
Ok, tested your changes they look good. One you fix up those comments/style issues we are good to go :)

WalkerCodeRanger:
All the code review comments have been addressed.

glennawatson:
I'll push out a new version later today. Thanks again.

WalkerCodeRanger:
Great, thank you for being very responsive and helpful.

PR #2428, opened 2020-06-03
feature: Add C# 8 Nullability

RLittlesII:

Adds C# Nullability Support

Not nullability support

**What is the new behavior?**

All the nulls!

**What might this PR break?**

A lot?  Technically nothing that isn't already potentially broken.

**Other information**:


src/ReactiveUI/Routing/RoutingState.cs
```diff
@@ -83,14 +85,14 @@ public IScheduler Scheduler
         /// Gets or sets a command which will navigate back to the previous element in the stack.
         /// </summary>
         [IgnoreDataMember]
-        public ReactiveCommand<Unit, Unit> NavigateBack { get; protected set; }
+        public ReactiveCommand<Unit, Unit>? NavigateBack { get; protected set; }
```
glennawatson, line 88: Don't think these ones should ever be null. It's due to the `SetupRx()` method being called in the constructor. Let's just make sure they are constructed.

src/ReactiveUI/RxApp.cs
```diff
@@ -56,14 +58,14 @@ public static class RxApp
...
         [ThreadStatic]
-        private static IScheduler _unitTestTaskpoolScheduler;
-        private static IScheduler _taskpoolScheduler;
+        private static IScheduler? _unitTestTaskpoolScheduler;
+        private static IScheduler? _taskpoolScheduler;
         [ThreadStatic]
-        private static IScheduler _unitTestMainThreadScheduler;
-        private static IScheduler _mainThreadScheduler;
-        private static IObserver<Exception> _defaultExceptionHandler;
+        private static IScheduler? _unitTestMainThreadScheduler;
```
glennawatson, line 64: Not sure if we want these to be nullable. Again probably should be set from the static constructor some how.

glennawatson:
Sorry @rlittlesii will need you to rerun the API change files.

src/ReactiveUI/Routing/RoutingState.cs
```diff
@@ -13,6 +13,7 @@
 using System.Runtime.Serialization;
 using DynamicData;
 using DynamicData.Binding;
+#pragma warning disable 8618
```
glennawatson, line 16: Remove `SetupRX()` and put it in the constructor, suspect that will make your nullables work without the need for this pragma.

Issue #2258, opened 2019-11-26
[BUG] WPF - ViewModelViewHost creates the old view again when switching to a new view model


TolikPylypchuk:
**Describe the bug**
When a view model is changed for a `ViewModelViewHost`, it creates the old view again and only after that creates a new view, which can cause brief flickering of views. This happens in a WPF app on .NET Core 3.0 (not sure about other frameworks). ReactiveUI version is 10.5.31.

**Steps To Reproduce**
I've created a simple [repo](https://github.com/TolikPylypchuk/BugReactiveUIViewModelViewHost) to reproduce the issue.

The view flickering is not actually visible there, but if you debug the app in Visual Studio, you'll see the log output. When switching the view from View1 to View2, the following log can be seen:

```
MainViewModel: Switching the view.
View1: Creating the view 1
View2: Creating the view 2
```

**Expected behavior**
The old view is not created again when switching to a new view.

**Environment**
- OS: Windows 10 Home 1909

**Additional context**

The source of error is probably [in the constructor](https://github.com/reactiveui/ReactiveUI/blob/master/src/ReactiveUI/Platforms/windows-common/ViewModelViewHost.cs#L109) of the `ViewModelViewHost` class:

```
var contractChanged = _updateViewModel.Select(_ => ViewContractObservable).Switch();
var viewModelChanged = _updateViewModel.Select(_ => ViewModel);

var vmAndContract = contractChanged.CombineLatest(viewModelChanged, (contract, vm) => new { ViewModel = vm, Contract = contract });

vmAndContract.Subscribe(x => ResolveViewForViewModel(x.ViewModel, x.Contract));
```

I've debugged this code, and `x => ResolveViewForViewModel(x.ViewModel, x.Contract)` is hit twice  when switching the view model - first time with the old view model (hence creating the old view) and the second time with the new view model. I think this is becuase when a view model is changed, the `contractChanged` is fired, and then the `viewModelChanged` is fired, which is why `vmAndContract` is fired twice.

Brimce:
Hi,
This bug can also be observed on the RoutedViewHost.
I have found it comes from the TransitioningContentControl which inherits ContentControl.

Navigation with RoutedViewHost or change of ViewModelViewHost.ViewModel has two consequences :
1. Change of Content property inherited from the ContentControl. (**Dispose the old view**)
2. Start of transition animation. (**Creates the old view again**)

To fix this i've first tried to inherits from Control and create my own Content property. 
[...]

PR #2447, opened 2020-07-09
fix: windows-common ViewModelViewHost creating the old view again whe…

HrlecMatej:
fix: windows-common ViewModelViewHost creating the old view again when switching to a new view model (#2258)

Bug fix

fixes https://github.com/reactiveui/ReactiveUI/issues/2258
[windows-common ViewModelViewHost](https://github.com/reactiveui/ReactiveUI/blob/main/src/ReactiveUI/Platforms/windows-common/ViewModelViewHost.cs) creates the old view again when switching to a new view model:
The way it was written before, a change of either ViewModelProperty or ViewContractObservableProperty resulted in both firing their corresponding Observables. Because of CombineLatest this resulted in the old ViewModel being run through ResolveViewForViewModel again, when the ViewModelProperty was changed. This created the old View for a moment again, which is a waste of resources, and it results in animation flickering (and other issues with DI lifetime management which I use).

**What is the new behavior?**

Now each Dependency property only fires their own Observable, instead of both at the same time.
The implementation has been mostly copied from [XamForms ViewModelViewHost](https://github.com/reactiveui/ReactiveUI/blob/main/src/ReactiveUI.XamForms/ViewModelViewHost.cs)

**What might this PR break?**
It's a simple change, that has been taken from another part of ReactiveUI, so I find it unlikely to break anything.

I cannot run the ReactiveUI.Tests, because I am missing build tools for all the multitude of test targets. I do not think this class even has any tests anyhow.
I would not know, where to look for that. Besides, this bug fix makes ViewModelViewHost ViewModelProperty follow expected behavior, so no documentation changes should be necessary.

**Other information**:
I suggest ReactiveUI's team to do an audit of the 3 ViewModelViewHost implementations, with the goal of putting shared code in a common class. Currently, there is a lot of repetition and a showcase, how to write the same thing in 3 different ways (ViewLocator part of the code being the biggest offender).

### DIFF
--- a/src/ReactiveUI/Platforms/windows-common/ViewModelViewHost.cs
+++ b/src/ReactiveUI/Platforms/windows-common/ViewModelViewHost.cs
@@ -52,14 +52,15 @@ namespace ReactiveUI
         /// The view model dependency property.
         /// </summary>
         public static readonly DependencyProperty ViewModelProperty =
-            DependencyProperty.Register(nameof(ViewModel), typeof(object), typeof(ViewModelViewHost), new PropertyMetadata(default(object)));
+            DependencyProperty.Register(nameof(ViewModel), typeof(object), typeof(ViewModelViewHost), new PropertyMetadata(null, SomethingChanged));
 
         /// <summary>
         /// The view contract observable dependency property.
         /// </summary>
         public static readonly DependencyProperty ViewContractObservableProperty =
-            DependencyProperty.Register(nameof(ViewContractObservable), typeof(IObservable<string>), typeof(ViewModelViewHost), new PropertyMetadata(Observable<string>.Default));
+            DependencyProperty.Register(nameof(ViewContractObservable), typeof(IObservable<string>), typeof(ViewModelViewHost), new PropertyMetadata(Observable<string>.Default, SomethingChanged));
 
+        private readonly Subject<Unit> _updateViewModel = new Subject<Unit>();
         private string _viewContract;
         private bool _isDisposed;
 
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/reactiveui/ReactiveUI?pullRequest=2447) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/reactiveui/ReactiveUI?pullRequest=2447) before we can accept your contribution.<br/><br/>:x: HrlecMatej [sign now](https://cla.dotnetfoundation.org/reactiveui/ReactiveUI?pullRequest=2447)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/reactiveui/ReactiveUI?pullRequest=2447) it.</sub>

src/ReactiveUI/Platforms/windows-common/ViewModelViewHost.cs
```diff
@@ -106,15 +105,17 @@ public ViewModelViewHost()
                 .StartWith(platformGetter())
                 .DistinctUntilChanged();
 
-            var contractChanged = _updateViewModel.Select(_ => ViewContractObservable).Switch();
-            var viewModelChanged = _updateViewModel.Select(_ => ViewModel);
+            // Observable that fires when ViewModel or ViewContractObservable change, or ViewContractObservable fires
+            var vmAndContract = Observable.CombineLatest(
```
glennawatson, line 109: This is going to cause memory leaks unfortunately due to the way dependency properties work. Eg WhenAnyValue will call into the static manager, and it'll hold an instance of the class until you dispose the WhenAnyValue.

This is the reason why the SomethingChanged exists since it doesn't suffer from this problem.

Happy to take a solution that keeps the SomethingChanged approach.

HrlecMatej: Did not know that could be an issue. Is that not a problem in https://github.com/reactiveui/ReactiveUI/blob/main/src/ReactiveUI.XamForms/ViewModelViewHost.cs as well then?

glennawatson:
Also in note regards to integrating the Xamarin Forms/Windows version, we can't do the xamarin forms version on windows due to the fact Xamarin Forms will use INotifyPropertyChanged approach for alerting you when BindableProperty's change.

src/ReactiveUI/Platforms/windows-common/ViewModelViewHost.cs
glennawatson: Xamarin Forms is different since they have BindableProperty invoke a INotifyPropertyChanged so have a non-memory leaking way of getting alerted of property changes.

glennawatson:
Thanks for the quick update. Looks heaps better thanks.

HrlecMatej:
So, I reverted my first commit, and simply added a new Subject for ViewContract changes and separate PropertyChangedCallbacks.
This will still fix the issue of triggering both properties at the same time, while preventing the memory leak you mentioned.

glennawatson:
The build error is due to a recent change in Splat, so will fix it. Thanks.

HrlecMatej:
After doing some tests with my project, I found a subtle bug changing ViewContract change notification caused.
The ViewContractObservable was previously set before the CombineLatest operator. That causes value change callback on ViewContractObservableProperty to fire before CombineLatest. This meant, if we never set ViewContract afterwards, CombineLatest would never emit a value, since _updateViewContract would notify of a change before anyone was listening.
This commit fixes that by simply moving the ViewContractObservable setup to the end.
Alternatively, we could also make _updateViewContract into a ReplaySubject(1).

glennawatson:
I'll do a release by the end of the week coming. Thanks for getting this in and testing it for me.

Just want to get some events bugs fixed first before the release.

You can use the GitHub package nuget server if you need it sooner though.

HrlecMatej:
You're welcome. Also, thank you for being so responsive.

PR #2453, opened 2020-07-23
feature: Add Injectable component base for blazor.

glennawatson:
Blazor has a option of using the Microsoft DI to inject properties onto a Component.

This allows you to inject your view model onto your view.


richbryant:
this is going to be extremely handy.

PR #2463, opened 2020-07-29
housekeeping: Minor solution/setting file cleanup

kronic:
1. Cleanup stylecop.json
2. Fix missing path in solution
3. Cleanup missing resources


glennawatson:
If you don't mind wouldn't mind keeping this one to keep for simpler code reviews. 

Your other pr is more complex and will take time to review.

PR #2468, opened 2020-07-30
Fix: Make Bind() ViewModel parameters nullable.

Noggog:
This PR modifies the `PropertyBindingMixins` used mainly within `ReactiveUserControl` codebehind `WhenActivated` blocks.
It changes the `viewModel` members to be nullable.

Bit of contextual overview:
The ViewModel parameter in `Bind` is mostly unused.  The reason it exists is to simplify generic specification of the call.  By passing the ViewModel in, the generics can be implied, whereas if the parameter was omitted, the user would have to type out the verbose generic args themselves for every call.  Otherwise, the parameter is mostly unneeded, and theoretically could be removed.

99% of the time `Bind` is used, `ReactiveUserControl.ViewModel` member is what is passed in.  The current misalignment in nullability, though, means the user has to pass in the ViewModel with the override operator:

`this.Bind(this.ViewModel!, ...`

## What is the new behavior?
Since the ViewModel is unused, and just there for typing, it doesn't actually matter if the parameter nullable or not.  

**So why make it nullable?**  
By shifting the parameter to be nullable, it aligns with the `ReactiveUserControl.ViewModel` member, meaning the call is cleaner for the end user:

`this.Bind(this.ViewModel, ...`

This is much cleaner for the user, and is less confusing, as requiring a `!` operator makes the user question if they're doing something wrong, and/or tempts them to need to check that `ViewModel` is not null first.. which we know is an unnecessary step.

This API shift takes the burden/questions off the user on if they're doing something wrong, and lets them use `Bind` more as they would expect.

## What might this PR break?
I added ArgumentNullExceptions to check that the viewmodel is not null.   We know in typical usage it will not be null, as WhenActivated will never be called with null ViewModels, which is where Bind is being invoked from.   Still, it is good to check that it is not null internally anyway, as it definitely means something is in an error state, as this should never happen.

One reason to do so, is that the ViewModel parameter IS used in one spot:  It's given to the ReactiveBinding class that is returned, which has a ViewModel member, which is not nullable.  It is good to check that the VM is not null, just as it's an error state if that was to ever happen, and it means we safely fill this member with a viewmodel object that is not null, as it expects.

One note is that the current methodology of passing in `this.ViewModel!` is no safer, we're still potentially "lying" and setting a null view model object to ReactiveBinding's non-null member.  The ArgumentNullException is actually a bit safer on that front, nipping an odd situation in the bud early.

## Summary
I think this change gives the user an experience they're expecting, while retaining the safety of the system overall.

I had some remaining errors related to some missing SDKs for some of the more obscure platform targets.  I think everything is okay.  We'll let the build system double check, though.

### DIFF
--- a/src/ReactiveUI.Winforms/ContentControlBindingHook.cs
+++ b/src/ReactiveUI.Winforms/ContentControlBindingHook.cs
@@ -17,7 +17,7 @@ namespace ReactiveUI.Winforms
     public class ContentControlBindingHook : IPropertyBindingHook
     {
         /// <inheritdoc/>
-        public bool ExecuteHook(object source, object target, Func<IObservedChange<object, object>[]> getCurrentViewModelProperties, Func<IObservedChange<object, object>[]> getCurrentViewProperties, BindingDirection direction)
+        public bool ExecuteHook(object? source, object target, Func<IObservedChange<object, object>[]> getCurrentViewModelProperties, Func<IObservedChange<object, object>[]> getCurrentViewProperties, BindingDirection direction)
         {
             if (getCurrentViewProperties == null)
             {
--- a/src/ReactiveUI.Winforms/ContentControlBindingHook.cs
+++ b/src/ReactiveUI.Winforms/ContentControlBindingHook.cs
@@ -17,7 +17,7 @@ namespace ReactiveUI.Winforms
     public class ContentControlBindingHook : IPropertyBindingHook
     {
         /// <inheritdoc/>
-        public bool ExecuteHook(object target, Func<IObservedChange<object, object>[]> getCurrentViewModelProperties, Func<IObservedChange<object, object>[]> getCurrentViewProperties, BindingDirection direction)
+        public bool ExecuteHook(object source, object target, Func<IObservedChange<object, object>[]> getCurrentViewModelProperties, Func<IObservedChange<object, object>[]> getCurrentViewProperties, BindingDirection direction)
         {
             if (getCurrentViewProperties == null)
             {
--- a/src/ReactiveUI.Winforms/ContentControlBindingHook.cs
+++ b/src/ReactiveUI.Winforms/ContentControlBindingHook.cs
@@ -17,7 +17,7 @@ namespace ReactiveUI.Winforms
     public class ContentControlBindingHook : IPropertyBindingHook
     {
         /// <inheritdoc/>
-        public bool ExecuteHook(object source, object target, Func<IObservedChange<object, object>[]> getCurrentViewModelProperties, Func<IObservedChange<object, object>[]> getCurrentViewProperties, BindingDirection direction)
+        public bool ExecuteHook(object target, Func<IObservedChange<object, object>[]> getCurrentViewModelProperties, Func<IObservedChange<object, object>[]> getCurrentViewProperties, BindingDirection direction)
         {
             if (getCurrentViewProperties == null)
             {
--- a/src/ReactiveUI/Interfaces/IPropertyBindingHook.cs
+++ b/src/ReactiveUI/Interfaces/IPropertyBindingHook.cs
@@ -23,7 +23,7 @@ namespace ReactiveUI
         /// <param name="getCurrentViewProperties">Get current view properties.</param>
         /// <param name="direction">The Binding direction.</param>
         bool ExecuteHook(
-            object source,
+            object? source,
             object target,
             Func<IObservedChange<object, object>[]> getCurrentViewModelProperties,
             Func<IObservedChange<object, object>[]> getCurrentViewProperties,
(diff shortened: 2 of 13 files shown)


glennawatson:
Sorry, didn't mean to close the PR. Will try out and let you know what I find.

src/ReactiveUI/Interfaces/IPropertyBindingHook.cs
```diff
@@ -17,13 +17,11 @@ public interface IPropertyBindingHook
         /// Called when any binding is set up.
         /// </summary>
         /// <returns>If false, the binding is cancelled.</returns>
-        /// <param name="source">The source ViewModel.</param>
         /// <param name="target">The target View (not the actual control).</param>
         /// <param name="getCurrentViewModelProperties">Get current view model properties.</param>
         /// <param name="getCurrentViewProperties">Get current view properties.</param>
         /// <param name="direction">The Binding direction.</param>
         bool ExecuteHook(
-            object source,
```
glennawatson, line 26: This is the part that concerns me about this PR with a breaking API change (not related to nullability)

src/ReactiveUI.Winforms/ContentControlBindingHook.cs
```diff
@@ -17,7 +17,7 @@ namespace ReactiveUI.Winforms
     public class ContentControlBindingHook : IPropertyBindingHook
     {
         /// <inheritdoc/>
-        public bool ExecuteHook(object source, object target, Func<IObservedChange<object, object>[]> getCurrentViewModelProperties, Func<IObservedChange<object, object>[]> getCurrentViewProperties, BindingDirection direction)
+        public bool ExecuteHook(object target, Func<IObservedChange<object, object>[]> getCurrentViewModelProperties, Func<IObservedChange<object, object>[]> getCurrentViewProperties, BindingDirection direction)
```
glennawatson, line 20: Can you please remove this from this PR and make it a separate PR, doesn't seem to be related to the nullability factor.

Noggog: Sure, can do.  I agree it was a tangential change

Noggog:
Sorry about the overeager change.

Reversed it, and the overall changes should be much cleaner now.  Let me know if you want a flatten or anything. 

The null check /w ArgumentNullException throw made the pass to ExecuteHook clean anyway.  So definitely better to leave as is.

glennawatson:
It's quite common in Xamarin Forms app btw that the "ViewModel" may be null initially btw. Eg on those platforms WhenActivated isn't required always.

So the `ArgumentNullException` could cause issues for that platform.

Noggog:
Alright, so then the solution might be to just remove ArgumentNullException, and utilize `!` internally.  

The only potential issue I can see is that the returned ReactiveBinding will have a `TViewModel ViewModel` member set to null.  This is the current behavior anyway, I imagine.  Still sort of a lie to say it wont be null when it potentially can be.  Perhaps we'd want to flip `IReactiveBinding.ViewModel` to be nullable to be more accurate for those situations you mentioned in Xamarin?

glennawatson:
If you have the viewModel nullable without the ArgumentNullException, does that compile?

That should allow users to keep their current syntax, or at least use ? right.

Noggog:
Removing the ArgumentNullException by itself will make it not compile.  This is because the ReactiveBinding we're returning wants a viewmodel that is not null.
So either we need to:
1)  check that it isn't null, and throw if it is
2)  Modify ReactiveBinding (and ExecuteHook) to have nullable viewmodel members
3)  Shove it in and lie with a `!`

glennawatson:
Yeah, 2 is fine as long as it doesn't change existing public api (apart from nullability)

glennawatson:
Thanks for putting together this PR, know there are some complications from it.

Noggog:
My pleasure, I don't mind putting in the work.  Part of the reason for the quick PR was to get some actual code changes to look at and discuss for clarity.  I expected to have to make some adjustments on the base.

Pushed up a new commit with the ArgumentNullExceptions removed, and the appropriate members from ReactiveBinding and IPropertyBindingHook made nullable.

Seems clean enough

glennawatson:
Will do some tests but this looks good. Thanks

Noggog:
Are these failing tests something I need to address?  Or are they part of the approval process with the different technologies?

glennawatson:
If you have a windows machine run the tests and change the .received extension to .approved on the API approval tests.

glennawatson:
Attempted to do this one for you but didn't have permissions. Here's a patch file for you.
[api_approval.zip](https://github.com/reactiveui/ReactiveUI/files/4999717/api_approval.zip)

Noggog:
Thanks for the diff patch!  I would've been quite lost looking around for those on my lonesome, haha.

gtbuchanan:
Is there a particular reason this wasn't done for command bindings as well (i.e. `BindCommand`)? I assume it was just an oversight but just wanted to make sure before I submit another issue/PR.

cabauman:
Hi @gtbuchanan . There's a related issue open, and the last comment left by @glennawatson mentioned he was working on something. Maybe he can give an update.

Noggog:
@gtbuchanan I didn't avoid them for any specific reason.  Just an oversight as I don't typically use those calls.  If they seem like a good candidate for the same refactor I'd say go for it

PR #2470, opened 2020-07-30
Add pragma warning disable 8600, 8605, 8603, 8613, 8619

kronic:
Add pragma warning disable 8600, 8605, 8603, 8613, 8619


kronic:
@glennawatson ready

PR #2504, opened 2020-08-27
Fix/memory leaks

### LINKED ISSUE #2384: Activation closures keep viewmodel alive
**Describe the bug**

When dynamically switching a view's viewmodel, the first viewmodel instance is being kept alive. 

Encountered the issue while memory profiling a WPF app on .NET Framework 4.6.1. 
ReactiveUI version is 11.3.1.

**Steps To Reproduce**
I've created a simple [repo](https://github.com/7unk/AnimalActivation) to reproduce the issue. The view can be bound to multiple types of viewmodels (to emphasize the issue).

- On start up a `DogViewModel` is bound, the text "Dog" is displayed.
- Take base memory snapshot.
- Clicking the `Switch` button a `CatViewModel` is bound instead.
- Take second memory snapshot.
- Comparing snapshots reveals `DogViewModel` has not been released / GC'ed.

Note if we switch once more, we can verify `CatViewModel` is GC'ed correctly.

**Expected behavior**
A viewmodel that was bound (on startup) should become eligible for garbage collection when it is no longer being referenced by application code.

 

**Environment**

- OS: Windows 10 Pro
- Version: 10.0.18362 Build 18362

 

**Additional context**

Memory profiling shows that the original viewmodel is being kept alive by three roots. Two of those can be traced back to the view's activation cycle, which closes over the instance (ref blue highlighting).

EventHandler

Note I didn't use `ViewModelViewHost` to eliminate it from possible causes.

cabauman:

Bug fix. Resolves #2384

If the initial view model is assigned _before_ the binding is set up, that view model gets tied to the lifetime of the binding. This means even if the view model gets swapped out for a different one, it never gets garbage collected. This includes Bind, OneWayBind, and BindCommand. The linked issue above includes a WPF reproduction project that I based my findings and fixes off of. I also tried with a Xamarin Forms Android project and experienced the same issue, so I assume this hidden memory leak affects most, if not all, platforms.

**What is the new behavior?**

The initial view model is not tied to the lifetime of the binding.

**What might this PR break?**
Nothing

**Other information**:

There ended up being two different memory leak sources.

1) The ReactiveBinding object returned by the Bind family, because of storing the initial view model in a property.

ReactiveBindingLeak

As you can see, that binding object is usually kept alive because of storing it in the CompositeDisposable of `WhenActivated`. So this fix was simple; just remove the ViewModel property assignment and deprecate the ViewModel property.

<br />

2) Passing the [_calculated_ `ObservedChange<TSender, TValue>.Value`](https://github.com/reactiveui/ReactiveUI/blob/2251609d377e03b7c1f2e760edd4ac4847979241/src/ReactiveUI/Mixins/ReactiveNotifyPropertyChangedMixin.cs#L188)  to [`StartWith`](https://github.com/reactiveui/ReactiveUI/blob/2251609d377e03b7c1f2e760edd4ac4847979241/src/ReactiveUI/Mixins/ReactiveNotifyPropertyChangedMixin.cs#L207). 

StartWithLeak

StartWith is implemented via [`ToObservable().Concat`](https://github.com/dotnet/reactive/blob/7ad606b3dcd4bb2c6ae9622f8a59db7f8f52aa85/Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.Single.cs#L330), and [`Concat` keeps a reference](https://github.com/dotnet/reactive/blob/7ad606b3dcd4bb2c6ae9622f8a59db7f8f52aa85/Rx.NET/Source/src/System.Reactive/Linq/Observable/Concat.cs#L11) to that object until the pipeline is disposed of.

So the solution is to not calculate the kicker value _outside_ of the pipeline, and instead take advantage of the [existing `Select` statement](https://github.com/reactiveui/ReactiveUI/blob/2251609d377e03b7c1f2e760edd4ac4847979241/src/ReactiveUI/Mixins/ReactiveNotifyPropertyChangedMixin.cs#L206) which uses the `GetValue` method to calculate it lazily _within_ the pipeline.

<br />

**Memory Leak Tests**

You may notice my use of local functions in the unit tests I added. This is necessary, at least for the netcoreapp unit tests, and it's the reason why the old memory leak tests failed. Quoted from a [JetBrains blog](https://blog.jetbrains.com/dotnet/2018/10/04/unit-testing-memory-leaks-using-dotmemory-unit/):

> Since all of our logic is being run in one method (our test method), the garbage collector will not clean up local variables that are still available in the context of our function. We will have to re-write our test a little bit, so that the local variables used are considered out of scope when validating memory usage. This can be done by creating a separate method to run the code being tested, or inline using a local function or an Action.

### DIFF
--- a/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.netcoreapp3.1.approved.txt
+++ b/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.netcoreapp3.1.approved.txt
@@ -257,6 +257,8 @@ namespace ReactiveUI
         ReactiveUI.BindingDirection Direction { get; }
         TView View { get; }
         System.Linq.Expressions.Expression ViewExpression { get; }
+        [System.Obsolete("This property has been deprecated. Refer to ViewModelExpression for a representat" +
+            "ion of the bound view model.")]
         TViewModel ViewModel { get; }
         System.Linq.Expressions.Expression ViewModelExpression { get; }
     }
--- a/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.net472.approved.txt
+++ b/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.net472.approved.txt
@@ -257,6 +257,8 @@ namespace ReactiveUI
         ReactiveUI.BindingDirection Direction { get; }
         TView View { get; }
         System.Linq.Expressions.Expression ViewExpression { get; }
+        [System.Obsolete("This property has been deprecated. Refer to ViewModelExpression for a representat" +
+            "ion of the bound view model.")]
         TViewModel ViewModel { get; }
         System.Linq.Expressions.Expression ViewModelExpression { get; }
     }
--- a/src/ReactiveUI.Tests/Platforms/windows-xaml/PropertyBindingTest.cs
+++ b/src/ReactiveUI.Tests/Platforms/windows-xaml/PropertyBindingTest.cs
@@ -396,6 +396,28 @@ namespace ReactiveUI.Tests.Xaml
         }
 
         [Fact]
+        public void OneWayBindInitialViewModelShouldBeGarbageCollectedWhenOverwritten()
+        {
+            static (IDisposable?, WeakReference) GetWeakReference()
+            {
+                var vm = new PropertyBindViewModel();
+                var view = new PropertyBindView { ViewModel = vm };
+                var weakRef = new WeakReference(vm);
+                var disp = view.OneWayBind(vm, x => x.Property1, x => x.SomeTextBox.Text);
+                view.ViewModel = new PropertyBindViewModel();
+
+                return (disp, weakRef);
+            }
+
+            var (disp, weakRef) = GetWeakReference();
+
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+
+            Assert.False(weakRef.IsAlive);
+        }
+
+        [Fact]
         public void BindToWithNullStartingValueToNonNullValue()
         {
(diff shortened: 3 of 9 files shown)


src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.netcoreapp3.1.approved.txt
```diff
@@ -257,6 +257,8 @@ namespace ReactiveUI
         ReactiveUI.BindingDirection Direction { get; }
         TView View { get; }
         System.Linq.Expressions.Expression ViewExpression { get; }
+        [System.Obsolete("This property has been deprecated. Refer to ViewModelExpression for a representat" +
```
glennawatson, line 260: representat is a spelling mistake I suspect. 

`Refer to ViewModelExpression for a way to retrieve the value.` maybe.

glennawatson (changes requested):

src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.netcoreapp3.1.approved.txt
cabauman: Hmm looks correct on the declaration but the approval text gets messed up. I'll correct it. Supposed to be "This property has been deprecated. Refer to ViewModelExpression for a representation of the bound view model."

glennawatson: Ah if it's just a approval text thing then don't worry about it.

glennawatson (approved):

glennawatson (approved):

PR #2505, opened 2020-08-27
Fixes UITextField linker setter override

mfmadsen:
In the linker override for iOS, the UITextField setter was not trickered (probably because a little too much use of copy/paste).


glennawatson:
The build server is broken at the moment due to a upgrade on the server. Planning on fixing tomorrow. 

So don't be shocked if the ci fails.

glennawatson (approved):

PR #2507, opened 2020-08-30
fix: registering specific namespaces

### LINKED ISSUE #2395: Allow configuring or detecting current UI framework to avoid exceptions during dependency registrations

RLittlesII:

Fixes #2395 
Related #2396

**What is the new behavior?**

**What might this PR break?**

**Other information**:


glennawatson (approved):

JesperTreetop:
I just came back to #2396 way too late in an attempt to finish it, only to see that I'd missed this was picked up and finished in the meantime! Great job and thanks @RLittlesII and @glennawatson - it was very satisfying to add one line of code and not have to click through several assembly load exceptions after launch.

PR #2539, opened 2020-09-23
fix: Add HasRegistered check in InitializeReactiveUI for DI Containers.

### LINKED ISSUE #2537: Allow registering custom services before calling InitializeReactiveUI by adding HasRegistration check
https://github.com/reactiveui/ReactiveUI/blob/4c03cbee86a46365c7e7297e709906a157707c84/src/ReactiveUI/Mixins/DependencyResolverMixins.cs#L42-L43

Allow registering custom services before calling InitializeReactiveUI by adding HasRegistration checks, similar to how [InitializeSplat handles that](https://github.com/reactiveui/splat/blob/d0670a7fb0a7207d6361884a28da0dd5d3c977d1/src/Splat/ServiceLocation/ServiceLocationInitialization.cs#L33)

With this approach, one would register a custom service before the initialization call (like you currently do to override e.g. ILogger in InitializeSplat):
// (Autofac DI is used in the example)
// Register your overrides first
containerBuilder.RegisterType&lt;MyCustomViewLocator&gt;().As&lt;IViewLocator&gt;().SingleInstance();
// Then call InitializeReactiveUI, which would ignore already registered IViewLocator
autofacResolver.InitializeReactiveUI();

This means there aren't any redundant registrations, and no potential unexpected side-effects from that.

HrlecMatej:

Fixes #2537
Allow registering custom services before calling InitializeReactiveUI by adding HasRegistration check.

This is mostly to allow cleaner substitution of custom services when using dependency inversion. You wouldn't need to unregister or have a redundant service registration anymore.

Registers a service, without checking if a registration already exists.

**What is the new behavior?**

Now it checks, if a service with a certain type is already registered.

With this approach, one would register a custom service before the initialization call (like you currently do to override e.g. ILogger in InitializeSplat). Autofac DI is used in the example:

```cs
// Register your overrides first
builder.RegisterInstance(new NullLogger()).As<ILogger>();
builder.RegisterType<MyCustomViewLocator>().As<IViewLocator>().SingleInstance();
// etc.

// Since I am using AutofacDependencyResolver, I create it at this point
var autofacResolver = builder.UseAutofacDependencyResolver();
// Register the resolver in Autofac so it can be later resolved
builder.RegisterInstance(autofacResolver);

// Then call InitializeReactiveUI, which would ignore already registered IViewLocator
autofacResolver.InitializeReactiveUI();
```

This means there aren't any redundant registrations, and no potential unexpected side-effects from that.

**What might this PR break?**
Nothing should brake.

**Other information**:


glennawatson:
This class is changing soon so will likely have to integrate the changes elsewhere. @RLittlesII mind looking at the changes and integrating into your work.

glennawatson:
This causes issues with some DI engines so will revisit later.

PR #2546, opened 2020-10-01
Update OAPH to skip initial value when deferring subscription

### LINKED ISSUE #2124: feature: Add ToProperty overload that takes a Func<T>
This [Stack Overflow](https://stackoverflow.com/a/57146328/2088094) answer points out a limitation with current functionality to defer subscription on the initial value of an `ObservableAsPropertyHelper`

Add a `ToProperty` overload that takes a `Func<T>`

``` csharp
public static ObservableAsPropertyHelper<TRet> ToProperty<TObj, TRet>(
            this IObservable<TRet> target,
            TObj source,
            Expression<Func<TObj, TRet>> property,
            Func<TRet> initialValue = default(Func<TRet>),
            bool deferSubscription = false,
            IScheduler scheduler = null)
```

The alternative discussed in that Stack Overflow post:

``` csharp
Observable
    .Defer(() => this.WhenAnyValue(x => x._model.Value))
    .ToProperty(
        this,
        nameof(this.Property),
        out this._propertyOAPH,
        deferSubscription: true,
        scheduler: RxApp.MainThreadScheduler);
```

Evangelink:
Allow to not emit the initial value when deferring the subscription.

If you subscribe to the property change it will emit a first default value.

**What is the new behavior?**
If you subscribe to the property change it will NOT emit a first default value.

**What might this PR break?**
As far as I can tell the biggest risk, and a risk I would be facing on some projects, is that to workaround this behavior I have used `Skip(1)` and if the first value is no longer emitted then the `Skip(1)` will skip the wrong value.

**Other information**:
Fix #2124

### DIFF
--- a/src/ReactiveUI/ObservableForProperty/OAPHCreationHelperMixin.cs
+++ b/src/ReactiveUI/ObservableForProperty/OAPHCreationHelperMixin.cs
@@ -4,7 +4,6 @@
 // See the LICENSE file in the project root for full license information.
 
 using System;
-using System.Diagnostics.Contracts;
 using System.Linq.Expressions;
 using System.Reactive.Concurrency;
 
@@ -38,7 +37,7 @@ namespace ReactiveUI
         /// A value indicating whether the <see cref="ObservableAsPropertyHelper{T}"/>
         /// should defer the subscription to the <paramref name="target"/> source
         /// until the first call to <see cref="ObservableAsPropertyHelper{T}.Value"/>,
-        /// or if it should immediately subscribe to the the <paramref name="target"/> source.
+        /// or if it should immediately subscribe to the <paramref name="target"/> source.
         /// </param>
         /// <param name="scheduler">
         /// The scheduler that the notifications will be provided on - this should normally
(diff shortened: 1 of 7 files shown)


glennawatson:
I'll fix the build issues today.

src/ReactiveUI/ObservableForProperty/ObservableAsPropertyHelper.cs
```diff
@@ -162,6 +167,8 @@ public T Value
         /// </summary>
         public IObservable<Exception> ThrownExceptions => _thrownExceptions.Value;
 
+        internal /* for testing purposes */ IObservable<T> Source { get; }
```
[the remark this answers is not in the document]
glennawatson: I would just make this a XML style comment.

glennawatson (changes requested):

Evangelink:
It's weird, I am having the same issue locally (Win10 machine) but the issue was already present on a fresh clone and latest commit of main so I thought this was only a local configuration being wrong.

glennawatson:
It's due to recent changes in splat.

Evangelink:
Oh ok! Thanks! BTW, I have left a comment/question on the ticket this PR is trying to fix. Feel free to give me your opinion and if you want I can file a follow-up a PR (or add a commit to this one).

glennawatson:
Having the internal is fine. Makes the test easier.

Evangelink:
I was more thinking about this one:

> I was tempted to also give a go to the overload with the `Func` parameter but to have it having any value that means that the initial value should not be emitted right away (because in this case there is no need to take the `Func`, we can have the value directly) but only at the time where the subscription is made (i.e. inside the `Value` property). Would it make any sense?

glennawatson:
Having a func and delay evaluation was where I was wanting to go.

glennawatson:
If you go for that approach the existing overload can call the func one. And then you can just add the evaluation step when the observable gets connected.

Evangelink:
Remaining work:

- Fix API list file
- Add tests for `ToProperty` (not sure if I should add only a dummy test or do something more complex).

src/ReactiveUI/ObservableForProperty/OAPHCreationHelperMixin.cs
```diff
@@ -55,7 +54,50 @@ public static class OAPHCreationHelperMixin
...
+        /// </param>
+        /// <returns>
+        /// An initialized ObservableAsPropertyHelper; use this as the backing field
+        /// for your property.
+        /// </returns>
+        public static ObservableAsPropertyHelper<TRet> ToProperty<TObj, TRet>(
+            this IObservable<TRet> target,
+            TObj source,
+            Expression<Func<TObj, TRet>> property,
+            Func<TRet> getInitialValue,
```
Evangelink, line 97: All overloads have a non-optional `Func` parameter to avoid the runtime resolution error that would happen otherwise.

src/ReactiveUI/ObservableForProperty/ObservableAsPropertyHelper.cs
```diff
@@ -98,6 +98,44 @@ public sealed class ObservableAsPropertyHelper<T> : IHandleObservableErrors, IDi
...
+        /// </param>
+        /// <param name="scheduler">
+        /// The scheduler that the notifications will provided on - this
+        /// should normally be a Dispatcher-based scheduler.
+        /// </param>
+        public ObservableAsPropertyHelper(
+            IObservable<T> observable,
+            Action<T> onChanged,
+            Action<T>? onChanging = null,
+            Func<T>? getInitialValue = null,
```
Evangelink, line 136: I have to declare an optional parameter to avoid having an overload with the `onChanging` parameter in another position.

worldbeater:
Probably worth updating API approval tests btw. Just copy-paste the `.received.txt` file contents into the `.approved.txt` file.

Evangelink:
@worldbeater Thanks! I was delaying this to get some feedback about the "new" API. I am still not able to properly build the project locally and it seems that I cannot access the CI build logs so I am a bit in the blur to provide a better PR.

worldbeater:
The changes look good overall. Happy to see they aren't breaking existing APIs :+1: 

Also, there is a lazy and a hacky way to build and test the core parts of the project if you can't install all the required frameworks — just cut the frameworks you don't have on your PC from the `<TargetFrameworks>` section in the `.csproj` files, build and test the project, and then do `git reset **/*.csproj` before committing. This is bad advice, but this would allow you to run API approval tests locally at least.

Evangelink:
I think there is an implicit breaking change in the sense where people wanting to avoid this first notification have rely on a `Skip(1)` (at least that's what I did) and so upgrading to the new version will cause the `Skip(1)` to skip the wrong notification.

I have installed all tools/frameworks and it doesn't work, doing some tests it seems to be linked to:

1. VS2019 Preview (seems to be ok with regular one)

2. This test `AutofacDependencyResolver_Should_Register_ReactiveUI_CreatesCommandBinding` keeps breaking but it doesn't seem related to my change.

glennawatson (approved):

PR #2554, opened 2020-10-14
build(deps): Use Shouldly 4.0.0-beta0003

worldbeater:

The breaking change in Shouldly `4.0.0-beta0003` https://github.com/reactiveui/ReactiveUI/pull/2538 fails our build, so here is a fix.

Currently, we are using Shouldly `4.0.0-beta0002`, and dependabot is unable to upgrade the package due to a breaking change. 

**What is the new behavior?**

Now, the package is upgraded and we are using the new `DiffEngine` API.

The new `DiffEngine` library lives here: https://github.com/VerifyTests/DiffEngine

**What might this PR break?**

Nothing, as this is just a minor dependency upgrade.

### DIFF
--- a/integrationtests/IntegrationTests.Shared.Tests/IntegrationTests.Shared.Tests.csproj
+++ b/integrationtests/IntegrationTests.Shared.Tests/IntegrationTests.Shared.Tests.csproj
@@ -9,7 +9,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.2.0" />
-    <PackageReference Include="Shouldly" Version="4.0.0-beta0002" />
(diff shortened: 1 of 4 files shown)


worldbeater:
`
System.IO.FileLoadException : Could not load file or assembly 'Shouldly, Version=4.0.0.0, Culture=neutral, PublicKeyToken=6042cbcb05cbc941' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)
`
Strong names are attacking. Looks like an issue with Shouldly 4.0.0-beta0003...

glennawatson (approved):

glennawatson:
Being replaced by #2559

PR #2558, opened 2020-10-15
Update unsupported expression error messages

### LINKED ISSUE #1209: Improve error message: Unsupported expression type: 'x'

Enhancement.

When an unsupported expression type appears within an expression, a `NotSupportedException` is thrown with message:

> Unsupported expression type: 'type'

This is not helpful in understanding and solving the problem.

The exception message should probably include the expression that is causing the problem, along with an indication of what normally causes this. For example, the following code:

```C#
this.WhenAnyValue(x => this.Foo)
```

Triggers the exception with an expression type of "Constant". We should probably trap this case and provide a helpful clue as to why the exception is probably being thrown.

Widen the pit of success.

RxUI 7.

Evangelink:

This PR intends to improve the error message when providing an unsupported expression type in the `WhenAny` calls.

Fix #1209

**What is the new behavior?**

**What might this PR break?**

**Other information**:

### DIFF
--- a/src/ReactiveUI.Tests/WhenAny/ReactiveNotifyPropertyChangedMixinTest.cs
+++ b/src/ReactiveUI.Tests/WhenAny/ReactiveNotifyPropertyChangedMixinTest.cs
@@ -24,6 +24,8 @@ namespace ReactiveUI.Tests
 {
     public class ReactiveNotifyPropertyChangedMixinTest
     {
+        public string? Dummy { get; set; }
+
         [Fact]
         public void AnyChangeInExpressionListTriggersUpdate()
         {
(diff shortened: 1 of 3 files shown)


src/ReactiveUI.Tests/WhenAny/ReactiveNotifyPropertyChangedMixinTest.cs
```diff
@@ -24,6 +24,8 @@ namespace ReactiveUI.Tests
 {
     public class ReactiveNotifyPropertyChangedMixinTest
     {
+        public string? Dummy { get; set; }
```
Evangelink, line 27: Used for one of the tests, not sure how to make it better.

Evangelink (commented):
I am not sure if there are more known-cases for which we could provide a better message.

src/ReactiveUI.Tests/WhenAny/ReactiveNotifyPropertyChangedMixinTest.cs
glennawatson: Probably make it internal and use https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=netcore-3.1

Usually put it in assemblyinfo.cs file

glennawatson (approved):

Issue #2572, opened 2020-11-06
[BUG] TransitionContentControl throwing exception on control width zero


midahenshaw:
Bug #2569 has only states height of zero as just a cause but a width of zero will also cause an Out-Of-Range exception.

I get this while using [MaterialDesignInXamlToolkit](https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit) with __md:DrawerHost__ and __rui:ViewModelViewHost__.

__ViewModelViewHost__ has both __VerticalContentAlignment__ and __HorizontalContentAlignment__ set to _Stretched_.

Let me know if you would like Demo project for testing.

glennawatson:
Demo projects always helpful. If you want to PR that's good too.

glennawatson:
Fixed by 21fb256c155f156a9ae0f83332b9ee6d23389f68

PR #2584, opened 2020-11-16
Fix#2572

kronic:
fix https://github.com/reactiveui/ReactiveUI/issues/2572


glennawatson:
Fixed by 21fb256c155f156a9ae0f83332b9ee6d23389f68

PR #2692, opened 2021-03-02
fix: reactive command doesn't delay execute

### LINKED ISSUE #2689: [BUG] ReactiveCommand Execute() runs before subscribe
**Describe the bug**

According to this documentation comment, ReactiveCommand Execute() should not run the associated command logic until the returned observable is subscribed.

"Invoking this method will return a cold (lazy) observable that, when subscribed, will execute the logic encapsulated by the command. It is worth restating that the returned observable is lazy. Nothing will happen if you call Execute and neglect to subscribe (directly or indirectly) to the returned observable."
https://github.com/reactiveui/ReactiveUI/blob/eaa4abfb970da20cacd9aaac9452eab924d3e235/src/ReactiveUI/ReactiveCommand/ReactiveCommandBase.cs#L153

However, this change replaced the call to
`Observable.StartAsync(execute)`
with
`() => execute().ToObservable()`,
which incorrectly executes the command logic immediately.

93823ff15c1574a21d19b7d48d60891c977301a0

**Steps To Reproduce**
* Call ReactiveCommand.Execute()

**Expected behavior**
* The logic in the command should not be executed (until .Subscribe() or await is used)

glennawatson:
fixes #2689

### DIFF
--- a/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.net472.approved.txt
+++ b/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.net472.approved.txt
@@ -225,6 +225,10 @@ namespace ReactiveUI
         TSender Sender { get; }
         TValue Value { get; }
     }
+    public interface IPlatformOperations
+    {
+        string? GetOrientation();
+    }
     public interface IPropertyBinderImplementation : Splat.IEnableLogger
     {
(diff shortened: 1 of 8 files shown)


glennawatson:
Do not merge this just yet. Rodney is going to do some testing tomorrow to make sure it works well.

PR #2739, opened 2021-04-08
Remove System.ValueTuple package dependency from net472 target framew…

Rivsoft:
This is a simple PR that removes the System.ValueTuple package dependency when targeting net472. This removes issues with binding redirects been required when your solution contains projects targeting both net472 and netstandard2.0 frameworks.


glennawatson (approved):

glennawatson:
I'll do a splat release first so the dependencies are handled. 

Thanks again

PR #2762, opened 2021-05-18
feature make BindToObject, ResolveView and OAPH Nullable

ChrisPulman:
make BindToObject,
ResolveView,
and OAPH Nullable
Fix some other code standard issues in ReactiveUI project

feature

BindToObject, ResolveView and ObservableAsPropertyHelper do not accept nullable types

BindToObject, ResolveView and ObservableAsPropertyHelper now accept nullable types

**What might this PR break?**

non expected

**Other information**:


.editorconfig
```diff
@@ -424,6 +425,9 @@ dotnet_diagnostic.SX1101.severity = error
 dotnet_diagnostic.SX1309.severity = error
 dotnet_diagnostic.SX1623.severity = none
 
+dotnet_diagnostic.IDE0046.severity = none
```
glennawatson, line 428: Probably worth opening up these disables for discussion.

Just curious justifications for each one?

src/ReactiveUI/Activation/CanActivateViewFetcher.cs
```diff
@@ -33,7 +33,7 @@ public class CanActivateViewFetcher : IActivationForViewFetcher
         /// <returns>An observable tracking whether the view is active.</returns>
         public IObservable<bool> GetActivationForView(IActivatableView view)
         {
-            if (!(view is ICanActivate canActivate))
+            if (view is not ICanActivate canActivate)
```
glennawatson, line 36: Probably worth adding a test for this condition to make sure it can handle non-IActivatableView's

src/ReactiveUI/Bindings/Command/CommandBinderImplementation.cs
```diff
@@ -64,7 +64,7 @@ public class CommandBinderImplementation : ICommandBinderImplementation
 
             var bindingDisposable = BindCommandInternal(source, view, controlExpression, Observable.Defer(() => Observable.Return(withParameter())), toEvent ?? string.Empty, cmd =>
             {
-                if (!(cmd is IReactiveCommand rc))
+                if (cmd is not IReactiveCommand rc)
```
glennawatson, line 67: Similar here for a test if ts possible I guess.

src/ReactiveUI/Bindings/Property/PropertyBindingMixins.cs
```diff
@@ -226,7 +226,7 @@ static PropertyBindingMixins()
             Func<TVProp, TVMProp?> viewToVmConverter)
             where TViewModel : class
             where TView : class, IViewFor =>
-            binderImplementation.Bind(viewModel, view, vmProperty, viewProperty, signalViewUpdate, vmToViewConverter, viewToVmConverter);
+            _binderImplementation.Bind(viewModel, view, vmProperty, viewProperty, signalViewUpdate, vmToViewConverter, viewToVmConverter);
```
glennawatson, line 229: Seems this one is not covered by a test.

.editorconfig
ChrisPulman: RCS1158: Static member in generic type should use a type parameter
All the Binding Extensions raise this suggestion, if we want to retain this rule I can go back and mask the warnings in the existing code

IDE0046: Use conditional expression for return
Added as many of the remaining code suggestions will break the code if this is actioned.  if we want to retain this rule I can go back and mask the warnings in the existing code

glennawatson: RCS1158 don't really care about.

IDE0046 make a `suggestion`

src/ReactiveUI/Bindings/Command/CommandBinderImplementation.cs
ChrisPulman: Added tests for IReactiveCommand need to add further tests for ICommand, Found bug in the Func<T> paramater overload where the type is returned as a Unit instead of T

ChrisPulman (commented):
Add Tests for ICommand in CommandBinderImplementation

glennawatson (approved):

PR #2772, opened 2021-05-30
fix: Bind With Trigger, add tests

ChrisPulman:
Added tests to increase codeCov
Added TODO: Create Test to public members without tests to ease the task of creating tests ongoing.
Fixed Bind with trigger to update UI, added test to cover this

Bug fix

Bind with a Observable trigger does not observe the trigger

Bind with a Observable trigger now observe the trigger and updates the view when the trigger is fired

**What might this PR break?**

none unless wrong overload was being used previously and still expecting values to update without the trigger

**Other information**:
other tests added to increase codeCov

### DIFF
--- a/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.net472.approved.txt
+++ b/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.net472.approved.txt
@@ -433,7 +433,7 @@ namespace ReactiveUI
     }
     public static class ObservableFuncMixins
     {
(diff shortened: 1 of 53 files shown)


ronaldvdv:
It seems that since this change, I'm having issues with binding. For example, my trigger is textBox.Events().LostKeyboardFocus(); the textbox remains empty until I've focused it and then moved to the next control.

ronaldvdv:
It seems that since this change, I'm having issues with binding. For example, my trigger is textBox.Events().LostKeyboardFocus(); the textbox remains empty until I've focused it and then moved to the next control. Has this scenario been tested?

PR #2823, opened 2021-07-09
tests Binding to List test

ChrisPulman:
Added Bind to List test with ReadOnlyObservableCollection
put solution items into two folders, Src and Tests

test

added test

**What might this PR break?**

none expected

**Other information**:


glennawatson:
An issue with this one is SourceList is not meant to be exposed to the UI layer at all. Its fundamnetally against the way DD works.

glennawatson:
Why use a SourceList<T> directly is the other concern I have, it's meant to be for data which you observe, transform, filter etc.

ChrisPulman:
Ok, I will rethink this to try a different approach

ChrisPulman:
0.02% hmmm every little helps I guess

glennawatson (approved):

PR #2860, opened 2021-08-06
fix for WhenAny WhenAnyValue ToProperty

ChrisPulman:
**What kind of change does this PR introduce?**

update  WhenAny WhenAnyValue to use nullable for selector
fix #2812 ToProperty to work with nullable OAPH and non nullable OAPH

**What is the current behaviour?**

causes compiler issue or runtime exception

**What is the new behaviour?**

WhenAnyValue and ToProperty can be used in conjunction with each other without compiler warnings or runtime exceptions.

**What might this PR break?**

nullablilty of OAPH needs to be checked.

**Please check if the PR fulfils these requirements**
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Other information**:

PR #2863, opened 2021-08-07
fix #2859 Command IsExecuting

### LINKED ISSUE #2859: [BUG] ReactiveCommand's IsExecuting not set to False after command throwing exception
**Describe the bug**
If a command throws an exception the IsExecuting observable is not set to False

**Steps To Reproduce**
This is a test to reproduce the issue. Thanks to @anaisbetts and @glennawatson for the constant support on Slack.

            [Fact]
            public void Test1()
            {
                        var subj = new Subject<Unit>();
                        var isExecuting = false;
                        Exception fail = null;
                        var fixture = ReactiveCommand.CreateFromTask(async _ => {
                        await subj.Take(1);
                        throw new Exception("die");
                        });
                        fixture.IsExecuting.Subscribe(x => isExecuting = x);

                        Assert.False(isExecuting);
                        Assert.Null(fail);

                        fixture.Execute().Subscribe();
                        Assert.True(isExecuting);
                        Assert.Null(fail);

                        subj.OnNext(Unit.Default);
                        Assert.False(isExecuting);  //here it fails
                        Assert.NotNull(fail);
            }

**Expected behavior**
IsExecuting should be set to False after exception

**Environment**
Blazor Server on .NET 5
- OS:       Windows 10

ChrisPulman:
Added test and updated command to ensure is executing is reset

fix

IsExecuting is not reset in a timely manner

IsExectuting is now reset along with the exception being raised allowing for a more aligned result

**What might this PR break?**

none expected

**Other information**:


ChrisPulman:
pending #2860 to be committed to merge sources

glennawatson (approved):

PR #2926, opened 2021-09-12
housekeeping: Mention new Uwp and WinUI packages

worldbeater:

This PR adds a mention of the new Uwp and WinUI packages to README.md file.

Currently, there is no info regarding Uwp and WinUI packages in README.md, and I got a bit confused when upgrading an app.

**What is the new behavior?**

Now, the docs are added.

**What might this PR break?**

Nothing.


ChrisPulman:
@worldbeater Hi, do you think you can add ReativeUI.Maui too?

ChrisPulman (approved):

worldbeater:
So many new packages ✨

PR #3038, opened 2021-11-26
feature Add Nullable type converters

### LINKED ISSUE #3021: [BUG] Unhandled `OverflowException` from a Textbox Data bound to a Nullable Integer Property
*Note:* this is similar to but different from issue #3000 that I filed recently. That bug was for `int`. This bug is for `int?` and exists in ReactiveUI v16.3.10

**Describe the bug**
Databinding a textbox to an `int?` causes an unhandled `OverflowException` when the user enters an out of range number (i.e. bigger than `int.MaxValue` or smaller than `int.MinValue`). This is inconsistent with the behavior when the user enters a non-integer value (e.g. "hello", or "3a4") or when databinding to an `int`.

**Steps To Reproduce**

1. Have a property in your view model that is an `int?`.
    ```csharp
    public class ExampleViewModel : ReactiveObject
    {
        [Reactive]
        public int? Percentage { get; set; };
    }
    ```
2. Bind that property to the text of a textbox in your view.
    ```csharp
    public ExampleView : ReactiveControl<ExampleViewModel>
    {
        public ExampleView()
        {
            InitializeComponent();
            this.WhenActivated(d =>
            {
                this.Bind(ViewModel, vm => vm.Percentage, v => v.PercentageTextBox.Text).DisposeWith(d);
            });
        }
    }
    ```
3. Add an event handler to the application dispatcher unhandled exception event.
    ```xaml
    <Application x:Class="Example.App"
                 xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                 xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                 Startup="Application_Startup"
                 DispatcherUnhandledException="Application_DispatcherUnhandledException">
        <Application.Resources>
        </Application.Resources>
    </Application>
    ```
    ```csharp
    public partial class App : Application
    {
        private void Application_DispatcherUnhandledException(object sender, DispatcherUnhandledExceptionEventArgs e)
        {
            // TODO report the exception to the user somehow
        }
    }
    ```
4. Run the app and enter a value larger than `int.MaxValue` into the textbox.

When step 4 is done, the unhandled exception handler will be called with an `OverflowException` from the attempt to parse the input string into an `int`.

**Expected behavior**
No unhandled exception. The view model property is unchanged.

When a user enters a non-integer value (e.g. "3a4"), there is no unhandled exception. The view model property is not changed. I would expect the same to happen in this case. 

**Environment**
- OS:       Windows 10 Pro
- Version: 10.0.19043 Build 19043
- Device:   Dell Laptop, Vostro 7590
- ReactiveUI Version:    16.3.10

**Additional context**
- .NET Framework 4.8
- WPF

ChrisPulman:

feature update

no automatic support for nullable binding converters

remove nullable and langversion from project files in favour of solution wide props
Add EnableNETAnalyzers and set to latest level
fix #3021 

**What might this PR break?**

Anyone using self coded Binding converters will need to ensure they have a higher value than 10 for affinity

**Other information**:


src/ReactiveUI.AndroidX/ReactiveRecyclerViewViewHolder.cs
```diff
@@ -88,47 +92,61 @@ protected ReactiveRecyclerViewViewHolder(View view)
...
 
     /// <summary>
-    /// Gets an observable that signals that this ViewHolder has been selected.
+    /// Gets an observable that signals that this ViewHolder has been long-clicked.
     ///
     /// The <see cref="int"/> is the position of this ViewHolder in the <see cref="RecyclerView"/>
     /// and corresponds to the <see cref="RecyclerView.ViewHolder.AdapterPosition"/> property.
     /// </summary>
-    public IObservable<int> Selected { get; }
+    public IObservable<int> LongClicked { get; }
```
glennawatson, line 121: Don't really want all these random changes where properties are located, thought it was a breaking change for a bit.

src/ReactiveUI/ReactiveUI.csproj
```diff
@@ -6,8 +6,6 @@
     <RootNamespace>ReactiveUI</RootNamespace>
     <PackageDescription>A MVVM framework that integrates with the Reactive Extensions for .NET to create elegant, testable User Interfaces that run on any mobile or desktop platform. This is the base package with the base platform implementations</PackageDescription>
     <PackageId>ReactiveUI</PackageId>
-    <Nullable>enable</Nullable>
```
glennawatson, line 9: Was this a mistake?

ChrisPulman: No intentional, they are automatic from the solution props

src/ReactiveUI.AndroidX/ReactiveRecyclerViewViewHolder.cs
ChrisPulman: I also just spotted this, sorry I had my auto format and sort on which caused the rearrangements, I will revert the few files it formatted and reapply the fixes

glennawatson (approved):

PR #3193, opened 2022-02-23
Fix Ensure the appState is loaded only once and GetAppState always returns

### LINKED ISSUE #3192: [BUG] RxApp.SuspensionHost.GetAppState sometimes returns null due to concurrency issues
**Describe the bug**

Sometimes upon restoring AppState, the state would be null, because GetAppState  can be called before app gets loaded and thus before suspension driver loaded app state.
https://user-images.githubusercontent.com/34779382/155291750-546fa71a-8897-4a30-9e8a-7205d1ace8a4.mp4

**Expected behaviour**

GetAppState should always return object, not null.

tomasfil:
fixes #3192


src/ReactiveUI/Suspension/SuspensionHostExtensions.cs
```diff
@@ -65,32 +81,65 @@ public static IDisposable SetupDefaultSuspendResume(this ISuspensionHost item, I
...
-}
\ No newline at end of file
+
+    /// <summary>
+    /// Ensures one time app state load from storage.
+    /// </summary>
+    /// <param name="item">The suspension host.</param>
+    /// <param name="driver">The suspension driver.</param>
+    /// <returns>A completed observable.</returns>
+    private static IObservable<Unit> EnsureLoadAppState(this ISuspensionHost item, ISuspensionDriver? driver = null)
```
glennawatson, line 115: You can possibly use the `Interlocked.Exchange(ref initFunction, null)?.Invoke()` then store a Action or Func with the init logic. This is thread safe as well but avoids the lock

src/ReactiveUI/Suspension/SuspensionHostExtensions.cs
```diff
@@ -16,38 +17,53 @@ namespace ReactiveUI;
...
         {
             throw new ArgumentNullException(nameof(item));
         }
 
-        return item.WhenAny(suspensionHost => suspensionHost.AppState, observedChange => observedChange.Value)
-                   .WhereNotNull()
-                   .Cast<T>();
+        if (item.AppState is null)
+        {
+            item.EnsureLoadAppState();
```
glennawatson, line 44: This would be a threading issue. If you use the interlock approach and always call the method could avoid that

src/ReactiveUI/Suspension/SuspensionHostExtensions.cs
tomasfil: Oh wow, it took me a while, I have never used `Interlocked`, but I understand and this is clever. Commit will be coming right up

glennawatson: Looks like you got the idea correctly.

glennawatson:
I've got to sleep but thanks for this

glennawatson:
Looks like Maui has changed so I'll have to fix those build errors. They are unrelated to your stuff

tomasfil:
Figured that out, but thanks for letting me know anyway

anaisbetts:
I'm a bit confused, loading the AppState may occur more than once during the lifecycle of an app, if an app is suspended then resumed on mobile platforms - why would we want to only load it once?

tomasfil:
I have never developed on mobile platforms, but the state shouldn't be expected to change in storage while suspended no? So the initially loaded one should be still actual.

If I am wrong, then this needs to be done otherwise.

PR #3246, opened 2022-04-27
fix: isExecuting when .Execute throws exception

### LINKED ISSUE #3243: [BUG] ReactiveCommand IsExecuting loses increment count when there is an exception
When a ReactiveCommand's ThrownExceptions is handled, and an exception is thrown, subsequent executions do not set IsExecuting to true.  It also allows two simultaneous executions of the command, and the second one **does** set IsExecuting to true, indicating the execution count had gotten stuck at -1.

I'm sure there are subtleties I'm not aware of, but a first look at the ReactiveCommand code seems quite wrong.  Both the catch and the finally will send the CreateEnd() signal

src/ReactiveUI/ReactiveCommand/ReactiveCommand.cs:718 (and 748)
```
            return Observable.Defer(
                                    () =>
                                    {
                                        _synchronizedExecutionInfo.OnNext(ExecutionInfo.CreateBegin());
                                        return Observable<TResult>.Empty;
                                    }).Concat(_execute(parameter))
                             .Do(result => _synchronizedExecutionInfo.OnNext(ExecutionInfo.CreateResult(result)))
                             .Catch<TResult, Exception>(
                                                        ex =>
                                                        {
/********************************** counter is decremented here ******************************/
                                                            _synchronizedExecutionInfo.OnNext(ExecutionInfo.CreateEnd());
                                                            _exceptions.OnNext(ex);
                                                            return Observable.Throw<TResult>(ex);
                                                        }).Finally(() => 
/********************************** but ALSO decremented here ******************************/
                                                                     _synchronizedExecutionInfo.OnNext(ExecutionInfo.CreateEnd()))
                             .PublishLast()
                             .RefCount();
```
I verified running the ReactiveUI code in the debugger and it his hitting both those blocks.

Sample code to re-create:

```
    public class TestThrownExceptionsViewModel : ReactiveObject
    {
        public ReactiveCommand<Unit, int> ErrorTestCommand { get; }
        [ObservableAsProperty] public bool CommandIsExecuting { get; }
        [ObservableAsProperty] public int CommandOutput { get; }
        [ObservableAsProperty] public string CommandError { get; }

        public TestThrownExceptionsViewModel()
        {
            int counter = 0;
            ErrorTestCommand = ReactiveCommand.CreateFromTask(async () =>
            {
                await Task.Delay(3000);
                counter++;
                if (counter == 2) throw new ApplicationException();
                return counter;
            });
            ErrorTestCommand.ToPropertyEx(this, x => x.CommandOutput);
            ErrorTestCommand.IsExecuting.ToPropertyEx(this, x => x.CommandIsExecuting);
            ErrorTestCommand.ThrownExceptions.Select(ex => ex.Message).ToPropertyEx(this, x => x.CommandError);
        }
    }

```
First execution works normally
Second execution fails normally
Third execution happens, but button is NOT disabled and IsExecuting display remains False
Fourth execution can be started while third execution is in progress.

I'm using ReactiveUI 18.0.7 on WPF on Windows with .NET Framework 4.8

tomasfil:

Closes #3243

Currently the .Execute for the command is implemented in such way, that when the Execute throws and exception, then the inFlightCount results in negative numbers.

**What is the new behavior?**

I have fixed the code to substract inFlightCount only in Finnaly (which takes much longer, thus breaking the specific unit test.)

This ReactiveCommandCreateFromTaskHandlesTaskExceptionAsync now correctly uses scheduler for unit testing and wait longer for the finnaly, which just takes longer.

**What might this PR break?**
The ReactiveCommandCreateFromTaskHandlesTaskExceptionAsync test might still be broken in some cases, because the Finnaly execution time is unpredictable.

- [ .] Tests for the changes have been added (for bug fixes / features)

**Other information**:
ReactiveCommandCreateFromTaskHandlesTaskExceptionAsync should be fixed somehow


tomasfil:
Imo this:
image

Should happen after, or inside the finnaly method somehow, because there is thread race, which is clearly visible in the unit test. 

This would require finnaly method obtaining some of the pipeline data, into which I will look tommorow (heading home now). I have rough idea how to solve that. For now it shouldnt introduce any other bug, only solve the onFly counter.

tomasfil:
This bug was introduced here 
https://github.com/reactiveui/ReactiveUI/pull/2863

Setting the IsExecuting is not timely with this PR, but it does go through.

Pr 2863 broke the counter. This might need some more tweaking.

glennawatson:
I've asked @ChrisPulman to have a look if he has time.

ChrisPulman:
@tomasfil Thanks for the PR I originally added the `_synchronizedExecutionInfo.OnNext(ExecutionInfo.CreateEnd()` to improve the operation of the Cancellation token option, however since this addition it has been traced much deeper to `Observable.FromAsync` I am happy to revert this, thanks very much for your efforts.

ChrisPulman (approved):

Issue #3035, opened 2021-11-25
[BUG] ReactiveUI navigation doesn't change the content of the page


vladislav-itm:
**Describe the bug**
Whe using ReactiveUI nagivation (using IScreen, RoutingState, IRoutableViewModel, etc.), the page content doesn't change when navigating to another page

**Steps To Reproduce**
A repo with a complete MAUI app showing the problem
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue

Steps to reproduce:
1. Create MAUI project from default VS 2022 template (VS 17.1.0 Preview 1.1)
2. Install ReactiveUI.Maui 16.3.10 package
3. Create view models for main page and another page (to which we will navigate). Both implement IRoutableViewModel
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue/blob/master/MauiApp5/MainPageViewModel.cs
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue/blob/master/MauiApp5/AnotherViewModel.cs
4. Modify MainPage.xaml and MainPage.xaml.cs to use ReactiveContentPage, remove unused controls. Bind button to view model's navigation command
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue/blob/master/MauiApp5/MainPage.xaml
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue/blob/master/MauiApp5/MainPage.xaml.cs
5. Create "AnotherPage" and its code-behind file which use ReactiveContentPage
code
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue/blob/master/MauiApp5/AnotherPage.xaml
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue/blob/master/MauiApp5/AnotherPage.xaml.cs
6. Create AppBootstrapper which will host "RoutingState", register views to DI container and return main page as "RoutedViewHost"
code
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue/blob/master/MauiApp5/AppBootstrapper.cs
7. Modify App.xaml.cs to instantiate AppBootstrapper and create main page
https://github.com/vladislav-itm/ReactiveUI-Navigation-Issue/blob/master/MauiApp5/App.xaml.cs
8. Run the application
9. Click "Click me" button
10. Title of the page specified in "UrlPathSegment" gets changed to new one
11. But the page content does change to new one

**Expected behaviour**
[...]

PR #3303, opened 2022-07-02
Fix for MAUI RoutedViewHost

ili:
Fix for #3035

1. Direct `Router.NavigationStack` manipulations are not affected on MAUI's `NavigationPage.Navigation.NavigationStack`
2. MAUI's `NavigationPage.PoppedToRoot` event is not supported
3. Double navigation to the same `VM` occures on application start

**What is the new behavior?**

Bugs are fixed
Sample app: https://github.com/ili/ReactiveUI-Navigation-Issue/tree/fix-maui-navigation

**Other information**:

Direct `NavigationPage.Navigation` manipulations are used to avoid page activations.

### DIFF
--- a/src/ReactiveUI.Maui/RoutedViewHost.cs
+++ b/src/ReactiveUI.Maui/RoutedViewHost.cs
@@ -40,80 +40,30 @@ public class RoutedViewHost : NavigationPage, IActivatableView, IEnableLogger
     {
         this.WhenActivated(disposable =>
         {
-            var currentlyPopping = false;
-            var popToRootPending = false;
-            var userInstigated = false;
-
-            this.WhenAnyObservable(x => x.Router.NavigationChanged)
-                .Where(_ => Router.NavigationStack.Count == 0)
-                .Select(x =>
-                {
-                    // Xamarin Forms does not let us completely clear down the navigation stack
-                    // instead, we have to delay this request momentarily until we receive the new root view
-                    // then, we can insert the new root view first, and then pop to it
-                    popToRootPending = true;
-                    return x;
-                })
-                .Subscribe()
-                .DisposeWith(disposable);
+            var currentlyNavigating = false;
 
             Router?
-                .NavigationChanged?
-                .CountChanged()
-                .Select(_ => Router.NavigationStack.Count)
-                .StartWith(Router.NavigationStack.Count)
-                .Buffer(2, 1)
-                .Select(counts => new
+                .NavigateBack
+                .Subscribe(async _ =>
                 {
-                    Delta = counts[0] - counts[1],
-                    Current = counts[1],
+                    try
+                    {
+                        currentlyNavigating = true;
+                        await PopAsync();
+                    }
+                    finally
+                    {
+                        currentlyNavigating = false;
+                    }
 
-                    // cache current viewmodel as it might change if some other Navigation command is executed midway
-                    CurrentViewModel = Router.GetCurrentViewModel()
+                    InvalidateCurrentViewModel();
+                    SyncNavigationStacks();
                 })
-                .Where(_ => !userInstigated)
-                .Where(x => x.Delta > 0)
-                .Select(
-                        async x =>
-                        {
-                            // XF doesn't provide a means of navigating back more than one screen at a time apart from navigating right back to the root page
-                            // since we want as sensible an animation as possible, we pop to root if that makes sense. Otherwise, we pop each individual
-                            // screen until the delta is made up, animating only the last one
-                            var popToRoot = x.Current == 1;
-                            currentlyPopping = true;
-
-                            try
-                            {
-                                if (popToRoot)
-                                {
-                                    await PopToRootAsync(true);
-                                }
-                                else if (!popToRootPending)
-                                {
-                                    for (var i = 0; i < x.Delta; ++i)
-                                    {
-                                        await PopAsync(i == x.Delta - 1);
-                                    }
-                                }
-                            }
-                            finally
-                            {
-                                currentlyPopping = false;
-                                if (CurrentPage is IViewFor page && x.CurrentViewModel is not null)
-                                {
-                                    page.ViewModel = x.CurrentViewModel;
-                                }
-                            }
-
-                            return Unit.Default;
-                        })
-                .Concat()
-                .Subscribe()
                 .DisposeWith(disposable);
 
             Router?
                 .Navigate
-                .SelectMany(_ => PageForViewModel(Router.GetCurrentViewModel()))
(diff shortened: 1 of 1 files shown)


glennawatson:
Let me test and check out the code for you tomorrow. If it looks good I'll do a release for you then. 

Thanks

src/ReactiveUI.Maui/RoutedViewHost.cs
```diff
@@ -239,4 +183,87 @@ protected IObservable<Page> PageForViewModel(IRoutableViewModel? vm)
...
-}
\ No newline at end of file
+
+    /// <summary>
+    /// Page for view model.
+    /// </summary>
+    /// <param name="vm">The vm.</param>
+    /// <returns>An observable of the page associated to a <see cref="IRoutableViewModel"/>.</returns>
+    [SuppressMessage("Design", "CA1822: Can be made static", Justification = "Might be used by implementors.")]
+    protected Page PageForViewModel(IRoutableViewModel vm)
```
glennawatson, line 193: If this might be used by implementors should this be virtual?

glennawatson: Along with the other methods in there

ili: In general yes, but i'v just copied previous implementation. Mostly I can't see the reason to have method, returning `IObservable<Page>`...

ChrisPulman:
Thanks for this PR. 
I have tested the Android version and this seems to work as expected.

When I tested the same demo on Windows unfortunately the demo displays a blank screen due to a slightly different start-up process. I refactored some of the code to get it working however then I was then faced with a multitude of cross threading issues. 
I will look into this in more detail to see if it's something to do with the manipulation of the core Navigation stack or just a scheduler issue elsewhere.

ChrisPulman:
I can verify that both Windows and Android Targets now operate with the latest commit.
Waiting for @RLittlesII to verify Mac/iOS and to accept the PR

src/ReactiveUI.Maui/RoutedViewHost.cs
```diff
@@ -178,6 +178,10 @@ protected virtual IObservable<Page> PagesForViewModel(IRoutableViewModel? vm)
         ret.ViewModel = vm;
 
         var pg = (Page)ret;
+        MainThread.BeginInvokeOnMainThread(() =>
+        {
+            pg.Title = vm.UrlPathSegment;
+        });
```
ili, line 185: @ChrisPulman there is the problem with setting title - setting it directly will break later binding.

Also if there are cross-thread issues may be it would be better to do smth like this:


```
			Router?
				.Navigate
				.ObserveOn(RxApp.MainThreadScheduler) // <-- here
				.SelectMany(_ => PagesForViewModel(Router.GetCurrentViewModel()))
				.SelectMany(async page =>

```
?

ChrisPulman: I believe that it will only affect the ability to bind the Title, which with other platforms such as Xamarin we set the Title property too. 
However there could be some argument that we shouldn't be setting this property on any platform to allow the implementor to choose what and how the properties are set or bound. 
There should be commonality across the platforms in my opinion.
I'll ask @glennawatson to add his opinion on this.

ili: @ChrisPulman i can't say about Xamarin, may be there are some difference between Xamarin bindings & MAUI compiled bindings, but the fact is - setting `Title` on navigating breaks bindings. I'v commented this code in my local `RoutedViewHost` implementation, and this saves :)

src/ReactiveUI.Maui/RoutedViewHost.cs
```diff
@@ -205,7 +209,13 @@ protected virtual Page PageForViewModel(IRoutableViewModel vm)
 
         ret.ViewModel = vm;
 
-        return (Page)ret;
+        var pg = (Page)ret;
+        MainThread.BeginInvokeOnMainThread(() =>
```
glennawatson, line 214: This isn't using the inbuilt RxUI way of doing this, so might be better to use RxApp.MainThreadScheduler

ili:
@glennawatson @ChrisPulman really sorry if my comment would be stupid and obvious, but i'v faced with Windows cross thread exceptions until added reference to `ReactiveUI.WinUI` to my project:

```
	<ItemGroup Condition="'$(TargetFramework)' == 'net6.0-windows10.0.19041.0'">
	  <PackageReference Include="ReactiveUI.WinUI">
	    <Version>18.2.5</Version>
	  </PackageReference>
	</ItemGroup>
```

src/ReactiveUI.Maui/RoutedViewHost.cs
```diff
@@ -32,6 +32,15 @@ public class RoutedViewHost : NavigationPage, IActivatableView, IEnableLogger
      typeof(RoutedViewHost),
      default(RoutingState));
 
+    /// <summary>
+    /// The Set Title on Navigate property.
+    /// </summary>
+    public static readonly BindableProperty SetTitleOnNavigateProperty = BindableProperty.Create(
```
ChrisPulman, line 38: Added Set Title On Navigate Property

src/ReactiveUI.Maui/RoutedViewHost.cs
```diff
@@ -40,80 +49,31 @@ public RoutedViewHost()
...
-                            return Unit.Default;
-                        })
-                .Concat()
-                .Subscribe()
                 .DisposeWith(disposable);
 
             Router?
                 .Navigate
-                .SelectMany(_ => PageForViewModel(Router.GetCurrentViewModel()))
+                .ObserveOn(RxApp.MainThreadScheduler)
```
ChrisPulman, line 75: Added ObserveOn(RxApp.MainThreadScheduler)

src/ReactiveUI.Maui/RoutedViewHost.cs
```diff
@@ -235,8 +198,97 @@ protected IObservable<Page> PageForViewModel(IRoutableViewModel? vm)
...
+            throw new Exception(msg);
+        }
+
+        ret.ViewModel = vm;
+
+        var pg = (Page)ret;
+
+        if (SetTitleOnNavigate)
+        {
+            RxApp.MainThreadScheduler.Schedule(() => pg.Title = vm.UrlPathSegment);
```
ChrisPulman, line 236: Used RxApp.MainThreadScheduler.Schedule to handle cross threading

ChrisPulman (commented):
Waiting for review of @RLittlesII before approval

glennawatson (approved):

PR #3316, opened 2022-07-19
[FIX] nullable conversions

### LINKED ISSUE #3166: [BUG] Databinding to Nullable Integer Property Does Not Properly Handle Null or Empty String
**Describe the bug**
If you databind a textbox to a nullable integer (`byte`, `int`, etc.) when the textbox is cleared out the view model property will not be set to `null` instead it will stay set to the previous value. I strongly suspect the reverse direction also has problems. That is, if you assign null to the property it would not clear out the textbox. I suspect this was broken by the fix to bug #3021.

**Steps To Reproduce**
1. Have a property in your view model that is an `int?`.
```csharp
public class ExampleViewModel : ReactiveObject
{
    [Reactive]
    public int? Percentage { get; set; } = 42;
}
```
2. Bind that property to the text of a textbox in your view.
```csharp
public ExampleView : ReactiveControl<ExampleViewModel>
{
    public ExampleView()
    {
        InitializeComponent();
        this.WhenActivated(d =>
        {
            this.Bind(ViewModel, vm => vm.Percentage, v => v.PercentageTextBox.Text).DisposeWith(d);
        });
    }
}
```
3. Clear the textbox and check the value of the view model property. It will still have the previous numeric value.

**Expected behavior**
The previous and expected behavior was to set the property to `null` when the textbox was cleared out. Likewise, if the textbox has a value and you set the property to `null`, it should clear the textbox out. I suspect that is not working either.

**Environment**

- OS:  Windows 10 Pro 
- Version   10.0.19043 Build 19043
- Device:   Dell Laptop, Vostro 7590
- ReactiveUI Version:    17.1.17

**Additional context**
The fix to bug #3021 added default `IBindingTypeConverter`s for all the nullable number types. I believe that introduced this bug because they did not properly handle this case. For example, the `NullableIntegerToStringTypeConverter` contains the following code:
```csharp
public bool TryConvert(object? from, Type toType, object? conversionHint, out object result)
{
    if (toType == typeof(string) && from is int fromInt)
    {
        if (conversionHint is int intHint)
        {
            result = fromInt.ToString($"D{intHint}");
            return true;
        }

        result = fromInt.ToString();
        return true;
    }

    if (from is string fromString)
    {
        var success = int.TryParse(fromString, out var outInt);
        if (success)
        {
            result = outInt;

            return true;
        }
    }

    result = null!;
    return false;
}
```
If `from` is `null` or empty string then `int.TryParse` will fail and cause `TryConvert` to return `false` indicating that the conversion failed and preventing assignment to the view model property. Instead, it should check `string.IsNullOrEmpty()` and in that case set the result to `null` and return true.

Looking at the reversion direction, if `from` is `null` then `from is int fromInt` will be false and it will likewise indicate a conversion failure.

tomasfil:
Current implementation doesnt respect the possibility to assign null on null/empty string.

This potentially Fixes #3166.

But this might be breaking change, so I am not sure if it is good idea.

I have personally already stumbled on this issue few times and caused me headache, but never got into checking what is the issue.

Because when I declare number as nullable, then I expect it to be null, when I delete the text from TextBox. The way it is now is counter intuitive.

The fix makes sense to me, but could break lot of apps relying in that """feature""".

### DIFF
--- a/src/ReactiveUI/Bindings/Converter/NullableByteToStringTypeConverter.cs
+++ b/src/ReactiveUI/Bindings/Converter/NullableByteToStringTypeConverter.cs
@@ -46,6 +46,12 @@ public class NullableByteToStringTypeConverter : IBindingTypeConverter
 
         if (from is string fromString)
         {
+            if (string.IsNullOrEmpty(fromString))
+            {
+                result = null!;
+                return true;
+            }
+
             var success = byte.TryParse(fromString, out var outByte);
             if (success)
             {
--- a/src/ReactiveUI/Bindings/Converter/NullableByteToStringTypeConverter.cs
+++ b/src/ReactiveUI/Bindings/Converter/NullableByteToStringTypeConverter.cs
@@ -44,6 +44,12 @@ public class NullableByteToStringTypeConverter : IBindingTypeConverter
             return true;
         }
 
+        if (from is null)
+        {
+            result = null!;
+            return true;
+        }
+
         if (from is string fromString)
         {
             if (string.IsNullOrEmpty(fromString))
--- a/src/ReactiveUI/Bindings/Converter/NullableDecimalToStringTypeConverter.cs
+++ b/src/ReactiveUI/Bindings/Converter/NullableDecimalToStringTypeConverter.cs
@@ -44,6 +44,12 @@ public class NullableDecimalToStringTypeConverter : IBindingTypeConverter
             return true;
         }
 
+        if (from is null)
+        {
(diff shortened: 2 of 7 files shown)


glennawatson:
If you can fix the compile errors your changes look good. Can approve it.

tomasfil:
The build should go through now, missed space..

Are you not afraid of breaking changes resulting from this? I can imagine some apps accidentally building around this bug

glennawatson (approved):

PR #3417, opened 2022-11-18
[FIX] Bump dynamic data version

tomasfil:
Fixes https://github.com/reactiveui/ReactiveUI.Validation/issues/500


tomasfil:
Hello, currently this for some reason doesnt work properly, or release wasnt made recently and ReactiveUI with ReactiveUI.Validation should be re-released, since the Validation breaks (see attached issue). 

If the re-release of Validation will fix the problem, then I am more than happy to close this PR without any change

glennawatson:
Need to fix the build then do a rerelease. Don't know why dotnet is not accepting the project anymore will have to investigate.

PR #3438, opened 2022-12-03
refactor: Clean up code duplication in ReactiveComponentBase classes

rcdailey:
There are three classes affected:

- `ReactiveComponentBase`
- `ReactiveInjectableComponentBase`
- `ReactiveLayoutComponentBase`

The goal is to reduce code duplication between these three classes, as well as eliminating any behavioral differences that may not have been intentional. With that in mind, the changes in each class is outlined below.

# ReactiveComponentBase

This class is now the base class for others. Logically it hasn't changed much. A few methods were made `virtual` in order for them to be overridden in derived classes. Its main purpose now is to house shared logic that was previously duplicated between the other two classes.

Its implementation of `OnAfterRender()` was modified to use the more succinct implementation from `ReactiveLayoutComponentBase`.

# ReactiveInjectableComponentBase

Nearly 100% gutted. It now derives from `ReactiveComponentBase` and only overrides `ViewModel` in order to specify the `[Inject]` attribute.

# ReactiveLayoutComponentBase

I'll be completely honest, I have no idea what the purpose of this class is. I made many efforts to understand:

- `git log` to read the history of the file to make out its purpose.
- Read [certain pull requests][pr] on github that involve changes to this file.
- Ask various individuals on Slack
- Carefully reading the code and comparing it to the implementation of `ReactiveComponentBase`.

Despite these efforts, I failed to find any meaningful differences. Semantically the only difference I found was that it did not use a `CompositeDisposable` to be rid of its connections to the `ViewModel` on its destruction, but I assume this doesn't matter since it can safely assume ownership of the `ViewModel`.

Based on this, I have marked the class `[Obsolete]` to discourage users from using it and migrate them to `ReactiveComponentBase`. It has been kept as an empty class that implements `ReactiveComponentBase` only for backward compatibility.

[pr]: https://github.com/reactiveui/ReactiveUI/pull/2336

### DIFF
--- a/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
+++ b/src/ReactiveUI.Blazor/ReactiveComponentBase.cs
@@ -20,13 +20,14 @@ namespace ReactiveUI.Blazor;
 /// A base component for handling property changes and updating the blazer view appropriately.
 /// </summary>
 /// <typeparam name="T">The type of view model. Must support INotifyPropertyChanged.</typeparam>
-public class ReactiveComponentBase<T> : ComponentBase, IViewFor<T>, INotifyPropertyChanged, ICanActivate, IDisposable
+public abstract class ReactiveComponentBase<T> : ComponentBase, IViewFor<T>, INotifyPropertyChanged,
+    ICanActivate, IDisposable
     where T : class, INotifyPropertyChanged
 {
     private readonly Subject<Unit> _initSubject = new();
     [SuppressMessage("Design", "CA2213: Dispose object", Justification = "Used for deactivation.")]
     private readonly Subject<Unit> _deactivateSubject = new();
-    private readonly CompositeDisposable _compositeDisposable = new();
+    protected readonly CompositeDisposable _compositeDisposable = new();
 
     private T? _viewModel;
 
@@ -37,7 +38,7 @@ public class ReactiveComponentBase<T> : ComponentBase, IViewFor<T>, INotifyPrope
 
     /// <inheritdoc />
     [Parameter]
-    public T? ViewModel
+    public virtual T? ViewModel
     {
         get => _viewModel;
         set
@@ -60,10 +61,10 @@ public class ReactiveComponentBase<T> : ComponentBase, IViewFor<T>, INotifyPrope
     }
 
     /// <inheritdoc />
-    public IObservable<Unit> Activated => _initSubject.AsObservable();
+    public virtual IObservable<Unit> Activated => _initSubject.AsObservable();
 
     /// <inheritdoc />
-    public IObservable<Unit> Deactivated => _deactivateSubject.AsObservable();
+    public virtual IObservable<Unit> Deactivated => _deactivateSubject.AsObservable();
 
     /// <inheritdoc />
     public void Dispose()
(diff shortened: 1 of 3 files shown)


rcdailey:
I apologize for ignoring the PR template. I had already typed up a large description of my change in my git commit and I didn't want to take the time to break it up to fit the template.

rcdailey:
@glennawatson Would love your feedback!

I was not able to compile or test my changes here because of this error:

>   Workarounds.targets(27, 5): If you are building projects that require targets from full MSBuild or MSBuildFrameworkToolsPath, you need to use desktop msbuild ('msbuild.exe') instead of 'dotnet build' or 'dotnet msbuild'

I'm not sure how to address that. I am using Jetbrains Rider as my IDE. 

I also had to modify the list of target frameworks in ReactiveUI.csproj because of errors finding .NET 6.0 for Android and the like. I made this change locally and did not add it to this PR:

```xml
<TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
```

Lastly, I had to unload all projects in the solution except for ReactiveUI and ReactiveUI.Blazor. There were just too many errors in the other projects, but I didn't need to build them anyway as I'm only concerned with Blazor.

rcdailey:
I'm aware of the build failures. I am not going to worry about addressing those until I get feedback on the overall design / approach (in order to avoid wasted effort on my part).

ChrisPulman:
@rcdailey I will take a look at this this evening

ChrisPulman:
@rcdailey Thanks for this PR, I can see what you are trying to achieve however having the separate functionality to extend the base classes of ComponentBase and LayoutComponentBase allows us to add functionality that is specific to those bases, as time goes by the base functionality tends to differ extensively meaning we need to have differing code to match.
On this occasion this is not the direction we wish to go.
Thank you very much for your efforts.

glennawatson:
@rcdailey This is one case where probably having a common base class is not the direction we want, mainly due to performance. If you want to fix the two components of the bugs you were encountering that's fine, this is one scenario where code duplication is fine.

LayoutComponentBase is also not obsolete, still used for making layout components.

rcdailey:
Thanks for the feedback. I do have some concerns / questions:

- What is the actual purpose of `LayoutComponentBase`? The documentation does not make this clear, and the implementation makes it even more confusing as I do not see a meaningful distinction. Can someone explain the differences?
- I understand that having the separate classes allows adding separate functionality, however the current situation is that there is no meaningful difference between them. Am I wrong on that? If I'm not, then wouldn't it be fair to say that when such a need arises, you can then duplicate code as needed based on those requirements?

Here's the issue I see. These classes contain code duplication that incurs significant maintainability overhead. That overhead does not seem fully justified here just based on the percentage of similarity. I didn't run any real numbers but it certainly seems like 90% duplication which is significant.

This also impacts future contributions. The reason why I'm even doing this is because I do not want to implement the same fix 3 different times. That's certainly discouraging and again goes back to the maintainability overhead I mentioned earlier.

I apologize for seeming argumentative. At the end of the day, you guys are calling the shots. I'm just sharing my perspective. My understanding of the objections is that you just don't want this, however I haven't heard any logical reasons. If you're not interested in further explanation that's fine too. I was hoping for more of an open discussion with some more technical deep dive into the PR itself and why what I'm doing isn't suitable.

If you can offer me that, I'd appreciate it.

glennawatson:
The big issue I have is just for a library like RxUI, adding additional inheritance adds potential performance hits. It can be minor except for a scaled-up Blazor app that was relying on performance and suddenly having an additional layer of inheritance.

Layout Component bases have some documentation here https://blazor-university.com/layouts/nested-layouts/

One potential way we can avoid duplication is by using T4-based library scripts, which we've done elsewhere. This is a compile-time feature and avoids the inheritance hit.

rcdailey:
@glennawatson For my own educational benefit, what are the specific performance concerns you have with an additional layer of inheritance?

Issue #3415, opened 2022-11-16
Change POCO type binding warning level to debug


database64128:
**Is your feature request related to a problem? Please describe.**

Currently the POCO type binding warning messages are at the **warn** level. My GUI application has a log widget in the main interface for displaying logs. These warnings tend to spam the logs and are quite annoying:

```
2022-11-16T11:23:46.4433272+01:00 [W] The class YoutubeDl.Wpf.MainWindow property rootDialogHost is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:46.4645744+01:00 [W] The class YoutubeDl.Wpf.MainWindow property mainTabControl is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:47.1145984+01:00 [W] The class YoutubeDl.Wpf.Views.PresetDialogView property nameTextBox is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:47.1169115+01:00 [W] The class YoutubeDl.Wpf.Views.PresetDialogView property formatTextBox is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:47.1170914+01:00 [W] The class YoutubeDl.Wpf.Views.PresetDialogView property containerTextBox is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:47.1268472+01:00 [W] The class YoutubeDl.Wpf.Views.PresetDialogView property backendYoutubeDlCheckBox is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:47.1310636+01:00 [W] The class YoutubeDl.Wpf.Views.PresetDialogView property backendYtDlpCheckBox is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:47.1417348+01:00 [W] The class YoutubeDl.Wpf.Views.PresetDialogView property argumentsItemsControl is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:47.1589010+01:00 [W] The class YoutubeDl.Wpf.Views.HomeView property linkTextBox is a POCO type and won't send change notifications, WhenAny will only return a single value!
2022-11-16T11:23:47.1677745+01:00 [W] The class YoutubeDl.Wpf.Views.HomeView property downloadButton is a POCO type and won't send change notifications, WhenAny will only return a single value!
[...]

PR #3444, opened 2022-12-12
Change POCO type binding warning level to debug #3415

herzbergc:
This change changes POCO type binding warning level to debug

Warning messages are at the warn level

**What is the new behavior?**
Warning messages are at the debug level

**What might this PR break?**
I don't think it should break anything, the same amount of tests passed before and after. The only thing changed is log messages, so it shouldn't harm anything

**Other information**:
Fulfils issue #3415


glennawatson:
The warning is deliberate.

glennawatson:
Warning is what we want in this case, better solution would be just to detect the case of the premade View's and add exemptions to them similar to what we do with winforms.

glennawatson:
Locking this PR, discussion can continue on #3415 thanks.

PR #3475, opened 2023-02-18
Add Feature to MAUI

ChrisPulman:

Feature

No Reactive version of Shell or ShellContent exists in MAUI

Add Feature ReactiveShell
Add Feature ReactiveShellContent
Update Blazor packages
Update AndriodX packages

**What might this PR break?**

Non expected

**Other information**:


ChrisPulman:
Decrease in coverage is due to code from Uwp which we have retired, some of the code is still being used by some projects, but we should do a run on increasing the test coverage again aiming get it closer to 80%.

PR #3501, opened 2023-03-21
Feature ReactiveCommand.CreateRunInBackground

ChrisPulman:

Feature for ReactiveCommand

ReactiveCommand.Create runs Synchronously
ReactiveCommand.CreateFromTask runs Asynchronously
ReactiveCommand.CreateFromObservable runs Asynchronously

To give the option for all Creation options to operate in the same way a new method CreateRunInBackground has been added to the ReactiveCommand creation options making:
	ReactiveCommand.Create runs Synchronously
	ReactiveCommand.CreateRunInBackground runs a Method Asynchronously
	ReactiveCommand.CreateFromTask runs a Task Asynchronously
	ReactiveCommand.CreateFromObservable runs a Observable Asynchronously

**What might this PR break?**

None expected as existing functionality is default

**Other information**:


glennawatson:
Can we change the new flag to be not a optional parameter.

Optional parameter are a breaking API and can break with certain tools, while a method overload is not.

ChrisPulman:
> I have a few questions about this PR, I think we could maybe simplify this implementation quite a bit and make its intent a bit more clear
> 
> * `ReactiveCommand.CreateAsync` - this is a _really_ ambiguous name, especially in the context of the CreateXXX methods - I think a better name would be something like `ReactiveCommand.CreateRunInBackground`, where its intended use is more explicit.
> 
> Also, I think we can really simplify its implementation, as just a simple proxy to `CreateFromObservable`, rather than making all of these internal changes:
> 
> ```cs
> public static ReactiveUI.ReactiveCommand<TParam, TResult> CreateRunInBackground<TParam, TResult>(
>         Func<TParam, TResult> execute, 
>         IObservable<bool>? canExecute = null,
>         // Rather than "runAsync", always provide a way to control concurrency via an IScheduler
>         IScheduler? backgroundScheduler? = null,
>         IScheduler? outputScheduler = null) {
> 
>     return ReactiveCommand.CreateFromObservable(
>         Observable.Start(execute, backgroundScheduler ?? RxApp.TaskPoolScheduler), canExecute, outputScheduler);
> }
> ```
> 
> Thoughts?

Thanks Ani, I will update this evening to realign to these thoughts and test to ensure the result is as desired.

glennawatson (approved):

Issue #3558, opened 2023-06-21
[PR] Support for default generic *-windows moniker in ReactiveUI.WPF 


the-black-wolf:
Continuation of #3557, template is there, we can copy it over if you like. 

Since you said you guys don't want to waste time, I "wasted" some of mine. I analyzed the problem, and the issue with upstream dependency in System.Reactive for -windows is EXACTLY the same you had with windows10.0.17763.0, and the same solution applies. 

**The overrides you made for 17763 work perfectly in a generic -windows target and can be safely used.** 

I made a PR with the change, also combined 17763 triggers into one, will post it in a minute when I get the issue number. 

----------------------------------------------------
Explanation: 

The issue in upstream dependency seems arbitrary and artificial. It is caused by the improperly defined HAS_WPF condition which is only set for 19041 target (how quaint), effectively removing DispatcherScheduler from the net6.0 without exposing the same functionality to net6.0-windows target. 
Didn't waste too much time right now analyzing why in the upstream repo because I already found a fix, but I guess it was just a way for them to "fix" some other issue not directly related to DispatcherScheduler, or it was a leftover from a missing dependency from net5.0, and figuring that out requires more analysis, because ultimately I see no specific Windows release dependency or P/I call, so do not immediately see what prevented net6.0-windows target. Regardless, its a completely different can of worms, starting with the fact they should have made a System.Reactive.WPF package targeting -windows if they wanted to mess with WPF, that code should not have been part of the core System.Reactive to begin with. 

For this particular scenario, the widest usability of ReactiveUI.WPF, and more precisely for -windows targets: binding to net6.0 and adding missing class SOLVES the problem, same as for 17763. 
I tested it on a net7.0-windows targeted wpf app in various scenarios and everything seems to work just fine, but the key point, the DispatcherScheduler standin, I stepped through it and it works perfectly.

Kindly review my changes and include this PR into the main ReactiveUI.WPF for the next release. If you need me to update IntegrationTests.WPF (since right now it only really tests the .net framework build), let me know, I will be willing to help with that too. 

[...]

Issue #3557, opened 2023-06-15
Support for core net7.0-windows moniker | please read


the-black-wolf:
**Is your feature request related to a problem? Please describe.**

Unable to use default win7.0-windows moniker with ReactiveUI, this later daisy chains into inflated requirements for our clients. 

**Describe the solution you'd like**

Unless using some very specific Windows 10 interop calls, please simplify all win7.0-window* package requirements to win7.0-windows. Do the same for net6.0 as well, since its still active. 

**Describe alternatives you've considered**
There are no alternatives, because TFMs are hardcoded into nuget package. The only alternative is to fork ReactiveUI and make a parallel build, which I would prefer not to do, especially over this. This should be resovled at ReactiveUI level. 

**Describe suggestions on how to achieve the feature**
change build TFMs for wpf package. 

**Additional context**

Re the issue #3523 which is now closed.

I get that Win 7 is no longer supported, but I also wished reality is so clear cut as this. meaning there are still legacy Win7 boxes to maintain. 

What exactly is preventing ReactiveUI.WPF from deploying under net7.0-windows (e.g. net7.0-windows7) or even running under it? As I see the package requires the same ReactiveUI dependency regardless of windows version, and ReactiveUI itself can run even under netstandard2, let alone net7-windows. Maybe I am wrong, but I see nothing that is required above the default win7.0-windows

Given that WPF itself binds just fine under win7, dependencies (including 3rd party UI libraries which also do not care that win7 is depracted) are all net7.0-windows, why go to all the trouble to artificially limit this? Shouldn't the guiding principle be reducing the requirements to the lowest usable? 

Among other things this makes our UI libraries which use your library unable to bind on win7.0-windows which is what most WPF projects use now. 

Can you please revisit this issue and simplify package dependencies?

glennawatson:
We have downstream dependencies on reactive which has those constraints. You can attempt to get the simpler versions there first

glennawatson:
One note. Having a larger doesn't exclude you targeting earlier versions of windows etc. you just get the bug fixes in your own programs. I'll close this one unless our downstream dependencies changes their practices.

glennawatson:
System.reactive.

the-black-wolf:
All System.Reactive.* packages bind to .netstandard2, which means they can bind just fine. There is no Windows 10 TFM requirement for any of them 

[...]

PR #3559, opened 2023-06-21
Support for generic -windows targets in ReactiveUI.WPF (#3558)

the-black-wolf:

Feature: Adding support for generic -windows targets

WPF applications require an OS version specific target. This forces the hand of develoeprs and 3rd party library makers. 

#3557 
#3558

**What is the new behavior?**

WPF application targeting net6.0-windows and net7.0-windows can use ReactiveUI.WPF. Also 3rd party libraries for WPF using ReactiveUI.WPF internally can expose the target frameworks.

**What might this PR break?**

Nothing really, it just adds more targets, existing targets are unaffected. 

The integration tests do not contain any dotnet builds as it is, so nothing to expand upon, we can upgrade those tests if you like. I did test the fix in my own WPF app. 

**Other information**:

### DIFF
--- a/src/ReactiveUI.Wpf/ReactiveUI.Wpf.csproj
+++ b/src/ReactiveUI.Wpf/ReactiveUI.Wpf.csproj
@@ -3,8 +3,9 @@
     <!--
     Added 19041 target as a workaround for System.Reactive targeting net5.0-windows10.0.19041.0
     When 19041 is selected by the end user it causes the DispatcherScheduler in System.Reactive to be exposed and used.
+    Added generic "*-windows" target workaround to include missing DispatcherScheduler from System.Reactive, in the same fashion 17763 is fixed. This allows inclusion of ReactiveUI.WPF in 3rd party libraries targeting generic netX.X-windows 
     -->
-    <TargetFrameworks>net462;net472;net48;net6.0-windows10.0.17763.0;net7.0-windows10.0.17763.0;net6.0-windows10.0.19041.0;net7.0-windows10.0.19041.0</TargetFrameworks>
+    <TargetFrameworks>net462;net472;net48;net6.0-windows;net7.0-windows;net6.0-windows10.0.17763.0;net7.0-windows10.0.17763.0;net6.0-windows10.0.19041.0;net7.0-windows10.0.19041.0</TargetFrameworks>
     <PackageDescription>Contains the ReactiveUI platform specific extensions for Windows Presentation Foundation (WPF)</PackageDescription>
     <PackageId>ReactiveUI.WPF</PackageId>
     <UseWpf>true</UseWpf>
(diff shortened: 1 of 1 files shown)


ChrisPulman (commented):
The Target Frameworks supported by ReativeUI are starting from net6.0-windows10.0.17763, we have base libraries that use windows10.0.19041 such as System.Reative, we have created some patches that allow us to go back to windows10.0.17763, this target has been there for some considerable time in these base libraries and actually caused quite a few issues when using net5.0-windows as an example where it failed over to netstandard code instead of the required target leading to things not working quite as expected but also unknown at compile time. 

Recent versions of Visual Studio 2022 have been far better at giving notifications and telling us about these underlying issues, but it would have been pure luck if people managed to run an application with a netx.x-windows target and not have cross threading issues for example.

You are more than welcome to target a lower Target version in your own solutions such as net6.0-windows which is actually net6.0-windows7.0 (Windows 7) however you will import either netstandard or net462 versions of the ReactiveUI libray and sub libraries which may not be 100% compatible with your application/ library or tool, you will then need to create the relevant patches required to make it work as expected.

Unfortunately we cannot keep targeting windows versions that are no longer supported or we will not be able to develop ReactiveUI or the related set of tools that we have.

the-black-wolf:
@ChrisPulman thanks for the reply.  So let me address your concerns. 

Naturally, you cannot target windows10 version target from a generic moniker, that is he core of the problem, it binds the net462 version which of course does not work. Hence the PR. 
However, there are no more issues with net6.0 and forward, whatever was blocking you (and them) on net5.0-windows does not appear to be there any more. If you look at the WPF project, and what you did for the 17763 version, you effectively copied over the missing class and built the observable features on it. The same trick applies to generic moniker. This issue is not caused by you, or the OS, its caused by a faulty HAS_WPF condition in System.Reactive. 

So what happens after this PR, when you bind to the ReactiveUI.WPF from a net6.0-windows (or net7.0-windows)?

1. It loads the ReactiveUI.WPF, finds a net6.0-windows and binds to it
2. This package is dependent on System.Reactive, it loads it and binds to net6.0 version, in the absence of net6.0-windows and failing to match net6.0-windows10.0.19041.0.
3. System.Reactive binds to its own net6.0 dependencies

It does not fall back to netstandard, so there is nothing missing, except the HAS_WPF code, which you compensated for in 17763 workaround. its really that simple, if -windows does not work, neither does 17763, and they both work. They both use the same code from System.Reactive (and all the threading stuff from net6.0/7.0 itself) and your stand-in class. None of that is Windows 10.19041 dependent. 

I have tested this quite a lot since implementing the fix, I have not had a single issue with it. Nothing really changes for your users who already have 17763 or 19041 targets, they will still use what they use today. If you want, I can work on tests for WPF for scenarios you think are problematic. If you know the issues reported for net5.0-windows, we can test them directly on this build. 

As for supporting deprecated OS comment, this is not really about that (although it is a valid concern, especially for maintaining closed systems with difficult upgrade process, such as ATMs et al, some of which are still running even on XP, examples are many, especially in embedded and industrial sector). Real life is not really what Microsoft Windows sales imagines it to be.  As you know WPF is not UWP, it is not windows version dependent. Even Microsofts own WPF libraries (and all the 3rd party libraries we worked with except ReactiveUI) have the generic -windows targets. Even you yourself have a mixed support for windows 10 (you lack 14393, which is still live in LTSC, and if you only want to support GAC, then you should scrap everything and only support 19045). I understand, its a nightmare caused by MS and their update policies. The bizzare thing being that 19041 build of ReactiveUI.WPF app will work just fine even on 10240, most probably even on Windows 7. Which is why having a -windows bypasses all of that. You should not care what OS is supported, unless you specifically use the OS features of that version (which again only matters for UWP). You should only care what tooling is supported (so net6.0 and net7.0 at the moment). 

BTW, unrelated to this, your tests are failing on the WinForms tests. 

Thanks for your time.

glennawatson:
Please stop opening the same issue over and over again. You been given your resolution go to dotnet/reactive and ask them to remove the constraint.

Issue #3190, opened 2022-02-21
Json serialization problem on System.Text.Json


VAllens:
`System.Text.Json` does not use the `System.Runtime.Serialization.IgnoreDataMember` attribute, it uses the `System.Text.Json.Serialization.JsonIgnore` attribute defined by itself.
Therefore, `Changing` and `Changed` and `ThrownExceptions` are serialized.

I have two suggestions: 
1. mark these three properties as `virtual` to allow subclasses to override them.
2. in the `ReactiveObject` and `ReactiveRecord` classes, add the `JsonIgnore` attribute to these three properties, but it needs to reference the `System.Text.Json` package.

FYI. :)

VAllens:
I personally recommend the first option, which gives more flexibility while reducing the possibility of other dependencies.

HugoRoss:
A quick check shows that there are other usages of the `IgnoreDataMember` in the repo:

- ReactiveUI\Platforms\android\ReactiveViewHost.cs
- ReactiveUI\ReactiveObject\ReactiveObject.cs
- ReactiveUI\ReactiveObject\ReactiveRecord.cs
- ReactiveUI\Routing\RoutingState.cs
- ReactiveUI.AndroidSupport\ReactiveRecyclerViewViewHolder.cs
- ReactiveUI.AndroidX\ReactiveRecyclerViewViewHolder.cs
- _ReactiveUI.Tests\ReactiveObject\Mocks\OaphNameOfTestFixture.cs_
- _ReactiveUI.Tests\ReactiveObject\Mocks\OaphTestFixture.cs_
- _ReactiveUI.Tests\ReactiveObject\Mocks\TestFixture.cs_
- _ReactiveUI.Tests\ReactiveObject\Mocks\WhenAnyTestFixture.cs_

Whether these are all candidates for Json serialization with System.Text.Json I can't say.

VAllens:
Does anyone have any suggestions?

glennawatson:
Yeah probably time we bite the bullet and swap to using full on system.text.json.

ChrisPulman:
We have added support for System.Text.Json, please raise a new issue with any new findings, thank you

VAllens:
goooooooooooooooooooooooooood !!!

PR #3562, opened 2023-06-23
Feature : Change Newtonsoft.Json to System.Text.Json

ChrisPulman:

Feature #3190 

No support for System.Text.Json

System.Text.Json support for Serialization added
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-7-0

**What might this PR break?**

If other Serializers are used verification of correct serialisation should be confirmed

**Other information**:


anaisbetts:
This is a Mega Double Plus breaking change because if you're using serialization, you could end up switching serialization frameworks across an upgrade

ChrisPulman:
> This is a Mega Double Plus breaking change because if you're using serialization, you could end up switching serialization frameworks across an upgrade

Yes this will be a Version update

glennawatson:
Yeah going to be a major version bump and some nice big disclaimers.

Issue #3248, opened 2022-04-30
[Bug][WinUI]: ViewModelViewHost not working on WinUI projects.


lighterhy:
### Describe the bug 🐞

ViewModelViewHost not working on WinUI projects.
When I bind a ObservableCollection<TestUserControlViewModel> to a ListView, an XamlParseException occurs with following message:

`Exception thrown: 'Microsoft.UI.Xaml.Markup.XamlParseException' in WinRT.Runtime.dll
WinRT information: The type 'ViewModelViewHost' was not found because 'clr-namespace:ReactiveUI;assembly=ReactiveUI.WinUI' is an unknown namespace. [Line: 1 Position: 276]
XAML parsing failed.
`

Currently I use ViewModelViewHost directly in my DataTemplate to resolve this temporary.

### Step to reproduce

I create a simple project to reproduce this:

[App1.zip](https://github.com/reactiveui/ReactiveUI/files/8595547/App1.zip)

### Reproduction repository

https://github.com/reactiveui/ReactiveUI

### Expected behavior

Get the view rendered by the view model.

### IDE

Visual Studio 2022

### Operating system

Windows

### Version

10.0.22000.652

### Device

Surface Book 2

### ReactiveUI Version

18.0.10

HavenDV:
WinUI has other namespace declarations. Try `xmlns:rx="using:ReactiveUI"`

jdcrutchley:
I can confirm this as well.  Just giving a ListView a name and binding in the code-behind produces an XAML parsing error, but if I give it a DataTemplate containing nothing but a ViewModelViewHost with its ViewModel property set to {x:Bind}, everything works fine.

PR #3575, opened 2023-07-08
Fix For ViewModelViewHost not working on WinUI projects

ChrisPulman:

Fix for #3248 

ViewModelViewHost not working on WinUI projects.

Updated namespace to allow to work

**What might this PR break?**

none expected

**Other information**:

### DIFF
--- a/src/ReactiveUI.Maui/Common/AutoDataTemplateBindingHook.cs
+++ b/src/ReactiveUI.Maui/Common/AutoDataTemplateBindingHook.cs
@@ -23,18 +23,14 @@ namespace ReactiveUI
         /// <summary>
         /// Gets the default item template.
         /// </summary>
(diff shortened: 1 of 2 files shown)


glennawatson (approved):

anaisbetts:
Seems good but make sure this doesn't break any other platform!

glennawatson:
> Seems good but make sure this doesn't break any other platform!

It's the one in the Maui project folder so won't affect anything else.

ChrisPulman:
> Seems good but make sure this doesn't break any other platform!

This would have been an issue on Maui-WinUI and WinUI as was in the same assembly, strangely the XAML Parser does not link it when you specify the same assembly. The Uwp base code that was in use previously has been moved into Maui to allow specific WinUI/Maui code changes, the intention is to remove the Uwp project folder in the near future.

PR #3681, opened 2023-11-26
feature: Allow InvokeCommand to use IReactiveCommand<T, TResult>

SuperJMN:
Use interface intead of implementation.

InvokeCommand **cannot** be used with instances of `IReactiveCommand<T, TResult>`

**What is the new behavior?**

InvokeCommand **can** be used with instances of `IReactiveCommand<T, TResult>`

**What might this PR break?**

Nothing.

**Other information**:
Minor change. Should be fine.

### DIFF
--- a/src/ReactiveUI/ReactiveCommand/ReactiveCommandMixins.cs
+++ b/src/ReactiveUI/ReactiveCommand/ReactiveCommandMixins.cs
@@ -51,10 +51,10 @@ public static class ReactiveCommandMixins
     /// <param name="command">The command to be executed.</param>
     /// <returns>An object that when disposes, disconnects the Observable
     /// from the command.</returns>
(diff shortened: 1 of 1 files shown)


SuperJMN:
Oh, build failures kicked in. I'll take a look later.

glennawatson:
Was just discussing this one between myself and @ChrisPulman 

I think this one might have to be a reject, due to the nature of C#

If the parameters only change based on having two interfaces, C# can't handle it and will cause ambiguity 

The solutions you could potentially do is create a new method `InvokeReactiveCommand(` but that gets away from the point we don't want users if possible `ICommand` except when required by their UI framework.

Also then there is a argument why aren't you using ReactiveCommandBase if you're using a IReactiveCommand

I think it was definitely a nice to have one if the compiler could handle the two interface declarations and would of supported it going in, but unfortunately in this case I don't want to have method name changes imposed on existing code base etc. 

Seemed simple but unfortunately not.

Issue #3640, opened 2023-10-17
[Bug]: On Maui initial NavigateAndReset pushes the same screen twice


kyurkchyan:
### Describe the bug 🐞

I have been following the official [ReactiveUI.Samples Cinephile Maui](https://github.com/reactiveui/ReactiveUI.Samples/tree/main/maui/Cinephile/Cinephile) sample project to create my own Maui Reactive UI project. I've set up basic `MainPage` and `MainViewModel`.  When I launch the app the `MainPage` is pushed twice to the navigation stack. 

  The bug is present in the official sample project as well.
  
  When I looked into the source code of the `RoutedViewHost` here's what I've noticed
  1. When the `RoutedViewHost` is created `SyncNavigationStacksAsync`
  2. `SyncNavigationStacksAsync` notices that the navigation stacks are different - `RoutingState` has 1 view model and the Navigation stack on the Maui side has none, and pushes the main view model
  3. Then `Router.Navigate` is invoked, which pushes the same page again
  
I guess that `SyncNavigationStacksAsync` is invoked to ensure that the screens pushed into the `RoutingState` while the navigation page wasn't displayed are pushed to the stack. However, I am unsure why the `Navigate` observable is fired again during the initialization.

 I've copied `RoutedViewHost` to my project and simply added `Skip(1)` here
 
 ```csharp
....
  Router?
                .Navigate
                .Skip(1)
                .ObserveOn(RxApp.MainThreadScheduler)
                .SelectMany(_ => PagesForViewModel(Router.GetCurrentViewModel()))
                .SelectMany(async page =>
                .....
 ```
 
 This solved the problem for me. However, I am not sure whether this is the right solution in the global context.

### Step to reproduce

1. Create an empty Maui project
2. Add `ReactiveUI.Maui` nuget
3. Create a `MainViewModel` that derives from `ReactiveObject` and implements `IRoutableViewModel`
4. Create `MainPage` that derives from `ReactiveContentPage<MainViewModel>`
5. Register the screen, router and the MainPage <-> Main view model mapping
```csharp
    public static class AppBootstrapper
    {
        public static MauiAppBuilder UseAppBootstrapper(this MauiAppBuilder builder)
        {
            var router = new RoutingState();
            var screen = new AppBootstrapScreen(router);
            Locator.CurrentMutable.RegisterConstant(screen, typeof(IScreen));
[...]

PR #3692, opened 2023-12-05
Fix MAUI RoutedViewHost Navigation

idotta:

Bug fix

#3640 When 'RoutedViewHost' is activated, it calls 'SyncNavigationStacksAsync', which pushes a page for the current ViewModel. However, the Router.Navigate subscription is also triggered, creating and pushing a new page for the same ViewModel.

**What is the new behavior?**

To fix this, I filtered Navigate to execute the action only when the NavigationStacks have different sizes, since it seems that is what the Navigate subscription was supposed to do anyway. I think this makes the subsequent call to SyncNavigationStacksAsync() useless. Should I remove it?

**What might this PR break?**
RoutedViewHost Navigate subscription.

**Other information**:
I was informed that there were no tests for MAUI and no documentation seemed to be present either.

### DIFF
--- a/src/ReactiveUI.Maui/RoutedViewHost.cs
+++ b/src/ReactiveUI.Maui/RoutedViewHost.cs
@@ -76,6 +76,7 @@ public class RoutedViewHost : NavigationPage, IActivatableView, IEnableLogger
 
             Router?
                 .Navigate
+                .Where(_ => Navigation.NavigationStack.Count != Router.NavigationStack.Count)
                 .ObserveOn(RxApp.MainThreadScheduler)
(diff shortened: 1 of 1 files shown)


glennawatson:
Give me a day to read through all this and test etc.

glennawatson:
Documentation is under https://github.com/reactiveui/website repo. At least user facing documentation. XML and inline comments used in the c# code

idotta:
Thank you for being so patient. I mentioned the documentation because I didn't know if there was anything I should add (just following the PR script). Maybe I didn't express myself very well.

PR #3699, opened 2023-12-29
Fix - Remove System.Text.Json from Xamarin targets

ChrisPulman:

Bug Fix for ReactiveUI.Validation

ReactiveUI.Validation uses System.Buffers, however due to missing TFM's for Xamarin in System.Text.Json Xamarin selects the Netstandard target which includes System.Buffers causing incompatibilities.

**What is the new behavior?**

Xamarin targets no longer have System.Text.Json and its features enabled.

**What might this PR break?**

Anyone who has implemented code with System.Text.Json in Xamarin will need to use 19.5.31 until a resolve can be found.

**Other information**:


dpvreony (approved):

glennawatson:
As discussed offline let's wait on this one a little bit.

PR #3718, opened 2024-01-24
Allow Developers Bypass the Default Fallback Behavior (resolves #3713)

### LINKED ISSUE #3713: ViewModelViewHost: Better to Allow Developers Bypass the Default Fallback Behavior That Resovles View With No Contract

I have multiple different views that share the same ViewModel:
```c#
class ViewA : UserControl, IViewFor<WaitingFunVM>
{
}
Locator.CurrentMutable.Register( typeof(ViewA), typeof(IVewFor<WaitingFunVM>));



class ViewB: UserControl, IViewFor<WaitingFunVM>
{
}
 // specify a ContractB here
Locator.CurrentMutable.Register( typeof(ViewB), typeof(IVewFor<WaitingFunVM>),  ConstractB);  

// specify a ViewContract=ConstractB
<rxui:ViewModelViewHost x:Name="myview"  ViewContract="ConstractB" />
```
The current [implementation of `ViewModelViewHost`](https://github.com/reactiveui/ReactiveUI/blob/9c36b0f0701ee7005556ccafaeb503a96ff6b75f/src/ReactiveUI.Wpf/Common/ViewModelViewHost.cs#L163C31-L163C31)  falls back to resolve view with no contract if nothing found.
```
  private void ResolveViewForViewModel(object? viewModel, string? contract)
  {
        // ...
        var viewInstance = viewLocator.ResolveView(viewModel, contract) ?? viewLocator.ResolveView(viewModel);
```
In other words, it will render `ViewA` if nothing found even if I request a `ViewContract=ContractB`.

In my case, the `ViewB` is implemented by other plugin authors and is optional. What I want is:
- if user provide the `ViewB` implementation,  it renders `ViewB`  
- if user doesn't implement `ViewB`, it renders nothing (instead of `ViewA`) 

add a property of `ContractFallbackByPass` for the ViewModelHost so that we can bypass the fallback to viewLocator.ResolveView(viewModel);?

e.g. :
```c#
var viewInstance = viewLocator.ResolveView(viewModel, contract) ;
if(viewInstance == null && !this.ContractFallbackByPass) 
{
    viewInstance = viewLocator.ResolveView(viewModel);
}
```

make the private `ResolveViewForViewModel` protected virtual so that we can override it.
```c#
protected virtual ResolveViewForViewModel(object? viewModel, string? contract) {

}
```

**Describe suggestions on how to achieve the feature**

1. add a property of `ContractFallbackByPass`  with a default value = false for all the ViewModelViewHost
2. or make the private ResolveViewForViewModel method protected virtual

**Additional context**

See discussion https://github.com/reactiveui/ReactiveUI/discussions/3712

newbienewbie:
This changes applys to the Maui/Wpf/XamarinForm platform.

- Feature Request. See  #3713 

1. The ViewModelViewHost resolves view by the ViewContract property. Currently ignores the `ViewContract` condition if nothing found.

**What is the new behavior?**

1. Add a property of `ContractFallbackByPass` so that we can bypass this fallback behavior.
2. Expose a virtual method , i.e. `protected virtual void ResolveViewForViewModel(object? viewModel, string? contract)` , which allows developers override this behavior.

**What might this PR break?**

As far as I can see, it does not break anying.

**Other information**:

For WPF/MAUI/XamForms/WinUI, the `ContractFallbackByPass` is set to false by default. So it won't breaking existing apps. 

However, I find the [current WinForms implementation](https://github.com/reactiveui/ReactiveUI/blob/9c36b0f0701ee7005556ccafaeb503a96ff6b75f/src/ReactiveUI.Winforms/ViewModelViewHost.cs#L210-L211) has no default fallback behaivor as same as WPF

```c#
   var viewLocator = ViewLocator ?? ReactiveUI.ViewLocator.Current;
   var view = viewLocator.ResolveView(x.ViewModel, x.Contract);
   if (view is not null)
   {
       view.ViewModel = x.ViewModel;
       Content = view;
   }
```

So I didn't add such a property for WinForms. Is it better to add such a property that is set to true by default ?

### DIFF
--- a/src/ReactiveUI.Maui/Common/ViewModelViewHost.cs
+++ b/src/ReactiveUI.Maui/Common/ViewModelViewHost.cs
@@ -32,6 +32,12 @@ public class ViewModelViewHost : TransitioningContentControl, IViewFor, IEnableL
     public static readonly DependencyProperty ViewContractObservableProperty =
         DependencyProperty.Register(nameof(ViewContractObservable), typeof(IObservable<string>), typeof(ViewModelViewHost), new PropertyMetadata(Observable<string>.Default));
 
+    /// <summary>
+    ///  The ContractFallbackByPass dependency property.
+    /// </summary>
+    public static readonly DependencyProperty ContractFallbackByPassProperty =
+        DependencyProperty.Register("ContractFallbackByPass", typeof(bool), typeof(ViewModelViewHost), new PropertyMetadata(false));
+
     private string? _viewContract;
 
     /// <summary>
@@ -120,11 +126,25 @@ public class ViewModelViewHost : TransitioningContentControl, IViewFor, IEnableL
     }
 
     /// <summary>
+    /// Gets or sets a value indicating whether should bypass the default contract fallback behavior.
+    /// </summary>
+    public bool ContractFallbackByPass
+    {
+        get => (bool)GetValue(ContractFallbackByPassProperty);
+        set => SetValue(ContractFallbackByPassProperty, value);
+    }
+
+    /// <summary>
     /// Gets or sets the view locator.
     /// </summary>
     public IViewLocator? ViewLocator { get; set; }
 
-    private void ResolveViewForViewModel(object? viewModel, string? contract)
+    /// <summary>
+    /// resolve view for view model with respect to contract.
+    /// </summary>
+    /// <param name="viewModel">ViewModel.</param>
+    /// <param name="contract">contract used by ViewLocator.</param>
+    protected virtual void ResolveViewForViewModel(object? viewModel, string? contract)
     {
         if (viewModel is null)
         {
@@ -133,7 +153,11 @@ public class ViewModelViewHost : TransitioningContentControl, IViewFor, IEnableL
         }
 
         var viewLocator = ViewLocator ?? ReactiveUI.ViewLocator.Current;
-        var viewInstance = viewLocator.ResolveView(viewModel, contract) ?? viewLocator.ResolveView(viewModel);
+        var viewInstance = viewLocator.ResolveView(viewModel, contract);
+        if (viewInstance is null && !ContractFallbackByPass)
+        {
+            viewInstance = viewLocator.ResolveView(viewModel);
+        }
 
         if (viewInstance is null)
         {
--- a/src/ReactiveUI.Maui/ViewModelViewHost.cs
+++ b/src/ReactiveUI.Maui/ViewModelViewHost.cs
@@ -40,6 +40,15 @@ public class ViewModelViewHost : ContentView, IViewFor
      typeof(ViewModelViewHost),
      Observable<string>.Never);
 
+    /// <summary>
+    ///  The ContractFallbackByPass dependency property.
+    /// </summary>
+    public static readonly BindableProperty ContractFallbackByPassProperty = BindableProperty.Create(
+        nameof(ContractFallbackByPass),
+        typeof(bool),
+        typeof(ViewModelViewHost),
+        false);
+
     private string? _viewContract;
 
     /// <summary>
@@ -66,21 +75,7 @@ public class ViewModelViewHost : ContentView, IViewFor
                 {
                     _viewContract = x.Contract;
 
-                    if (x.ViewModel is null)
-                    {
-                        Content = DefaultContent;
-                        return;
-                    }
-
-                    var viewLocator = ViewLocator ?? ReactiveUI.ViewLocator.Current;
-                    var view = (viewLocator.ResolveView(x.ViewModel, x.Contract) ?? viewLocator.ResolveView(x.ViewModel)) ?? throw new Exception($"Couldn't find view for '{x.ViewModel}'.");
-                    if (view is not View castView)
-                    {
-                        throw new Exception($"View '{view.GetType().FullName}' is not a subclass of '{typeof(View).FullName}'.");
-                    }
-
-                    view.ViewModel = x.ViewModel;
-                    Content = castView;
+                    ResolveViewForViewModel(x.ViewModel, x.Contract);
                 })
             });
     }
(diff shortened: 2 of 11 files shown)


glennawatson:
The win forms scenario likely we do want the property added but keep the default the same as current win form functionality.

newbienewbie:
> The win forms scenario likely we do want the property added but keep the default the same as current win form functionality.

Fine. 

And should I also add a new method of `protected virtual void ResolveViewForViewModel(object? viewModel, string? contract)`?

glennawatson:
Yeah. Make it all consistent

ChrisPulman (changes requested):
Please rerun the tests and update the verified files in the API checks, thank you.

newbienewbie:
> Please rerun the tests and update the verified files in the API checks, thank you.

Sorry, at that time, the api vertified tests always failed if I didn't add the following API to verified.txt.  :
```
namespace XamlGeneratedNamespace
{
    public sealed class GeneratedInternalTypeHelper : System.Windows.Markup.InternalTypeHelper
    {
        public GeneratedInternalTypeHelper() { }
        protected override void AddEventHandler(System.Reflection.EventInfo eventInfo, object target, System.Delegate handler) { }
        protected override System.Delegate CreateDelegate(System.Type delegateType, object target, string handler) { }
        protected override object CreateInstance(System.Type type, System.Globalization.CultureInfo culture) { }
        protected override object GetPropertyValue(System.Reflection.PropertyInfo propertyInfo, object target, System.Globalization.CultureInfo culture) { }
        protected override void SetPropertyValue(System.Reflection.PropertyInfo propertyInfo, object target, object value, System.Globalization.CultureInfo culture) { }
    }
}
```
I don't known why the VS generated this `GeneratedInternalTypeHelper`. I'm also pretty new to the Verify. So I added the above snippet to verified.txt and then all the tests succeeded in my local dev environments. 

However, it's quite strange that when I try to add the  `ContractFallbackByPass` property for WinForm today, I have to remove all the `XamlGeneratedNamespace` again otherwise the tests fails. I have no idea about the `GeneratedInternalTypeHelper` phantom for the present.  

Anyway, I'll remove them from the verified.txt later together with the commits for WinForm

ChrisPulman (approved):

ChrisPulman:
@glennawatson are you able to merge this then do another PR to fix the 6 API files that are are failing.

glennawatson:
Believe @newbienewbie is still working on the PR at the moment with some winforms changes

newbienewbie:
> Yeah. Make it all consistent

Hi, I'm currently blocked. I find the current WinForm's behavior is quite different from the WPF's.

## CacheViews

The WinForm's implementation has a unique `CacheViews` flag. Currently, it will [return early](https://github.com/reactiveui/ReactiveUI/blob/9c36b0f0701ee7005556ccafaeb503a96ff6b75f/src/ReactiveUI.Winforms/ViewModelViewHost.cs#L194-L207) if the ViewModel's type remains the same. 
```c#
  if (CacheViews)
  {
      // when caching views, check the current viewmodel and type
      var c = _content as IViewFor;

      if (c?.ViewModel is not null && c.ViewModel.GetType() == viewModel?.GetType())
      {
          c.ViewModel = viewModel;

          // return early here after setting the viewmodel
          // allowing the view to update it's bindings
          return;
      }
  }
```


At the same time, the `ViewContractObservable` is [initialized in the `SetupBindings()` method](https://github.com/reactiveui/ReactiveUI/blob/9c36b0f0701ee7005556ccafaeb503a96ff6b75f/src/ReactiveUI.Winforms/ViewModelViewHost.cs#L172)
```c#
ViewContractObservable = Observable.Return(string.Empty);
```
In the case where I have `CacheViews`=true, even if I change the `ViewContractObservable` to `Observable.Return("ContractB")` later, the view instance remains the same (instead of ViewB)

An easy way is the asumming the `ContractFallbackByPass` is only available when `CacheViews`=false. But I don't know whether it is suitable. 


## DefaultContent

The [WinForm](https://github.com/reactiveui/ReactiveUI/blob/9c36b0f0701ee7005556ccafaeb503a96ff6b75f/src/ReactiveUI.Winforms/ViewModelViewHost.cs#L211) **applies the DefaultContent only if ViewModel is null** and **won't assign the `Content` property when view becomes null**:
```c#
if (ViewModel is null)       // ********************* the DefaultContent is applied only if ViewModel is NULL******************
{
    if (DefaultContent is not null)
    {
        Content = DefaultContent;
    }

    return;
}

if (CacheViews)
{
    // ...
}

 var viewLocator = ViewLocator ?? ReactiveUI.ViewLocator.Current;
 var view = viewLocator.ResolveView(x.ViewModel, x.Contract);
 if (view is not null)     // ********************* change the content only if view is not null ************************
 {
     view.ViewModel = x.ViewModel;
     Content = view;
 } 
```
While the [WPF](https://github.com/reactiveui/ReactiveUI/blob/9c36b0f0701ee7005556ccafaeb503a96ff6b75f/src/ReactiveUI.Wpf/Common/ViewModelViewHost.cs#L165) will set the `Content` = `DefaultCotent` in both cases :
```c#
    if (viewModel is null)
    {
        Content = DefaultContent;
        return;
    }
  ...

  if (viewInstance is null)
  {
      Content = DefaultContent;
      this.Log().Warn($"The {nameof(ViewModelViewHost)} could not find a valid view for the view model of type {viewModel.GetType()} and value {viewModel}.");
      return;
  }
```
My concern is:  in a WinForm App,  if I dynamically change the ViewContractObservable to a contract that is not registered before (e.g. "Contract_That_Doesnot_Exist"), the above `Content = view;` within `if(view is not null) { ... }` won't be executed. 


I don't know how to go futher without breaking existing behavoir.

glennawatson:
Let's just keep winforms as is at the moment. Maybe open a issue with your findings and next major release we can consider changing the functionality. that also allows time for users to see if the new functionality etc.

newbienewbie:
> Maybe open a issue with your findings and next major release we can consider changing the functionality. that also allows time for users to see if the new functionality etc.

I‘ll open a discussion together with a repository later.

PR #3729, opened 2024-02-15
Feature: Add ReactiveProperty

ChrisPulman:

Feature

Boiler plate code is required to get an Observable Property

**What is the new behavior?**

A quick and simple ReactiveProperty has been added to suit a small number of properties.
ReactiveProperty  is a two way bindable declarative observable property with imperative get set Value.
Useful when mixing different UI Frameworks without the requirement for a ReactiveObject based ViewModel

Declare as 
```
IReactiveProperty<string> MyProperty { get; } = new ReactiveProperty<string>();
```

Use declarativly
```
MyProperty.Subscribe(x => // use x as desired);
```

Use imperativly
```
MyProperty.Value = "Set the value";
var value = MyProperty.Value;
```

Use XAML Bindings
```
{Binding MyProperty.Value}
```
**What might this PR break?**

None

**Other information**:

### DIFF
--- a/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.DotNet6_0.verified.txt
+++ b/src/ReactiveUI.Tests/API/ApiApprovalTests.ReactiveUI.DotNet6_0.verified.txt
@@ -12,7 +12,9 @@
 [assembly: System.Runtime.CompilerServices.InternalsVisibleTo("ReactiveUI.Winforms")]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleTo("ReactiveUI.Wpf")]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleTo("ReactiveUI.XamForms")]
+[assembly: System.Runtime.Versioning.SupportedOSPlatform("Windows10.0.17763.0")]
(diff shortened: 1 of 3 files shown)


glennawatson:
Wonder if additions like this, which isn't required for the core code, should go into a ReactiveUI.Extras or something.

ChrisPulman:
> Wonder if additions like this, which isn't required for the core code, should go into a `ReactiveUI.Extras` NuGet package or something.

I think the key thing is to have this within the .Net Foundation so that people can use it in certain situations where they cant use other sources of packages.

As its such a small addition I thought that having it in the base allows it to be used in any implementation.

If it needs to be in a separate package then i can move it to another project within ReactiveUI.

glennawatson:
> > Wonder if additions like this, which isn't required for the core code, should go into a `ReactiveUI.Extras` NuGet package or something.
> 
> I think the key thing is to have this within the .Net Foundation so that people can use it in certain situations where they cant use other sources of packages.
> 
> As its such a small addition I thought that having it in the base allows it to be used in any implementation.
> 
> If it needs to be in a separate package then i can move it to another project within ReactiveUI.

I wouldn't count on RxUI being permanently within the foundation so don't think that is a valid argument.

glennawatson (approved):

PR #3799, opened 2024-04-29
Housekeeping Update Build to use dotnet build

ChrisPulman:

Update build script

using MsBuild

**What is the new behavior?**

using DotNet Build

**What might this PR break?**

none expected

**Other information**:

### DIFF
--- a/src/global.json
+++ b/src/global.json
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "8.0.204",
+        "version": "8.0.10",
         "rollForward": "latestMinor"
     },
     "msbuild-sdks": {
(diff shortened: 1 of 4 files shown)


src/global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "8.0.204",
+        "version": "8.0.10",
```
dpvreony, line 3: am i missing something with .10?  that's a pre release version?

ChrisPulman: "rollForward": "latestMinor" brings it to the latest available, I have had issues with using the latest version as sometimes even though its released the Github Build server fails to get it causing failure

ChrisPulman: Historically we just specified "version": "8.0" this caused an issue when using DotNet Build, MsBuild doesnt care and all is well, now that we are switching away from MsBuild its more important. "version": "8.0.10" is enough detail for it to be happy and roll forward

dpvreony (approved):

PR #3875, opened 2024-08-10
Fix to enable use of DynamicData 9.0

ChrisPulman:

fix

ReactiveUI 20.1.1 includes DynamicData 8
DynamicData 9 has an update to the Observable Change Set which introduces a default value, this results in a mismatch in count between V8 and V9

**What is the new behavior?**

the default value has been removed from the RoutingState CurrentViewModel

**What might this PR break?**

Execution should match previous versions

**Other information**:


dpvreony:
do we need to class this as a breaking change? as the dynamicdata behaviour may have knock on effects?

dpvreony (approved):

PR #3887, opened 2024-08-27
Fix InitRxUI to account for order of passed registration namespaces

Metadorius:

Bug fix?

Originally the `InitializeReactiveUI` function didn't account for order of passed registration, because the allowed namespace dictionary was enumerated instead of the array (which is kind of backwards if you think about it).

**What is the new behavior?**

Order of the namespaces is preserved when registering.

**What might this PR break?**
Multi-platform projects that rely on current non-determinant order may break because of different order of registrations.

**Other information**:
Not sure on what kind of tests/docs are expected, so I am leaving it as draft for now, feel free to undraft if you feel this is OK.


glennawatson:
we'll discuss as a team if we want to keep this PR.

If we bring it across would you be okay just adding some simple unit tests for it?

Metadorius:
> If we bring it across would you be okay just adding some simple unit tests for it?

Sure, I just was not sure what specifically needs to be tested here (or how, because the effect can vary in different setups), so would not mind an idea on what to write tests for.

Yinimi:
I think I had the same issue. When loading data async and trying to update the properties I got a `System.InvalidOperationException: 'The calling thread cannot access this object because a different thread owns it.'`
So it looks like `.ObserveOn(RxApp.MainThreadScheduler)` is just ignored.

I managed to fix this by setting `RxApp.MainThreadScheduler` manually:
**ElementHost parent (WinForms UserControl):**
```
public void ShowWpfContent()
{
    RxApp.MainThreadScheduler = new ControlScheduler(this);
    var myViewModel= Resolve<MyViewModel>();
    var myView = new MyView()
    { DataContext = myViewModel };
    WpfHost.Child = myView;
}
```
Looks like an ugly workaround but it works for me...

PR #3936, opened 2024-12-05
ReactiveUI.Avalonia migration

maxkatz6:

Pushes `ReactiveUI.Avalonia` backend to this repository and updates it to match other backends more.
Additionally, `IntegrationTests.Avalonia` was added.

Avalonia integration is owned by Avalonia.ReactiveUI package, maintained by Avalonia core team.
Functionally, that package is not up to date with new ReactiveUI features, as well as some existing features implemented inconsistently compared to other backends.

**What is the new behavior?**

New ReactiveUI.Avalonia package is going to be created, with code in this repository. 
This PR also already adds some features, that were missing previously:
- [x] `ICreatesObservableForProperty` is a low hanging fruit

And some are pending:
- [ ] `ICreatesCommandBinding` - does it make sense to have Avalonia specific implementation? From WinForms implementation, I don't see a benefit.
- [ ] `IPlatformOperations` - how this one is used? Avalonia `11.2` got Screen.Orientation API, which we can use, but it requires a window/toplevel to get screen from, as there can be multiple. Or we can _assume_ main window from global statics.

**What might this PR break?**

Users who currently use Avalonia.ReactiveUI. NuGet package redirects should be created, and users should be notified.

**Open questions**

1. Testing: Avalonia had ReactiveUI specific tests in the core repository: https://github.com/AvaloniaUI/Avalonia/tree/master/tests/Avalonia.ReactiveUI.UnitTests. It would make sense to move them as well, or even better integrate into existing testing projects. From the first look, there is some connection in `ReactiveUI.Tests` project, but these only run for windows frameworks? How would I add Avalonia implementation there?
2. Initialization: Avalonia.ReactiveUI always had `.UseReactiveUI()` app builder method, it's also used in our [ReactiveUI templates](https://github.com/AvaloniaUI/avalonia-dotnet-templates/blob/main/templates/csharp/app-mvvm/Program.cs#L25). It is different from other backends, which provide `IWantsToRegisterStuff` implementation instead. My question is, would it be preferable to do the same in Avalonia? My only concern is trimmability of library, how is it handled in other backends?

cc @glennawatson @ChrisPulman


src/ReactiveUI.Avalonia/AppBuilderExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+public static class AppBuilderExtensions
+{
+    /// <summary>
+    /// Initializes ReactiveUI framework to use with Avalonia. Registers Avalonia scheduler,
+    /// an activation for view fetcher, a template binding hook.
+    /// Remember to call this method if you are using ReactiveUI in your application.
+    /// </summary>
+    /// <param name="builder">This builder.</param>
+    /// <returns>The builder.</returns>
+    public static AppBuilder UseReactiveUI(this AppBuilder builder)
```
maxkatz6, line 24: This extension method can be replaced with `IWantsToRegisterStuff`. Is this a recommended approach?

src/ReactiveUI.Avalonia/AutoSuspendHelper.cs
```diff
@@ -0,0 +1,90 @@
...
+    private readonly Subject<IDisposable> _shouldPersistState = new Subject<IDisposable>();
+    private readonly Subject<Unit> _isLaunchingNew = new Subject<Unit>();
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AutoSuspendHelper"/> class.
+    /// </summary>
+    /// <param name="lifetime">Pass in the Application.ApplicationLifetime property.</param>
+    public AutoSuspendHelper(IApplicationLifetime lifetime)
+    {
+        RxApp.SuspensionHost.IsResuming = Observable.Never<Unit>();
```
maxkatz6, line 34: With Avalonia 11.1 we can have this observable properly implemented. But likely should be done in dedicated PR, as well as it raises minimal Avalonia version (right now this PR targets 11.0).

src/ReactiveUI.Avalonia/AppBuilderExtensions.cs
dpvreony: considering one of the things on the wishlist for reactiveui is to improve the start up and trimming etc, stick with this?

anaisbetts:
> how this one is used? Avalonia 11.2 got Screen.Orientation API, which we can use, but it requires a window/toplevel to get screen from, as there can be multiple. Or we can assume main window from global statics.

This is used in RoutedViewHost on mobile platforms - on Desktop it should follow what other desktop platforms do (probably return a constant); in general, don't worry about this too much

> ICreatesCommandBinding

We probably don't need it, but it would be nice to get `BindCommand` working on Avalonia, afaik at the moment it does not.

> Users who currently use Avalonia.ReactiveUI. NuGet package redirects should be created, and users should be notified.

"Avalonia.ReactiveUI is deprecated, please move to ReactiveUI.Avalonia" is a phrase no user of either of our platforms should ever have to think about because it is pants-on-head silly. Can we make ".UseReactiveUI" in Avalonia.ReactiveUI and make it depend on ReactiveUI.Avalonia?

glennawatson:
Also nuget has an automatic redirect functionality to at least point the user.

maxkatz6:
@ChrisPulman I am currently busy with internal project. But I still plan to get this PR done. Thanks for keeping it up to date.

ChrisPulman:
> @ChrisPulman I am currently busy with internal project. But I still plan to get this PR done. Thanks for keeping it up to date.

Thank you, we are looking at getting a release out soon. We can schedule it for the next release if that suits you better.

glennawatson:
@maxkatz6 can you check please https://github.com/reactiveui/ReactiveUI.Avalonia

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/reactiveui/ReactiveUI.Avalonia">GitHub - reactiveui/ReactiveUI.Avalonia: Avalonia related components for the ReactiveUI family</a></strong></div><div>Avalonia related components for the ReactiveUI family - reactiveui/ReactiveUI.Avalonia</div></blockquote>

glennawatson:
Closing in favour of the independent repo

Issue #4244, opened 2025-12-27
[Bug]: AutoSuspendHelper warnings


dabbinavo:
### Describe the bug 🐞

Commit fe2a8c73c540f1fdc5721ddab4873d975f4ba06c introduced some warnings as `AutoSuspendHelper` symbol is only existing for target frameworks which include a `src/ReactiveUI/Platforms/.../AutoSuspendHelper.cs` class.

### Step to reproduce

See https://github.com/reactiveui/ReactiveUI/actions/runs/20299208836/job/58300290045#step:11:37

### Reproduction repository

https://github.com/reactiveui/ReactiveUI

### Expected behavior

Build should generate no warnings

### ReactiveUI Version

main branch from commit fe2a8c73c540f1fdc5721ddab4873d975f4ba06c

glennawatson:
This was fixed in commit 1e44aed27 (PR #4245). The fix is already on main and will be included in the next release.

PR #4245, opened 2025-12-27
fix(doc): AutoSuspendHelper not known by all target frameworks

dabbinavo:
Bug fix

See #4244 

**What is the new behavior?**
Build generates no warnings

**What might this PR break?**
Nothing

**Other information**:
AutoSuspendHelper symbol is only existing for target frameworks which include a src/ReactiveUI/Platforms/.../AutoSuspendHelper.cs class.

In order to hide this warning for all other platforms, the `<c>...</c>` syntax is used instead of a `<see cref="..."/>`.

This PR also fixes an ambigous reference issue as there are two overloads of `Activity.OnSaveInstanceState`


glennawatson:
Thanks. Nice to get rid of the warnings.

PR #4358, opened 2026-05-16
fix: support .NET Core in winforms activation design mode check

xackus:
1. Delay the design mode check in Winforms activation until after the control is sited.
2. Use `Control.IsAncestorSiteInDesignMode` (polyfilled on older targets) to support deeply nested controls.
3. Remove the `LicenseManager.UsageMode` check because it doesn't work on .NET Core and is no longer needed on .NET Framework after the other fixes.
4. While at it, reduce nesting of RX operators if the control is a `Form`.

## What is the current behavior?
Design mode detection is broken on .NET Core.

1. The detection is done too early. Controls are not yet sited in their constructor, so `control.Site?.DesignMode` is never true.
`LicenseManager` is not supported on .NET Core. `LicenseManager.UsageMode == LicenseUsageMode.Designtime` is always false.

2. Design mode detection does not support deeply nested controls.
Only the current control and its parent are checked for design mode.

## What is the new behavior?
Design mode detection works both on .NET Framework and .NET Core and supports deeply nested controls.

## What might this PR break?
This shouldn't change anything at runtime, only design time.

## Checklist
- [X] I have read the [Contribute guide](https://www.reactiveui.net/contribute/index.html)
- [ ] Tests have been added or updated (for bug fixes / features)
- [ ] Docs have been added or updated (for bug fixes / features)
- [X] Changes target the `main` branch
- [X] PR title follows [Conventional Commits](https://www.conventionalcommits.org/)

## Additional information
Manually tested on `net481` and `net10.0-windows10.0.19041.0`.

There are still more issues with the Winforms Designer.
Unless you inherit your controls from a base class that ensures ReactiveUI is initialized in its constructor, `WhenActivated` fails to resolve `IActivationForViewFetcher`.
I don't know how to fix this in a way that can be upstreamed.

### DIFF
--- a/src/ReactiveUI.Winforms/ControlPolyfills.cs
+++ b/src/ReactiveUI.Winforms/ControlPolyfills.cs
@@ -3,35 +3,29 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for full license information.
 
-#if !NET6_0_OR_GREATER
-
 namespace ReactiveUI.Winforms;
 
 internal static class ControlPolyfills
 {
-    extension(Control control)
+    public static bool GetIsAncestorSiteInDesignMode(this Control control)
     {
-        public bool IsAncestorSiteInDesignMode
-        {
-            get
-            {
-                ArgumentExceptionHelper.ThrowIfNull(control);
+#if NET6_0_OR_GREATER
+        return control.IsAncestorSiteInDesignMode;
+#else
+        ArgumentExceptionHelper.ThrowIfNull(control);
 
-                if (control.Site is { DesignMode: true })
-                {
-                    return true;
-                }
-                else if (control.Parent is null)
-                {
-                    return false;
-                }
-                else
-                {
-                    return control.Parent.IsAncestorSiteInDesignMode;
-                }
-            }
+        if (control.Site is { DesignMode: true })
+        {
+            return true;
         }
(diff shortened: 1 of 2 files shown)


src/ReactiveUI.Winforms/ControlPolyfills.cs
```diff
@@ -0,0 +1,37 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+#if !NET6_0_OR_GREATER
+
+namespace ReactiveUI.Winforms;
+
+internal static class ControlPolyfills
+{
+    extension(Control control)
```
glennawatson: extension blocks can cause us dramas with stylecop, so we are avoiding for the moment.

glennawatson: Let's change back to a method please, but you can make a regular extension method instead which should allow you to keep most of the code as is.

xackus: I changed it to an extension method and renamed it to `GetIsAncestorSiteInDesignMode` (if it has the same name as the property, it doesn't show up in IDE suggestions).

glennawatson:
Might be about a week before we release btw. Just waiting on validation of a new code signing certificate.

PR #4381, opened 2026-06-13
Fix: WhenAnyValue subscribes to PropertyChanged before reading the initial value

### LINKED ISSUE #4380: Race condition: WhenAnyValue loses PropertyChanged events fired between its initial read and its handler attach
## What happens

`WhenAnyValue`'s chain sink reads the current property value, emits it downstream, and *then* subscribes to `INotifyPropertyChanged.PropertyChanged`. If another thread writes the property between the read and the subscribe, the resulting `PropertyChanged` invocation hits an empty handler list and the new value is dropped. The downstream observer stays pinned to the pre-write value until the next change.

This bites any view model whose properties can be written off the subscribing thread. That includes the common MVVM pattern of a view model that aggregates data from background services (network, IPC, file watchers, device callbacks): the subscriber sits on the UI/dispatcher thread, the producer runs on a worker thread, and they meet at a `ReactiveObject` property in between.

## Reproduction

I have a standalone Gist that reproduces the bug deterministically against the published `ReactiveUI 23.2.28` package:

<https://gist.github.com/dwcullop/5964512be500b4f9829af1c40d53a964>

```
dotnet run
```

Output:
```
Final property value : 2
Values observed       : [1]
Last observed         : 1

FAIL: subscriber's last observed value does NOT match property
```

The Gist runs on one thread. It works because the chain sink holds no synchronization between its read and its subscribe, so anything that runs in that window (whether that's a real second thread or a synchronous reentrant call from the subscriber's own initial `OnNext`) puts the sink in the same broken state.

To confirm the same defect surfaces under genuine concurrency, here is a multi-threaded stress test: one mutator thread writes the property 32 times in a tight loop, the main thread subscribes via `WhenAnyValue`. After both threads finish, the subscriber's last value should equal the property's final value. Over 2 000 iterations, ~7.5% diverge:

```csharp
[Test]
public async Task WhenAnyValue_ConcurrentMutationDuringSubscribe_NeverLosesFinalValue_Stress()
{
    const int iterations = 2_000;
    const int mutationsPerIteration = 32;
    var divergences = new List<string>();

    for (var i = 0; i < iterations; i++)
    {
        var fixture = new TestFixture { IsOnlyOneWord = "v0" };
        using var mutatorReady = new ManualResetEventSlim(false);
        using var mutatorDone = new ManualResetEventSlim(false);

        var mutator = new Thread(() =>
        {
            mutatorReady.Set();
            for (var j = 1; j <= mutationsPerIteration; j++)
            {
                fixture.IsOnlyOneWord = "v" + j.ToString(CultureInfo.InvariantCulture);
            }
            mutatorDone.Set();
        }) { IsBackground = true };
        mutator.Start();
        mutatorReady.Wait();

        string? latest = null;
        using (fixture.WhenAnyValue(x => x.IsOnlyOneWord).Subscribe(v => latest = v))
        {
            mutatorDone.Wait();
            mutator.Join();
        }

        if (!string.Equals(latest, fixture.IsOnlyOneWord, StringComparison.Ordinal))
        {
            divergences.Add($"iter {i}: latest='{latest}' property='{fixture.IsOnlyOneWord}'");
        }
    }

    await Assert.That(divergences).IsEmpty();
}
```

Failure output (excerpt):
```
iter 12: latest='v31' property='v32'
iter 63: latest='v26' property='v32'
iter 68: latest='v28' property='v32'
iter 79: latest='v27' property='v32'
iter 113: latest='v12' property='v32'
...and 144 more
```

Each `latest='v<n>' property='v32'` is a lost write: the mutator's later writes raised `PropertyChanged` while the chain sink was between its read and its subscribe.

## Root cause

The same defect (read-and-emit before subscribing to `PropertyChanged`) is present in both the released `23.2.28` code path and `main`, in different shapes:

**`23.2.28`** — `src/ReactiveUI/Mixins/ReactiveNotifyPropertyChangedMixin.cs`, `NestedObservedChanges`:

```csharp
// Make sure a change at a root node propagates events down
var kicker = new ObservedChange<object?, object?>(sourceChange.Value, expression, default);

return NotifyForProperty(sourceChange.Value, expression, beforeChange, suppressWarnings)
    .StartWith(kicker)
    .Select(static x => new ObservedChange<object?, object?>(x.Sender, x.Expression, x.GetValueOrDefault()));
```

`StartWith(kicker)` emits the kicker to the downstream *before* `NotifyForProperty(...).Subscribe(...)` is called. The downstream `.Select(... x.GetValueOrDefault())` reads the property when projecting the kicker. Any `PropertyChanged` raised between that read and the `Subscribe` to `NotifyForProperty` lands on an empty handler list and is lost.

**`main`** (post the custom-reactive-sinks rework) — `src/ReactiveUI/Internal/ExpressionChainSink.cs`, `Level.SetParent`:

```csharp
// Kicker: propagate the current value immediately, then subscribe for updates.
Push(ReadValue(parent));
_subscription.Disposable = sink._notify(parent, link, sink._beforeChange, sink._suppressWarnings)
    .Subscribe(new Observer(this));
```

The comment in the source already names the ordering: read and emit, *then* subscribe. Same defect as `23.2.28`'s `NestedObservedChanges`, just in a custom sink instead of via Rx operators.

## Fix

The conceptual fix is the same in both code paths: **subscribe to `PropertyChanged` before reading and emitting the initial value**. A racing mutation that fires between the subscribe and the read is then queued behind the chain's serialization and gets delivered cleanly.

For `23.2.28`, that means restructuring `NestedObservedChanges` so the `NotifyForProperty(...)` subscription is established before the kicker is delivered to the downstream (e.g., via `Observable.Create` with the subscribe-first pattern, or `Defer` plus an explicit ordering).

For `main`, swap the two statements in `Level.SetParent`: subscribe first, then `Push(ReadValue(parent))`. When the race fires under the new order, both the kicker's emit and the racing handler's queued emit report the same post-mutation value, so the downstream would see the same value twice. The leaf's existing `isDistinct` gate handles this for the default `WhenAnyValue` path; the `isDistinct: false` overloads need a separate one-shot "suppress next emission if equal to the kicker" guard so the user-visible distinct semantics don't change.

## Environment

- ReactiveUI **23.2.28**
- .NET 8 / 9 / 10 SDKs
- Reproduced on Windows 11; defect is platform-independent

dwcullop:
Fixes #4380.

`ExpressionChainSink.Level.SetParent` reads the property and emits the kicker value before subscribing to `PropertyChanged`. Any mutation that fires `PropertyChanged` between the two is raised against an empty handler list and silently lost, leaving downstream observers pinned to the pre-write value until the next change. The race fires non-deterministically when the property can be written from a thread other than the one calling `Subscribe` (typical MVVM pattern: a view model aggregating data from background services).

This swaps the order to subscribe-then-read. A racing handler emit that arrives queued behind the chain sink's gate then delivers the same post-mutation value the kicker just emitted, so a one-shot `_suppressNextIfSameAsLast` guard collapses the single racing duplicate. The guard fires independently of the user-facing `isDistinct` flag, so existing distinct semantics are unchanged.

## Scope

The same defect exists in the released `23.2.28` code (via `NestedObservedChanges.StartWith(kicker)` then `Subscribe(NotifyForProperty)`), but that code path is being replaced by the custom-reactive-sinks rework on `main`. This PR fixes the defect in the rework's new location so the next release ships with the fix. Backporting to a `23.2.x` hotfix branch (if one is maintained) would be a separate change.

## Tests

In `WhenAnyValueSubscribeRaceTests`:

- `WhenAnyValue_MutationBetweenInitialEmitAndHandlerAttach_IsLost` — deterministic single-thread repro on `ReactiveObject`.
- `..._IsLost_PlainInpc` — same shape with a hand-rolled `INotifyPropertyChanged` source, to rule out `ReactiveObject` specifics.
- `WhenAnyValue_ConcurrentMutationDuringSubscribe_NeverLosesFinalValue_Stress` — 2 000-iteration multi-threaded stress (~7.5% failure rate without the fix; 0% with it).

All three fail without the fix and pass with it. Full `ReactiveUI.Tests` suite (1 899 pre-existing + 3 new) passes locally on `net8.0`.

### DIFF
--- a/src/ReactiveUI/Internal/ExpressionChainSink.cs
+++ b/src/ReactiveUI/Internal/ExpressionChainSink.cs
@@ -81,6 +81,16 @@ internal sealed class ExpressionChainSink<TSender, TValue>(ExpressionChainParame
         /// <summary>Whether <see cref="_last"/> holds a value yet.</summary>
         private bool _hasLast;
 
+        /// <summary>
+        /// When <see langword="true"/>, the next emission that equals <see cref="_last"/> is suppressed
+        /// regardless of <see cref="_isDistinct"/>. Set immediately after the kicker emits during
+        /// <see cref="Level.SetParent"/>: any <c>PropertyChanged</c> event that races the
+        /// subscribe-then-read window has been queued behind <see cref="_gate"/> and will re-emit the
+        /// same value once we release the lock. The suppression collapses that single racing duplicate
+        /// without altering the user-facing distinct semantics.
+        /// </summary>
+        private bool _suppressNextIfSameAsLast;
+
         /// <summary>Latched once this chain subscription has been disposed.</summary>
         private bool _disposed;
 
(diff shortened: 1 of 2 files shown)


glennawatson (approved):

glennawatson:
Thanks for the PR. I got a major refactor incoming (mostly all internal) where we swap over internally to using ReactiveUI.Primitives, basically still keeps IObservable<T> but swaps out a lot of the internals. 

So would like to get this fix in before that refactor.

We also have a source generated version of the WhenAny pipeline in the project in ReactiveUI.Binding.SourceGenerators mind having a peak and seeing if the issue is also there @dwcullop

glennawatson:
The fails are false negatives. Going to get your fixes in as part of the refactor.

dwcullop:
Happy to help. 😄 Let me know if you want a fix for the `23.2.28` version.  It's very similar.

I'll take a look at the Binding SourceGen repo as well and see if I find any issues.

PR #4382, opened 2026-06-16
feat: re-platform ReactiveUI onto ReactiveUI.Primitives — leaner, faster, System.Reactive optional

glennawatson:
## ⚠️ Breaking Changes

ReactiveUI now ships in **two interchangeable distributions with an identical public API**. Both are built on
the **same ReactiveUI.Primitives internals and the same high-performance custom schedulers/sinks** — the *only*
difference is which reactive **interop types** appear in the public API:

| You want… | Reference these packages | Public reactive types |
|---|---|---|
| The new, lighter default (no System.Reactive dependency) | `ReactiveUI`, `ReactiveUI.Wpf`, `ReactiveUI.WinForms`, `ReactiveUI.WinUI`, `ReactiveUI.Maui`, `ReactiveUI.Blazor`, `ReactiveUI.AndroidX`, … | ReactiveUI.Primitives — `RxVoid`, `ISequencer`, `Signal<T>` |
| Drop-in interop with existing System.Reactive code | `ReactiveUI.Reactive`, `ReactiveUI.Wpf.Reactive`, `ReactiveUI.WinForms.Reactive`, `ReactiveUI.WinUI.Reactive`, `ReactiveUI.Maui.Reactive`, `ReactiveUI.Blazor.Reactive`, … | System.Reactive — `Unit`, `IScheduler` |

`WhenAnyValue`, `ReactiveCommand`, `ToProperty`/OAPH, bindings, routing and interactions are the same in both — you
pick a distribution, you don't rewrite code. **The `.Reactive` family is not "old ReactiveUI"**: it runs on the exact
same Primitives engine and custom schedulers as the default; it simply surfaces `System.Reactive.Unit`/`IScheduler`
(and `Subject<T>`) in its public API so it composes with code that already uses System.Reactive.

**If you upgrade the default `ReactiveUI` packages**, the public reactive types change:

- `System.Reactive.Concurrency.IScheduler` → `ReactiveUI.Primitives.Concurrency.ISequencer` (e.g. `RxApp.MainThreadScheduler`/`TaskpoolScheduler`, also exposed via `RxSchedulers`).
- `System.Reactive.Unit` → `ReactiveUI.Primitives.RxVoid` (e.g. `ReactiveCommand<Unit, Unit>` → `ReactiveCommand<RxVoid, RxVoid>`).
- `Subject<T>` / `BehaviorSubject<T>` / `ReplaySubject<T>` → `Signal<T>` / `BehaviorSignal<T>` / `ReplaySignal<T>`.
- The default packages **no longer pull in System.Reactive transitively**. If your code also calls System.Reactive operators (`System.Reactive.Linq`), add a direct `System.Reactive` reference or use the `.Reactive` packages.

**DynamicData is no longer a dependency of core.** The DynamicData-backed routing and collection helpers (the
change-set / `ToObservableChangeSet`-style routing and auto-persist mixins) moved into a new **`ReactiveUI.Routing`**
package (`ReactiveUI.Routing.Reactive` for the System.Reactive flavor). Core `RoutingState`/`IScreen`/`RoutedViewHost`
stay in the main package; if you used the DynamicData routing/collection extensions, add `ReactiveUI.Routing`.

**To take the upgrade with zero source changes**, reference the matching `ReactiveUI*.Reactive` packages — they keep
`IScheduler`, `Unit` and `Subject<T>`.

## Why this change?

System.Reactive is a large, general-purpose dependency, but ReactiveUI only uses a small, well-defined slice of it —
schedulers, a handful of subjects, and a few operators behind `WhenAnyValue`/`ToProperty`/`ReactiveCommand`. Carrying
the whole library meant every ReactiveUI app inherited its size, its transitive footprint, and its allocation/
throughput characteristics on the hottest MVVM paths.

**ReactiveUI.Primitives** is a purpose-built, allocation-conscious reactive core with fused sinks tailored to exactly
those paths, plus **our own scheduler implementations written largely for speed**. As part of this, **most of the
schedulers now live in ReactiveUI.Primitives** and are shared by both distributions. Moving onto it lets us:

- **Drop the System.Reactive (and DynamicData) dependency** from the default packages — smaller closure, better
  trimming/AOT story.
- **Cut allocations and time** dramatically on the operators apps use constantly, for *both* distributions.
- **Keep everyone supported** — the `.Reactive` distribution keeps the System.Reactive interop types, so this is an
  opt-in change of public types, not a forced rewrite.

### Performance

Representative micro-benchmarks (BenchmarkDotNet, net10.0), last fully-System.Reactive release **23.2.28** vs the new
**default (Primitives)** distribution. Lower is better; allocation is per operation.

**`WhenAnyValue` — subscribe**

| Scenario | 23.2.28 (System.Reactive) | New default (Primitives) |
|---|---|---|
| Arity 1 | 2,900 ns · 3.63 KB | **736 ns · 3.06 KB** |
| Arity 2 | 5,697 ns · 7.57 KB | **1,634 ns · 3.47 KB** |
| Arity 3 | 8,322 ns · 11.25 KB | **2,524 ns · 5.20 KB** |

**`WhenAnyValue` — emit (10k changes)**

| Scenario | 23.2.28 (System.Reactive) | New default (Primitives) |
|---|---|---|
| Arity 1 | 3,325 µs · 6,797 KB | **707 µs · 509 KB** |
| Arity 2 | 3,915 µs · 6,797 KB | **1,134 µs · 703 KB** |
| Arity 3 | 4,115 µs · 7,031 KB | **1,268 µs · 937 KB** |

**`ToProperty` / OAPH**

| Scenario | 23.2.28 (System.Reactive) | New default (Primitives) |
|---|---|---|
| Create | 7.33 µs · 7.76 KB | **1.02 µs · 1.93 KB** |
| Emit (10k) | 5,893 µs · 10,859 KB | **1,961 µs · 2,109 KB** |

Net effect: roughly **3–4× faster** on subscribe and emit, and **5–13× less allocation** (e.g. `WhenAnyValue.Emit`
arity 1 drops from ~6.8 MB to ~0.5 MB — a ~13× reduction; `ToProperty.Create` from ~7.3 µs to ~1.0 µs). Because the
fast schedulers/sinks live in Primitives and back both distributions, the `.Reactive` family improves over 23.2.28 too
while keeping the System.Reactive interop types.

> Numbers are from a short BenchmarkDotNet job on a developer machine, so absolute values have wide error bars — but
> the released-vs-new gap is far beyond run-to-run noise.

Feature — a re-platforming of ReactiveUI onto ReactiveUI.Primitives, delivered as two interchangeable distributions.

## What is the new behavior?

- Every ReactiveUI library is produced in two flavors from one shared source:
  - the **default** family (`ReactiveUI`, `ReactiveUI.Wpf`, …) exposes the **ReactiveUI.Primitives** interop types and
    carries **no System.Reactive dependency**;
  - the **`.Reactive`** family (`ReactiveUI.Reactive`, `ReactiveUI.Wpf.Reactive`, …) exposes the **System.Reactive**
    interop types (`Unit`/`IScheduler`) over the same Primitives engine.
- Both share the same internals and the new custom schedulers/sinks; **most schedulers now live in Primitives**.
- DynamicData integration (change-set routing/collection/auto-persist mixins) is now a separate **`ReactiveUI.Routing`**
  package, so core no longer depends on DynamicData.
- Same public surface across distributions, so you choose a distribution rather than rewriting code.
- Public API baselines (`PublicAPI.Shipped.txt`) are tracked again for every target framework so the surface of both
  distributions is locked and reviewable.

## What is the current behavior?

ReactiveUI is a single distribution built directly on System.Reactive and DynamicData, so every consumer takes those
transitive dependencies and the System.Reactive types (`IScheduler`, `Unit`, `Subject<T>`) appear throughout the public
surface. There is no System.Reactive-free option.

## What might this PR break?

See **Breaking Changes** above. In short: upgrading the default `ReactiveUI*` packages swaps the public reactive types
(`IScheduler`→`ISequencer`, `Unit`→`RxVoid`, `Subject`→`Signal`) and drops the transitive System.Reactive/DynamicData
dependencies; the DynamicData routing/collection helpers move to `ReactiveUI.Routing`. Consumers who want no source
changes can move to the `ReactiveUI*.Reactive` packages, which retain `Unit`/`IScheduler`/`Subject<T>`.

## Checklist
- [x] I have read the [Contribute guide](https://www.reactiveui.net/contribute/index.html)
- [x] Tests have been added or updated (for bug fixes / features)
- [x] Docs have been added or updated (for bug fixes / features)
- [x] Changes target the `main` branch
- [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/)

## Additional information

Verification: the full solution builds clean on Linux and Windows; the core, routing, AOT, builder, Splat, testing and
Blazor suites are green on Linux, and the WPF (330) and WinForms (2011) suites are green on real Windows for both
distributions.


src/ReactiveUI.Core/Mixins/PreserveAttribute.cs
```diff
@@ -7,18 +7,14 @@
...
-/// Indicates that the attributed code element should be preserved during linking or code optimization processes.
-/// </summary>
+/// <summary>Indicates that the attributed code element should be preserved during linking or code optimization processes.</summary>
 /// <remarks>Apply this attribute to prevent code elements from being removed or altered by tools that perform
 /// code trimming, such as linkers or obfuscators. This is typically used to ensure that reflection or dynamic access to
 /// the member remains functional after build-time optimizations.</remarks>
 [ExcludeFromCodeCoverage]
 [AttributeUsage(AttributeTargets.All)]
-internal sealed class PreserveAttribute : Attribute
+public sealed class PreserveAttribute : Attribute
```
ChrisPulman, line 16: Should this stay internal

src/ReactiveUI.Core/ReactiveUI.Core.csproj
```diff
@@ -0,0 +1,66 @@
...
+       platform libs reach them via the Core reference + IVT). EventPatternObservable is consumed by
+       core's CreatesCommandBindingViaEvent, so it is linked here too. -->
+  <ItemGroup>
+    <Compile Include="..\Shared\ArgumentValidation.cs" Link="Shared\ArgumentValidation.cs"/>
+    <Compile Include="..\Shared\DelegateObserver.cs" Link="Shared\DelegateObserver.cs"/>
+    <Compile Include="..\Shared\Platform\EventPatternObservable.cs" Link="Shared\Platform\EventPatternObservable.cs"/>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="ReactiveUI.Primitives" ExcludeAssets="analyzers" />
```
ChrisPulman, line 45: This is correct, but I spotted some places where the analysers were not turned off.

ChrisPulman (approved):

